PR #1, opened 2019-04-05
Add docker setup for development purposes

trimailov:
Turned to be a bit heavy pull request.

Added:
- `docker-compose.yml`
- `Dockerfile` for the python app
- script to check for PSQL available connections
- ENV var and a fix in `conftest.py` to be able to set test database DSN

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,28 @@
+FROM python:3.7-alpine as builder
+
+WORKDIR /opt/
+
+COPY requirements-dev.txt .
+
+RUN apk update && \
+    apk add --virtual build-deps python-dev build-base && \
+    apk add postgresql-dev libxslt-dev libxml2-dev
+
+RUN pip wheel --wheel-dir=/tmp -r requirements-dev.txt
+
+
 FROM python:3.7-alpine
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+COPY --from=builder /tmp/*.whl /tmp/wheels/
+
+COPY . .
+
+RUN pip install --no-index -f /tmp/wheels -r requirements-dev.txt -e .
+
+CMD ["uvicorn", "spinta.scripts.spinta_backend:app", "--debug", "--host", "0.0.0.0"]
+
+EXPOSE 8000
```
sirex: I'm not sure, why you need docker for spinta development? I think python virtual env is a much better option.

trimailov: To have a consistent development environment. 

Especially when working from non-linux machine.

sirex:
Probably looks good. But I don't understand, why would you want to create a docker image for app, and install all development dependencies into it? And since you are using it only locally, why you need multi stage build? Multi stage build is good for production images, but I think it is totally not needed for development.

PR #6, opened 2019-04-16
Mongo database backend

sirex:


docker-compose.yml
```diff
@@ -4,24 +4,20 @@ services:
...
-        depends_on:
-            - db
+            - MONGO_INITDB_DATABASE=spinta
+            - MONGO_INITDB_ROOT_USERNAME=admin
+            - MONGO_INITDB_ROOT_PASSWORD=admin123
         volumes:
-            - ".:/opt/spinta"
-        stdin_open: true
-        tty: true
+            - "./mongo-data:/data/db"
```
sirex: Why do you need data directory on host machine? Maybe it is better to let Docker handle it?

spinta/backends/mongo.py
```diff
@@ -0,0 +1,102 @@
...
+    # prepared from manifest declarations using `prepare` command.
+    #
+    # Backend also must have a `transaction` method which must return read or
+    # write transaction object containing an active `connection` to database.
+    metadata = {
+        'name': 'mongo',
+        'properties': {
+            'dsn': {'type': 'string', 'required': True},
+        },
+    }
```
sirex: `metadata` is an leftover from the past and no longer needed or used. You can convert it to:
```suggestion
    schema = {
        'dsn': {'type': 'string', 'required': True},
    }
```

spinta/backends/mongo.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @contextlib.contextmanager
+    def transaction(self, write=False):
+        with self.engine.begin() as connection:
+            if write:
+                # TODO: get a real transaction id
+                transaction_id = 1
```
sirex: If you are not going to implement this now, maybe create a task.

tests/backends/test_mongo.py
```diff
@@ -0,0 +1,23 @@
+def test_schema_loader(context):
```
sirex: It is more push/get test, not schema loader.

spinta/backends/mongo.py
```diff
@@ -46,13 +46,15 @@ def __init__(self, connection, id):
 
 @load.register()
 def load(context: Context, backend: Mongo, config: BackendConfig):
-    # Load Mongo component from configuration.
-    pass
+    # Load Mongo client using configuration.
+    backend.name = config.name
+    backend.db_name = config.db_name
+    backend.engine = pymongo.MongoClient(config.dsn)
```
sirex: Maybe rename `engine` to `mongo`? Engine is more sqlalchemy thing.

spinta/backends/mongo.py
```diff
@@ -81,13 +83,18 @@ def push(context: Context, model: Model, backend: Mongo, data: dict):
...
 
 
 @get.register()
-def get(context: Context, model: Model, backend: Mongo, id: int):
-    # Return single entry by given `id`.
+def get(context: Context, model: Model, backend: Mongo, id: ObjectId):
     transaction = context.get('transaction')
+    db = backend.engine[backend.db_name]
+    model_collection = db[model.get_type_value()]
+    return model_collection.find_one({"_id": ObjectId(id)})
```
sirex: What is `ObjectId`? In Mongo case it is a `str`? Anyway, all ides will have to be unified, instead of using native backend ids, but this is another task.

spinta/backends/mongo.py
```diff
@@ -81,13 +83,18 @@ def push(context: Context, model: Model, backend: Mongo, data: dict):
...
+    data_id = model_collection.insert_one(data).inserted_id
+    return data_id
 
 
 @get.register()
-def get(context: Context, model: Model, backend: Mongo, id: int):
-    # Return single entry by given `id`.
+def get(context: Context, model: Model, backend: Mongo, id: ObjectId):
     transaction = context.get('transaction')
+    db = backend.engine[backend.db_name]
```
sirex: Since we are using same db everywhere, maybe this code should go to `load` part, and `backend.mongo` or `backend.db` would point to `backend.engine[backend.db_nam]`?

tests/conftest.py
```diff
@@ -100,10 +101,12 @@ def context(postgresql):
             'default': {
                 'backend': 'spinta.backends.postgresql:PostgreSQL',
                 'dsn': postgresql,
+                'dbName': 'spinta_test',
```
sirex: Lets be consistent and use `camel_case` everywhere?

    dbName -> db_name

Or even better:

    dbName -> db

sirex: Sorry, this comment was supposed to go for mongo, PostgreSQL does not need this configuration parameter at all. Each backend can have different configuration parameters.

spinta/backends/mongo.py
trimailov: currently without `metadata` I am getting this error:

```
  File "/Users/justas/work/tilaajavastuu/spinta/spinta/testing/pytest.py", line 25, in context
    load_commands(config.get('commands', 'modules', cast=list))
  File "/Users/justas/work/tilaajavastuu/spinta/spinta/utils/commands.py", line 19, in load_commands
    module = importlib.import_module(module_path)
  File "/Users/justas/.pyenv/versions/3.7.1/Python.framework/Versions/3.7/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/justas/work/tilaajavastuu/spinta/spinta/backends/mongo.py", line 13, in <module>
    class Mongo(Backend):
  File "/Users/justas/work/tilaajavastuu/spinta/spinta/types/__init__.py", line 41, in __new__
    assert 'metadata' in attrs
AssertionError
```

It's implemented using `PostgreSQL` backend as example.

spinta/backends/mongo.py
trimailov: Created a task in our internal JIRA, as discussed.

spinta/backends/mongo.py
sirex: Ok, then leave it for now.

tests/backends/test_mongo.py
trimailov: Renamed.

spinta/backends/mongo.py
trimailov: Renamed.

spinta/backends/mongo.py
trimailov: Fixed.

tests/conftest.py
trimailov: ok, will fix this.

spinta/backends/mongo.py
trimailov: Let's have a ticket for this.

docker-compose.yml
trimailov: For persistence between docker containers.

sirex: `- "/data/db"` will do the same, except you don't need to specify a path on host machine.

trimailov:
@sirex - tests are passing. Maybe you want to take another quick look?

Some of your suggestions have been implemented also I've merge new `Config` implementation

trimailov:
.. one more commit was on my local machine..  :)

PR #8, opened 2019-04-17
Configuration refactoring

sirex:
Changed how configuration is loaded and how store, manifests and
backends are loaded from configuration.

Now store, backends and manifests all are loaded from raw configuration
component, each backend can have it's own configuration parameters.

Configuration can be easily overriden in tests and other places.

Consistent bahaviour for setting configuration parameters from env, cli
or directly from python.

### DIFF
--- a/spinta/config/__init__.py
+++ b/spinta/config/__init__.py
@@ -1,5 +1,4 @@
-import copy
-import os
+import collections
 import pathlib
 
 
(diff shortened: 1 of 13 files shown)


spinta/config/__init__.py
```diff
@@ -74,115 +73,106 @@
...
+class Config:
+
+    def __init__(self):
+        self._config = {
+            'cliargs': {},
+            'environ': {},
+            'envfile': {},
+            'default': {},
+        }
+        self.add(CONFIG)
```
trimailov: I do not understand from where this `CONFIG` is coming from? Should it be passed as constructor argument?

spinta/config/__init__.py
```diff
@@ -74,115 +73,106 @@
...
+def _traverse(value, path=()):
     if isinstance(value, dict):
         for k, v in value.items():
-            yield from traverse(v, path + (k,))
+            yield from _traverse(v, path + (k,))
     else:
         yield path, value
+
+
+def _get_inner_keys(value):
```
trimailov: As far as I understand `value` is of type `dict`. Which makes it very confusing when we work with dictionary's keys and _values_.

In my opinion, better names could be `map_value`, `dict_value`, `map_`, `dict_`, etc.

trimailov: To be fair, I do not understand the purpose of this function.

spinta/config/__init__.py
```diff
@@ -74,115 +73,106 @@
...
-    return copy.deepcopy(get_by_path(config, path))
-
-
-def traverse(value, path=()):
+    def keys(self, *key, **kwargs):
+        return self.get(*key, cast=list, **kwargs)
+
+    def _get_config_value(self, key: tuple, default, env):
+
+        # 1. Get value from comman line arguments.
```
trimailov: Typo. `command`

spinta/config/__init__.py
```diff
@@ -74,115 +73,106 @@
...
+
+    def set_env(self, environ):
+        self._config['environ'] = environ
+
+    def get(self, *key, default=None, cast=None, env=True, required=False):
+        value = self._get_config_value(key, default, env)
+
+        if cast is not None:
+            if cast is list and isinstance(value, str):
+                value = value.split(',')
```
trimailov: Hardcoded case when we want to cast string as list.

What if we'll want to split on `.`, `:`, `;`? Or just just convert a string to list `list(value_str)` ?

It requires developers to know implementation details of how `Config.get` works.

spinta/types/config.py
```diff
@@ -1,42 +1,38 @@
...
 from spinta.commands import load
-from spinta.components import Context, Config, BackendConfig
+from spinta.components import Context
+from spinta.config import Config
+from spinta import components
 
 
 @load.register()
-def load(context: Context, config: Config, data: dict) -> Config:
+def load(context: Context, config: components.Config, c: Config) -> Config:
```
trimailov: We have `components.Config` and `config.Config`, this will get confusing very quickly.

As I am not very knowledgable of how `components` module works and is used - I am not able to propose meaningful solution to this confusing naming.

spinta/config/__init__.py
```diff
@@ -74,115 +73,106 @@
...
+
+    def _get_config_value(self, key: tuple, default, env):
+
+        # 1. Get value from comman line arguments.
+        if key in self._config['cliargs']:
+            return self._config['cliargs'][key]
+
+        # Auto-generate environment variable name.
+        if env is True:
+            env = 'SPINTA_' + '_'.join(key).upper()
```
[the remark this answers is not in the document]
sirex: Why you don't like overriding function arguments? Python does it all the time, for example `isinstance(value, type)`, here `type` can be a single type or a tuple, there is no third arguments which tells that `type` is a `tuple`.

So same thing is with `env`. If `env` is True, then it means env name can be used, and the name must be autogenerated.

`isinstance(value, type, multiple=False)` vs `isinstance(value, type)`.

`Config.get(..., env)` vs `Config.get(..., env, autogenerate=True)`.

spinta/config/__init__.py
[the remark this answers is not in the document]
sirex: Added type hint.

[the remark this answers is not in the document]
sirex: It took me 56 lines to explain it, but it's done.

spinta/config/__init__.py
```diff
@@ -74,115 +73,106 @@
...
+                    continue
+                line = line.strip()
+                if line == '':
+                    continue
+                if '=' not in line:
+                    continue
+                name, value = line.split('=', 1)
+                self._config['envfile'][name] = value
+
+    def set_env(self, environ):
```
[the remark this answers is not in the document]
sirex: In this case, environment variables are not added, but are set. All other `add*` methods do add things on top of existing config vars. But in environment case, I don't want to make copy of `os.environ`, so instead of modifying it in place, I just add it.

So naming here is correct, because it reflects real behavior which is different from other `add*` methods.

spinta/config/__init__.py
sirex: This should be explained in documentation. Here, in order to pass a list via environment variables, you need to separate them with commas. That just a requirement.

spinta/types/config.py
sirex: Yes, probably name should be changed. Maybe `spinta.config.Config` could be renamed to `RawConfig`.

spinta/config/__init__.py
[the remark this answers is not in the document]
trimailov: agreed.

spinta/types/config.py
trimailov: Maybe `AppConfig` sounds better? Though does it fit the semantics properly?

sirex:
> P.P.S Have you tried to look into https://github.com/willkg/everett ? I know that it's flexible and should be stable.

Yes I know about everett and I don't like it, it seems too complicated. Personally I think [DynaConf](https://dynaconf.readthedocs.io/) is a better alternative. But decided to go with my own configuration implementation, just to make sure, that I will not end up fighting with limitations. And anyway, whole code for configuration management is less than 100 lines of code.

sirex:
Merging this, my code reviews fixes goes into another pull request, which is already in progress.

trimailov:
> > P.P.S Have you tried to look into https://github.com/willkg/everett ? I know that it's flexible and should be stable.
> 
> Yes I know about everett and I don't like it, it seems too complicated. Personally I think [DynaConf](https://dynaconf.readthedocs.io/) is a better alternative. But decided to go with my own configuration implementation, just to make sure, that I will not end up fighting with limitations. And anyway, whole code for configuration management is less than 100 lines of code.

sounds reasonable. 

if dependencies creates limitations then of course it's better to use what fits, especially if rolling your own config is not too hard.

spinta/types/config.py
sirex: `spinta.config.Config` is just a small wrapper on top of raw config dicts, env files and env vars.

`spinta.components.Config` uses `spinta.config.Config` to set attributes to a config objects that will be used in all other places in app. So this on sounds more like an `AppConfig`, and the other one like `RawConfig`.

PR #9, opened 2019-04-18
Configuration improvements

sirex:
Many things are fixed here:

- Add support for nested properties. This is no a full support, but just enough to make sure, that manifest file loads without errors.
- Add environments to configuration. This will help to reduce configuration diversification, basically we should have same and non changing configuration in one place and change only small and necesarry bits.
- Replace hardcoded environment variables in tests, use same configuration as everywhere else.
- Add new `wait` command, to wait for backend.
- Removed hardcoded use of `default` backend, always take backend from node or manifest.
- Address some issues from previous code review.
- Fix node inheritance issue.
- Fix backend inheritance.
- Fix check for unknown node attributes, in order to check that both property and property type schemas must be checked.
- Replace all add/set methods in Config to single read method.


sirex:
Thanks for the fix! :heart: :beer:

PR #11, opened 2019-04-25
Create `prepare` commands for parsing urls and creating responses

trimailov:
Created 2 new components: `spinta.urlparams:UrlParams`, `spinta.urlresponse:UrlResponse`.
These 2 components have `prepare` commands respectively.

This will let `spinta` to be more flexibly used as a framework. Client application would
need to  create subclasses for `UrlParams` and `UrlResponse`, with their custom url parsing
and response logic.

IMHO, this can be used as a foundation to further work on request authorization
and other features (as well as clients, could easily implement custom logic on top).

This is a first mockup, thus highly likely to be changed and/or refactored.


trimailov:
@sirex : I am not particularly happy how this PR turned out to be.

There may be not great decisions starting from naming and ending with interfaces.

Also I've ignored UrlParams version configuration for now (will it be needed at all?)

But it's still a foundation to continue working with authentication implementation on a client.

PR #13, opened 2019-04-29
Almost working /auth/token endpoint

sirex:
Implemented /auth/token endpoint, using client_credentials grant type,
with JWT token as access token. Everytihg is built using authlib.

Also there is a command `spinta genkeys path/to/keys/dir` to generate
private and public keys.


spinta/testing/pytest.py
```diff
@@ -61,6 +64,8 @@ def context(config, postgresql, mongo):
     prepare(context, store)
     migrate(context, store)
 
+    context.bind('auth', AuthorizationServer, context)
```
trimailov: Do I understand correctly, that `AuthorizationServer` is bind only in tests, but not in the server?

Tests are passing, but trying to request local server at `/auth/token` throws this error:

```python
> curl -X POST localhost:8000/auth/token
  File "/Users/justas/work/tilaajavastuu/spinta/env/lib/python3.7/site-packages/starlette/middleware/errors.py", line 103, in asgi
    await asgi(receive, _send)
  File "/Users/justas/work/tilaajavastuu/spinta/env/lib/python3.7/site-packages/starlette/exceptions.py", line 74, in app
    raise exc from None
  File "/Users/justas/work/tilaajavastuu/spinta/env/lib/python3.7/site-packages/starlette/exceptions.py", line 63, in app
    await instance(receive, sender)
  File "/Users/justas/work/tilaajavastuu/spinta/env/lib/python3.7/site-packages/starlette/routing.py", line 41, in awaitable
    response = await func(request)
  File "./spinta/api.py", line 30, in auth_token
    auth = context.get('auth')
  File "./spinta/components.py", line 26, in get
    raise Exception(f"Unknown context variable {name!r}.")
```

sirex: Yes, this is the reason, why it is almost done.

PR #14, opened 2019-04-30
Add checks for valid `create` requests

trimailov:
_NOTE: as discussed this  is implemented within `spinta` and not in the client. It's implemented in straightforward, but ugly way until refactoring pattern will emerge._

As per ticket:
- [X] 201 (Created) - if the posted content has been stored
- [X] 400 (Content is not valid JSON) - if the content is not valid JSON
- [X] 400 (New item has id already set) - if the content contains id without appropriate scope
- [X] 400 (New item has revision already set) - if the content contains revision
- [X] 400 (New item has conflicting id) - if the caller has permission to set the id, but it exists already in the data storage:
- [X] 401 (Authorization header is missing) - if there is no Authorization header in the request
- [ ] 401 (Access token is invalid: <token error>) - if validation of the access token fails in any way (missing scope, wrong signature etc.)
- [X] 403 (Authorization header is in invalid format, should be "Bearer TOKEN") - if the header is in any other format
- [X] 415 (Unsupported media type) - if the content is not specified as JSON

One point (`401 (Access token is invalid: <token error>)`) is not implemented and parts related to scopes are mocked for
development and testing purposes, though should be implemented easily, when scopes and authorization is implemented.

Related: SPLAT-12

### DIFF
--- a/spinta/backends/postgresql/__init__.py
+++ b/spinta/backends/postgresql/__init__.py
@@ -11,12 +11,13 @@ from sqlalchemy.dialects.postgresql import JSONB, BIGINT
 from sqlalchemy.ext.compiler import compiles
 from sqlalchemy.sql.expression import FunctionElement
 
-from spinta.types import NA
+from spinta.backends import Backend
 from spinta.commands import wait, load, prepare, migrate, check, push, get, getall, wipe
 from spinta.components import Context, Manifest, Model, Property
-from spinta.backends import Backend
-from spinta.types.type import Type
 from spinta.config import Config
+from spinta.types import NA
+from spinta.types.type import Type
+from spinta.utils.exceptions import MultipleRowsException, NoResultsException
 
 # Maximum length for PostgreSQL identifiers (e.g. table names, column names,
 # function names).
@@ -82,12 +83,12 @@ class PostgreSQL(Backend):
 
         elif len(result) == 0:
             if default is NA:
-                raise Exception("No results where found.")
+                raise NoResultsException("No results where found.")
             else:
                 return default
 
         else:
-            raise Exception("Multiple rows were found.")
+            raise MultipleRowsException("Multiple rows were found.")
 
 
 class ReadTransaction:
--- a/spinta/utils/response.py
+++ b/spinta/utils/response.py
@@ -112,19 +112,23 @@ async def create_http_response(context, params, request):
             # only `application/json` is supported
             ct = request.headers.get('content-type')
             if ct != 'application/json':
-                return JSONResponse(
-                    {"error": "only 'application/json' content-type is supported"},
-                    status_code=415
+                raise HTTPException(
+                    status_code=415,
+                    detail="only 'application/json' content-type is supported",
                 )
 
             auth_header = request.headers.get('authorization')
             if auth_header is None:
-                return JSONResponse({"error": "missing authorization header"},
-                                    status_code=401)
+                raise HTTPException(
+                    status_code=401,
+                    detail="missing authorization header",
+                )
 
             if not auth_header.startswith('Bearer '):
-                return JSONResponse({"error": "invalid authorization header"},
-                                    status_code=403)
+                raise HTTPException(
+                    status_code=403,
+                    detail="invalid authorization header",
+                )
 
             jwt_token = auth_header.strip('Bearer ') if auth_header else None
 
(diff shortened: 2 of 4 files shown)


trimailov:
Only now understood that I can use this for showing errors: https://github.com/atviriduomenys/spinta/blob/master/spinta/api.py#L52

spinta/backends/postgresql/__init__.py
```diff
@@ -11,12 +11,13 @@
...
-from spinta.types import NA
+from spinta.backends import Backend
 from spinta.commands import wait, load, prepare, migrate, check, push, get, getall, wipe
 from spinta.components import Context, Manifest, Model, Property
-from spinta.backends import Backend
-from spinta.types.type import Type
 from spinta.config import Config
+from spinta.types import NA
+from spinta.types.type import Type
+from spinta.utils.exceptions import MultipleRowsException, NoResultsException
```
sirex: There is `sptina.exceptions` module already.

spinta/backends/postgresql/__init__.py
```diff
@@ -82,12 +83,12 @@ def get(self, connection, columns, condition, default=NA):
 
         elif len(result) == 0:
             if default is NA:
-                raise Exception("No results where found.")
+                raise NoResultsException("No results where found.")
```
sirex: There is exception called `NotFound` already.

spinta/utils/response.py
```diff
@@ -106,7 +108,58 @@
...
             context.bind('transaction', manifest.backend.transaction, write=True)
-            data = await request.json()
+
+            # only `application/json` is supported
+            ct = request.headers.get('content-type')
+            if ct != 'application/json':
+                return JSONResponse(
+                    {"error": "only 'application/json' content-type is supported"},
+                    status_code=415
+                )
```
sirex: You shoul dalways raise an `HTTPError` and response will be handled by http exception handler.

sirex: Since we don't want to be too dependent on web framework, we should have our on set of http exceptions in `spinta.exceptinos`.

spinta/utils/response.py
```diff
@@ -106,7 +108,58 @@
...
+            auth_header = request.headers.get('authorization')
+            if auth_header is None:
+                return JSONResponse({"error": "missing authorization header"},
+                                    status_code=401)
+
+            if not auth_header.startswith('Bearer '):
+                return JSONResponse({"error": "invalid authorization header"},
+                                    status_code=403)
+
+            jwt_token = auth_header.strip('Bearer ') if auth_header else None
```
sirex: That is not part of you task. I will rewrite this anyway using authlib. You can leave it as is, and leave everything else related to auth to me, because I'm currently working on it.

trimailov:
> Only now understood that I can use this for showing errors: https://github.com/atviriduomenys/spinta/blob/master/spinta/api.py#L52
> 
> @sirex - should I refactor? or ignore this for a bit, due to many changes incoming and eventual refactoring?

Fixed this.

Though in one place it did not work, but I did not bother to dig deeper: https://github.com/atviriduomenys/spinta/pull/14/commits/32bb5b7960ab137b970b584467bcfd8c6aa7f244#diff-764fb8a1cff4bc5eafcb8ea4c0be7c80R168

spinta/utils/response.py
```diff
@@ -106,7 +108,58 @@
...
+                    return JSONResponse({"error": "cannot create 'id'"},
+                                        status_code=400)
+                else:
+                    try:
+                        id = commands.get(context, model, model.backend, data['id'])
+                    except (MultipleRowsException, NoResultsException):
+                        pass
+                    else:
+                        return JSONResponse({"error": "cannot create duplicate 'id'"},
+                                            status_code=400)
```
sirex: Wrong error message, and you don't need to check that here, this is already checked by `check` command in `spinta.backends.postgresql.check`.

spinta/utils/response.py
```diff
@@ -106,7 +108,58 @@
...
+
+            if 'revision' in data.keys():
+                return JSONResponse({"error": "cannot create 'revision'"},
+                                    status_code=400)
```
sirex: This also should go to `check` command.

spinta/utils/response.py
```diff
@@ -106,7 +108,73 @@
...
+                        commands.get(context, model, model.backend, data['id'])
+                    except (MultipleRowsException, NoResultsException):
+                        pass
+                    else:
+                        # XXX: should raise HTTPException here,
+                        # but for some reason in tests, when raising
+                        # HTTPException here, the database connection
+                        # closes sooner than the test can finish,
+                        # so I return just a JSONResponse with HTTP/400
+                        return JSONResponse({"error": "cannot create duplicate 'id'"},
```
sirex: `raise` instead of `return`?

PR #16, opened 2019-04-30
Oauth client configuration

sirex:
Added a smple, file based oauth client configuraiton. In order to create
new client you need to run this command:

    spinta client add

It will add a new client to `config.config_path / 'clients'` directory
as YAML file. Client secret will be shown in output and an ecrypted
value saved in YAML file.

Once can edit those YAML files to configure client scopes.

`/auth/token` endpoint now does a read client_credentials authorization
using those YAML files.

Authorization server public/private keys were moved to
`config.config_path / 'keys'` directory. `config.config_path` should be
a configuration directory where all kinds of configuration fiels are
stored.

All this client configuration should be optional and an external
authorization server can be used, but this is not yet supported.

All fuctionalite works only tests, I did not tired to run it manually
yet, because funcitonality is not yet complete, authorization tokens are
not yet integrated into the whole request handling process.


spinta/utils/passwords.py
```diff
@@ -0,0 +1,52 @@
...
+    value = value.ljust(math.ceil(len(value) / 4) * 4, '=')
+    return base64.urlsafe_b64decode(value)
+
+
+def crypt(password, iterations=None, salt=None):
+    if iterations is None:
+        iterations = random.randint(100_000, 999_999)
+    if salt is None:
+        salt = os.urandom(16)
+    elif isinstance(salt, str):
```
trimailov: should we check that `salt` is `bytes` , if won't be `None` or `str`?

spinta/utils/passwords.py
```diff
@@ -0,0 +1,52 @@
...
+        salt = salt.encode()
+    if isinstance(password, str):
```
trimailov: same for password

spinta/utils/passwords.py
```diff
@@ -0,0 +1,52 @@
+import base64
+import hashlib
+import hmac
+import math
+import os
+import random
+
+
+def _bencode(value: bytes) -> str:
+    return base64.urlsafe_b64encode(value).decode().rstrip('=')
```
trimailov: Why are we removing '=' characters at the end of the string?

That will explain re-adding them in line 14

spinta/utils/passwords.py
sirex: Python uses duck typing, that means, it does not matter if salt is bytes or not as long as it behaves like bytes.

spinta/utils/passwords.py
sirex: Mainly because of esthetic reasons and because python does not implement full base64 standard. Padding in base64 URL safe encoding is not required if there is no need to concatenate multiple base64 encoded strings. So in this case password hashes will never be concatenated.

https://tools.ietf.org/html/rfc4648#section-5

PR #19, opened 2019-05-06
WIP: Data validation before creation

trimailov:
WIP for data validation, for simple flat case (no nesting).

WIP, because not finished on validation implementation and tests.

### DIFF
--- a/spinta/types/model.py
+++ b/spinta/types/model.py
@@ -69,6 +69,15 @@ def check(context: Context, model: Model):
         context.deprecation("`id` property must specify real type like 'string' or 'integer'. Use of 'pk' is deprecated.")
 
 
+@check.register()
+def check(context: Context, model: Model, data: dict):
+    for name, prop in model.properties.items():
+        if name in data:
+            data_value = data[name]
+            if not prop.type.is_valid(data_value):
+                context.error(f"{data_value} is not valid type: {prop.type}")
+
+
 @error.register()
 def error(exc: Exception, context: Context, model: Model):
     message = (
--- a/spinta/types/model.py
+++ b/spinta/types/model.py
@@ -75,7 +75,7 @@ def check(context: Context, model: Model, data: dict):
         if name in data:
             data_value = data[name]
             if not prop.type.is_valid(data_value):
-                context.error(f"{data_value} is not valid type: {prop.type}")
+                raise Exception(f"{data_value} is not valid type: {prop.type}")
 
 
 @error.register()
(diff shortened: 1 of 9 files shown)


spinta/types/model.py
```diff
@@ -69,6 +69,15 @@ def check(context: Context, model: Model):
         context.deprecation("`id` property must specify real type like 'string' or 'integer'. Use of 'pk' is deprecated.")
 
 
+@check.register()
+def check(context: Context, model: Model, data: dict):
+    for name, prop in model.properties.items():
+        if name in data:
+            data_value = data[name]
+            if not prop.type.is_valid(data_value):
+                context.error(f"{data_value} is not valid type: {prop.type}")
```
sirex: `context.error` is an old leftover, don't use it. Probably here it is better to raise `ValidationError` or something like that, then you should handle these check errors and return `400` response.

spinta/types/type.py
```diff
@@ -16,27 +18,64 @@ class Type:
...
+            except ValueError:
+                return False
+            else:
+                return True
+
+
+
+class DateTime(Date):
+
+    DEFAULT_FMT = "%Y-%m-%d %H:%M:%S"
```
sirex: Maybe use `fromisoformat`?

spinta/types/type.py
```diff
@@ -16,27 +18,64 @@ class Type:
...
 
 class Integer(Type):
-    pass
+
+    def is_valid(self, value):
+        if isinstance(value, int):
+            return True
+        else:
+            try:
+                int(value)
```
sirex: Maybe add two step validation, using `load` and `check`? `load` would make sure, that value is loaded from a primitive value to a python native value, and then `check` would assume, that all types are already correct, and you only need to check if value is correct.

You will have to convert values to python-native values anyway.

Also a third step could be needed `prepare`, before saving data to database, python native values must be converted to a database native types. But maybe that is for another pull request.

trimailov: Do you mean to have something like:

```
data = load(context, model, data)
check(context,model, data)
check(context, model, model.backend, data)
```

in here: https://github.com/atviriduomenys/spinta/pull/19/files#diff-11540b6ded55485b62d4f20cf795971cR95

sirex: Yes, something like that. First data should be `load`ed, going down all the tree and then loaded data should be passed to `check`.

trimailov: Will refactor this with next PR.

Will merge as is (tests are passing).

spinta/types/type.py
trimailov: Will do on next PR.

PR #22, opened 2019-05-08
Remove context block

trimailov:
As we return async response - bound context may be gone, before full
response is returned.

### DIFF
--- a/spinta/api.py
+++ b/spinta/api.py
@@ -6,9 +6,11 @@ from authlib.common.errors import AuthlibHTTPError
 
 from starlette.applications import Starlette
 from starlette.exceptions import HTTPException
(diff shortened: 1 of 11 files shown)


spinta/api.py
```diff
@@ -28,6 +30,71 @@
...
+                assert message["type"] == "http.response.body"
+                yield message["body"]
+            task.result()
+
+        response = StreamingResponse(
+            status_code=message["status"], content=body_stream()
+        )
+        response.raw_headers = message["headers"]
+        return response
+    # -------------------------->8--------------------------
```
sirex: A workaround for https://github.com/encode/starlette/issues/472

PR #24, opened 2019-05-13
Data validation for non-nested reports

trimailov:
Basic implementation for converting pushed data to native python types.

WIP: because we need `prepare` command , which would be able to convert data from native python datatypes, to types which can be consumed by data stores.


spinta/types/model.py
```diff
@@ -47,6 +47,16 @@ def load(context: Context, prop: Property, data: dict, manifest: Manifest) -> Pr
     return prop
 
 
+@load.register()
+def load(context: Context, model: Model, data: dict) -> dict:
+    for name, prop in model.properties.items():
+        if name in data:
+            data_value = data[name]
+            # XXX: rewrite into command instead of type method?
```
sirex: It is OK to have methods on component classes as long, as those methods stay within component domain. If you need to tho something more, that is out of component domain, then you need to use commands or functions.

spinta/types/type.py
```diff
@@ -18,63 +18,69 @@ class Type:
         'link': {'type': 'string'},
     }
 
+    def load(self, value):
```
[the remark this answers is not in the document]
trimailov: I am not sure how I should annotate type, when I do not know what value will be given?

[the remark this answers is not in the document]
sirex: `typing.Any`.

tests/test_store.py
```diff
@@ -68,33 +68,51 @@ def test_nested(context):
...
 
 
 def test_invalid_report(context, app):
+    app.authorize([
+        'spinta_report_insert',
+    ])
+
     with pytest.raises(ValueError):
-        list(context.push([{
+        # FIXME: instead of raising an error should return HTTP/4XX code?
```
sirex: Yes, HTTP status code 400 should be returned instead of 500.

trimailov:
Currently there should be load/check/prepare steps for POST /report

trimailov:
.. though `pdf` property is still not tested :/

spinta/types/model.py
```diff
@@ -64,6 +76,30 @@ def check(context: Context, model: Model, data: dict):
...
+    return data
+
+
+@prepare.register()
+def prepare(context: Context, model: Model, backend: Mongo, data: dict) -> dict:
+    for name, prop in model.properties.items():
+        if name in data:
+            data_value = data[name]
+            data[name] = prop.type.prepare_for_mongo(data_value)
+    return data
```
sirex: These two methods should go to `sptinta/backends/...`.

spinta/types/type.py
```diff
@@ -18,63 +20,84 @@ class Type:
...
+        # prepares value for Mongo store
+        # for simple types - loaded native values should work
+        # otherwise - override for this method is necessary
+        return value
+
+    def prepare_for_postgres(self, value: typing.Any):
+        # prepares value for PostgreSQL store
+        # for simple types - loaded native values should work
+        # otherwise - override for this method is necessary
+        return value
```
sirex: These two methods should be replaced by commands and should go to `spinta/backends/...`.

spinta/types/type.py
```diff
@@ -18,63 +20,84 @@ class Type:
...
+        return date.fromisoformat(value)
+
+    def is_valid(self, value: date):
+        # self.load() ensures value is native `datetime` type
+        return True
+
+    def prepare_for_mongo(self, value: date) -> datetime:
+        # `pymongo` expects datetime objects for dates
+        # https://api.mongodb.com/python/current/api/bson/timestamp.html
+        return datetime(value.year, value.month, value.day)
```
sirex: Move this to a command in `sptinta/backends/mongo.py`.

tests/test_store.py
trimailov: I think I will work on this on other sub-task. Ok?

sirex: Ok :+1:

spinta/types/model.py
```diff
@@ -76,7 +76,12 @@ def check(context: Context, model: Model, data: dict):
 
 @prepare.register()
 def prepare(context: Context, model: Model, data: dict) -> dict:
-    data = prepare(context, model, model.backend, data)
+    # prepares model's data for storing in Mongo
+    backend = model.backend
+    for name, prop in model.properties.items():
+        if name in data:
+            data_value = data[name]
+            data[name] = prepare(context, prop.type, backend, data_value)
```
sirex: `prepare` should be recursive, so you need to define two more commands for objects and arrays.

trimailov: I'll do it with a separate ticket.

spinta/backends/__init__.py
```diff
@@ -15,6 +15,14 @@ def transaction(self):
         raise NotImplementedError
 
 
+@prepare.register()
```
sirex: I think, you forgot to import `prepare`.

spinta/backends/__init__.py
```diff
@@ -1,11 +1,12 @@
...
 
-from spinta.types import Type
+from spinta.types import Type as MetaType
+from spinta.types.type import Type
 from spinta.components import Context, Model
 from spinta.commands import error, prepare
 
 
-class Backend(Type):
+class Backend(MetaType):
```
sirex: `spinta.types.Type` is deprecated, please remove it and leave just `Backend:`.

trimailov: then remove `error(Exception, Context, Backend, MetaType, **kwargs)` command too?

spinta/backends/__init__.py
```diff
@@ -24,7 +25,7 @@ def prepare(context: Context, type: Type, backend: Backend, value: object) -> ob
 
 
 @error.register()
-def error(exc: Exception, context: Context, backend: Backend, type: Type, **kwargs):
+def error(exc: Exception, context: Context, backend: Backend, type: MetaType, **kwargs):
```
sirex: I think, here correct signature should be `type: Type`, because there is no such thing as `(Context, Backend, spinta.types.Type)`.

spinta/backends/__init__.py
sirex: Read my next comment about `error()`.

spinta/backends/__init__.py
trimailov: Though `spinta.types.Type` itself shouldn't be removed?

sirex: Yes, it can be removed, if no one else is using it.

trimailov: It's used here: https://github.com/atviriduomenys/spinta/blob/master/spinta/utils/response.py#L388

sirex: There is should also be `spinta.types.type.Type`.

trimailov:
@sirex: any more comments?

I'll do nested load/check/prepare with a separate PR.

PR #25, opened 2019-05-14
Return HTTP/400 on data load exception

trimailov:


spinta/types/type.py
```diff
@@ -33,7 +34,12 @@ def is_valid(self, value: typing.Any):
 class PrimaryKey(Type):
 
     def load(self, value: typing.Any):
-        return str(value)
+        try:
+            # is it possible to get ValueError when doing string conversion?
+            return str(value)
```
trimailov: @sirex  - isn't it that with `str`  conversion you cannot get an error?

spinta/types/type.py
```diff
@@ -33,7 +34,12 @@ def is_valid(self, value: typing.Any):
...
+        except ValueError:
+            raise DataError(f'Invalid value for pk type: {value}')
```
sirex: I don't think it is possible to get ValueError from `str(value)`, so I would add error handlers only for code proven with tests.

spinta/types/type.py
```diff
@@ -143,7 +161,7 @@ def load(self, value: typing.Any):
             # if value is list - return it
             return value
         else:
-            raise ValueError
+            raise DataError(f'Invalid value for array type: {value}')
```
sirex: I would also add `{value!r}`.

tests/test_store.py
```diff
@@ -103,16 +103,73 @@ def test_report(context, app):
...
+    resp = app.post('/report', json={
+        'type': 'report',
+        'report_type': 'simple',
+        'status': 'valid',
+        'valid_from_date': '2019-04',  # invalid conversion to date
+    })
+
+    assert resp.status_code == 400
+    assert resp.json() == {
+        "error": "Invalid value for date type: 2019-04",
```
sirex: It would be useful if field name would be included in error message.

sirex: Also error should tell what is expected, now it is clear that value is incorrect, but it is not clear what date format is correct.

spinta/types/type.py
```diff
@@ -43,7 +44,10 @@ def is_valid(self, value):
 class Date(Type):
 
     def load(self, value: typing.Any):
-        return date.fromisoformat(value)
+        try:
+            return date.fromisoformat(value)
+        except ValueError:
+            raise DataError(f"Expected date value in ISO 8601 format ('YYYY-MM-DD'), got {value!r}")
```
sirex: By the way, what error message is given by Python? Maybe it would be even better to use error message generated by Python standard library?

```python
except ValueError as e:
    raise DataError(f"Invalid date value: {e}.")
```

trimailov:
```
In [1]: import datetime

In [2]: datetime.datetime.fromisoformat('2019-04')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-03964cd7b0d8> in <module>
----> 1 datetime.datetime.fromisoformat('2019-04')

ValueError: Invalid isoformat string: '2019-04'
```

It's concise and not as explicit, but maybe good enough?

trimailov:
> Looks good. I just figured out, that maybe we don't need to obfuscate original error messages, by given our own messages? Maybe it is even better to show original message?

Problem is that, as we spoke before, for some other types, values are exposed.

```
In [4]: int('p4ssw0rd')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-4-295c079b9c41> in <module>
----> 1 int('foobar')

ValueError: invalid literal for int() with base 10: 'p4ssw0rd'
```

spinta/types/type.py
sirex: I think it is good enough. Maybe there are other cases, where different error messages are given. So probably using original error messages will be more informative.

sirex:
> > Looks good. I just figured out, that maybe we don't need to obfuscate original error messages, by given our own messages? Maybe it is even better to show original message?
> 
> Problem is that, as we spoke before, for some other types, values are exposed.
> 
> ```
> In [4]: int('p4ssw0rd')
> ---------------------------------------------------------------------------
> ValueError                                Traceback (most recent call last)
> <ipython-input-4-295c079b9c41> in <module>
> ----> 1 int('foobar')
> 
> ValueError: invalid literal for int() with base 10: 'p4ssw0rd'
> ```

I think, we should have a special type or a type parameter, that would specify if value is sensitive or not. Because integers usually are not sensitive. Strings might be sensitive.

PR #26, opened 2019-05-15
Load nested structures into python native types

trimailov:
Load array items and object properties.

Also fix issue with `Date` and `DateTime` raising `TypeError` (not only `ValueError`)


spinta/types/type.py
```diff
@@ -235,6 +241,33 @@ def load_type(context: Context, prop: Node, data: dict, manifest: Manifest):
...
+
+
+@load.register()
+def load(context: Context, type: Array, value: object) -> object:
+    # loads value into native python list, including all list items
+    array_item_type = type.items.type
+    loaded_array = type.load(value)
+    for j, item in enumerate(loaded_array):
+        # overwrite loaded list, with items loaded to native python types
+        loaded_array[j] = load(context, array_item_type, item)
```
sirex: I'm not sure if this is a good idea, to update input value in place. Lets append loaded values to a new array and leave input array as is.

spinta/types/type.py
```diff
@@ -235,6 +241,33 @@ def load_type(context: Context, prop: Node, data: dict, manifest: Manifest):
...
+    return loaded_array
+
+
+@load.register()
+def load(context: Context, type: Object, value: object) -> object:
+    # loads value into native python dict, including all dict's items
+    loaded_obj_value = type.load(value)
+    for name, value in type.properties.items():
+        # overwrite given object key value, with a loaded python native key value
+        loaded_obj_value[name] = load(context, value.type, loaded_obj_value[name])
```
sirex: Same with dict, lets avoid mutating input values.

trimailov:
I've realised `prepare` is not tested for nested structures.

Thus pulled back status to Work in Progress, though @sirex  feel free to look around what is now implemented.

trimailov:
@sirex: finished nested prepare.

spinta/backends/__init__.py
```diff
@@ -23,6 +23,25 @@ def prepare(context: Context, type: Type, backend: Backend, value: object) -> ob
     return value
 
 
+@prepare.register()
+def prepare(context: Context, type: Array, backend: Backend, value: object) -> object:
```
sirex: Incorrect type anotations, here you clearly expect `value` to be array and also returning `array`.

spinta/backends/__init__.py
```diff
@@ -23,6 +23,25 @@ def prepare(context: Context, type: Type, backend: Backend, value: object) -> ob
...
+    # prepare array and it's items for datastore
+    array_item_type = type.items.type
+    new_prepare_array = []
+    for item in value:
+        new_prepare_array.append(prepare(context, array_item_type, backend, item))
+    return new_prepare_array
```
sirex: It is always more difficult to read too verbose code with lots of unnecessary operations. You can do the same with this:

```python
return [prepare(context, type.items.type, backend, v) for v in value]
```

At least to my eye, it is very clear, that each item of `value` is mapped to `prepare` function call.

In you case, I see new variables introduced, `array_item_type` is used just once, also there are three operations with `new_prepare_array`. It just takes more time to undrestand.

spinta/backends/__init__.py
```diff
@@ -23,6 +23,25 @@ def prepare(context: Context, type: Type, backend: Backend, value: object) -> ob
...
+
+
+@prepare.register()
+def prepare(context: Context, type: Object, backend: Backend, value: object) -> object:
```
sirex: `value: dict) -> dict`.

spinta/types/type.py
```diff
@@ -235,6 +241,33 @@ def load_type(context: Context, prop: Node, data: dict, manifest: Manifest):
...
 
 
+@load.register()
+def load(context: Context, type: Type, value: object) -> object:
+    # loads value to python native value according to given type
+    return type.load(value)
+
+
+@load.register()
+def load(context: Context, type: Array, value: object) -> object:
```
sirex: `-> list`.

spinta/types/type.py
```diff
@@ -235,6 +241,33 @@ def load_type(context: Context, prop: Node, data: dict, manifest: Manifest):
...
+    array_item_type = type.items.type
+    loaded_array = type.load(value)
+    new_loaded_array = []
+    for item in loaded_array:
+        new_loaded_array.append(load(context, array_item_type, item))
+    return new_loaded_array
+
+
+@load.register()
+def load(context: Context, type: Object, value: object) -> object:
```
sirex: `-> dict`.

spinta/types/type.py
```diff
@@ -235,6 +241,33 @@ def load_type(context: Context, prop: Node, data: dict, manifest: Manifest):
...
+    # loads value into native python dict, including all dict's items
+    loaded_obj = type.load(value)
+    new_loaded_obj = {}
+    for k, v in type.properties.items():
+        new_loaded_obj[k] = load(context, v.type, loaded_obj[k])
```
sirex: Keep in mind, that here value is at the very early stage and `k` might not be present in `loaded_obj`, but because of that loading should not raise an error. You should only load given items and make sure, that there are no unknown (not in `type.properties`) items.

tests/test_validation.py
```diff
@@ -0,0 +1,184 @@
...
+    resp = app.post('/report', json={
+        'type': 'report',
+        'report_type': 'simple',
+        'status': 'valid',
+        'count': 'c0unt',  # invalid conversion to int
+    })
+
+    assert resp.status_code == 400
+    assert resp.json() == {
+        "error": "invalid literal for int() with base 10: 'c0unt'",
```
sirex: Probably this is for another pull request, but here I would like to see field name, otherwise these error messages could be really difficult to understand and fix.

tests/test_validation.py
```diff
@@ -0,0 +1,184 @@
...
+        'update_time': '2019-04',  # invalid conversion to datetime
+    })
+
+    assert resp.status_code == 400
+    assert resp.json() == {
+        "error": "Invalid isoformat string: '2019-04'",
+    }
+
+
+def test_non_string_report_datetime(context, app):
```
sirex: Here and in almost all other tests, you don't need `context`.

tests/test_validation.py
```diff
@@ -0,0 +1,184 @@
...
+    resp = app.post('/nested', json={
+        'some': [{
+            'nested': 'object' # invalid object property type
+        }]
+    })
+
+    assert resp.status_code == 400
+    assert resp.json() == {
+        "error": "Invalid object type: <class 'str'>",
+    }
```
sirex: I would like to see a test, where you pass unknown properties. Not sure if checks for required properties, that would be good to test too.

sirex: Also, add tests when nested dict does not have a known property.

sirex:
> @sirex: finished nested prepare.

Please remove WIP if you done with this pull request.

tests/test_validation.py
trimailov: This is indeed for later.

tests/test_validation.py
trimailov: > Not sure if checks for required properties, that would be good to test too.

I have not done any logic to check for property's properties such as required.

trimailov:
@sirex - any more comments? 

- fixed type annotations
- added check for unknown properties in model load and object type load
- no fail when properties are missing in an object

spinta/backends/__init__.py
```diff
@@ -34,7 +34,9 @@ def prepare(context: Context, type: Object, backend: Backend, value: dict) -> di
     # prepare objects and it's properties for datastore
     new_loaded_obj = {}
     for k, v in type.properties.items():
-        new_loaded_obj[k] = prepare(context, v.type, backend, value[k])
+        # only load value keys which are available in schema
+        if k in value:
+            new_loaded_obj[k] = prepare(context, v.type, backend, value[k])
```
sirex: If `k` is not in `value`, then value should be `NA`. Now if you skip a property entirely, `v.type` does have a chance to prepare NA values. You can add a fixme comment and some one will fix it later.

tests/test_validation.py
```diff
@@ -182,3 +182,76 @@ def test_invalid_nested_object_property(app):
...
+
+    resp = app.post('/report', json={
+        'type': 'report',
+        'report_type': 'simple',
+        'status': 'valid',
+        'count': '42',
+        'valid_from_date': '2019-04-20',
+        'update_time': '2019-04-20 03:14:15',
+        'notes': [{
+            'note': 'hello report',  # missing object properties
```
sirex: I think note is in schema.

```yaml
  notes: {
      type: array,
      items: {
          type: object,
          properties: {
              note: {type: string},
              note_type: {type: string}
          }
      }
  }
```

sirex: Sorry, ignore this comment.

trimailov: Yes, and test does not throw any error that `note_type` is missing, though as mentioned above `FIXME` is added to load missing values as `NA`.

spinta/backends/__init__.py
sirex: Also, I think, `prepare` will always have all properties, so probably there is no need to check if `k` is in `value` at all.

PR #27, opened 2019-05-20
Add FIXME comment

trimailov:


spinta/backends/__init__.py
```diff
@@ -35,6 +35,7 @@ def prepare(context: Context, type: Object, backend: Backend, value: dict) -> di
     new_loaded_obj = {}
     for k, v in type.properties.items():
         # only load value keys which are available in schema
+        # FIXME: If k is not in value, then value should be NA. Do not skip a value
         if k in value:
```
sirex: I would remove this condition, if tests pass. In `prepare` we convert correct python-native types to types accepted by database. That means, that all values from all properties must be in `value`. It a property is not in `value`, then the error is somewhere else.

trimailov: Created a ticket for this.

PR #28, opened 2019-05-21
Qvarn revision compatibility

trimailov:
Qvarn compatible `revision` property on data.

### DIFF
--- a/spinta/backends/mongo.py
+++ b/spinta/backends/mongo.py
@@ -109,13 +109,6 @@ def push(context: Context, model: Model, backend: Mongo, data: dict, *, action: 
     transaction = context.get('transaction')
     model_collection = backend.db[model.get_type_value()]
 
-    # FIXME: hack to remove built-in properties
-    # if data['id'] is None:
-    #     del data['id']
-    # if data['revision'] is None:
-    #     del data['revision']
-    # if data['type'] is None:
-    #     data['type'] = model.name
 
     # Make a copy of data, because `pymongo` changes the reference `data`
     # object on `insert_one()` call.
@@ -140,9 +133,6 @@ def push(context: Context, model: Model, backend: Mongo, data: dict, *, action: 
 
     # parse `ObjectId` to string and add it to our object
     raw_data['id'] = str(data_id)
-
-    # do not return inner Mongo `_id`
-    # del raw_data['_id']
 
     return prepare(context, action, model, backend, raw_data)
 
--- a/spinta/backends/mongo.py
+++ b/spinta/backends/mongo.py
@@ -201,4 +201,7 @@ def prepare(context: Context, action: str, model: Model, backend: Mongo, value: 
             **value,
         }
     elif action in ('insert', 'update'):
-        return value['id']
+        return {
+            'type': model.name,
+            'id': value['id']
+        }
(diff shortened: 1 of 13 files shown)


spinta/backends/mongo.py
```diff
@@ -176,10 +167,7 @@ def getall(context: Context, model: Model, backend: Mongo, **kwargs):
     # Yield all available entries.
     model_collection = backend.db[model.get_type_value()]
     for row in model_collection.find({}):
-        id = str(row.pop('_id'))
-        yield {
-            'id': id,
-        }
+        yield prepare(context, 'getall', model, backend, row)
```
sirex: I think, here an iterator should be passed to `prepare`, because you need to have control, how all rows are wrapped, for example they can be wrapped into `{"resources": [...]]}` or something like that.

sirex: Same for postgre.

trimailov: I assumed that exporter would be responsible for this?

https://github.com/atviriduomenys/spinta/blob/master/spinta/config/__init__.py#L73

spinta/backends/mongo.py
```diff
@@ -196,3 +184,21 @@ def wipe(context: Context, model: Model, backend: Mongo):
...
+        # TODO: this must be fixed/implemented in the spinta/types/store.py::get()
+        # just like it's done on spinta/types/store.py::push()
+        id = str(value.pop('_id'))
+        return {
+            'type': model.name,
+            'id': id,
+            **value,
+        }
+    elif action in ('insert', 'update'):
+        return value['id']
```
sirex: `prepare` must be recursive, because nested properties might have backend specific values, that needs to be converted to python-native values, for example dates are good candidates for this conversion.

spinta/types/type.py
```diff
@@ -34,8 +34,8 @@ def is_valid(self, value: typing.Any):
 class PrimaryKey(Type):
 
     def load(self, value: typing.Any):
-        if value is None:
-            return None
+        # if value is None:
+        #     return None
```
sirex: Why these are commented out?

spinta/backends/mongo.py
```diff
@@ -110,12 +110,12 @@ def push(context: Context, model: Model, backend: Mongo, data: dict, *, action:
...
-    if data['revision'] is None:
-        del data['revision']
-    if data['type'] is None:
-        data['type'] = model.name
+    # if data['id'] is None:
+    #     del data['id']
+    # if data['revision'] is None:
+    #     del data['revision']
+    # if data['type'] is None:
+    #     data['type'] = model.name
```
sirex: Why these are commented out?

trimailov: https://github.com/atviriduomenys/spinta/pull/28/commits/6f0e4d3c081be8700ba30e69dd569ce9dd32569b

> Hardcoded changes made to make qvarn-like revisions work, though it
basically broke every spinta's test. For now - roll back (or comment
out) these changes for the sake, that tests would be green.

trimailov: I've commented them out - as it's highly likely that they will be uncommented

Though following best practices it would've been best to just remove them overall..

spinta/backends/mongo.py
sirex: Then, why you need `action` here?

spinta/backends/mongo.py
sirex: Could you give me an example, of how tests were broken?

spinta/types/type.py
trimailov: removed

spinta/backends/mongo.py
trimailov: Because different actions may have differently structured responses, e.g. POST in spinta only returns the `type` and `id` instead of the whole object.

PR #304, opened 2022-10-02
Add support for text type

sirex:


adp-atea:
1) It seems that prop@lang is not parsing, when when reading from external source
2) When selecting app.get(f'/{model}/?select(title@lt)') is not parsing. Also same with prop@lange = value and sort(prop@lang)

tests/dtypes/test_text.py
```diff
@@ -0,0 +1,226 @@
...
+        '_type': model,
+        'title': {'lt': 'latvija', 'en': 'latvia'},
+    })
+
+    assert resp.status_code == 201
+
+    # Test if select(prop @ lang) works
+    select_by_prop = app.get(f'/{model}/?select(title@lt)')
+
+    # assert select_by_prop = select_by_prop
```
sirex, line 217: Please do not leave commented out code.

sirex (commented):
There are multiple errors see my notes in `notes/types/text.sh`.

spinta/types/model.py
```diff
@@ -329,8 +333,14 @@ def load(context: Context, model: Model, data: dict) -> dict:
     # check that given data does not have more keys, than model's schema
     non_hidden_keys = []
     for key, prop in model.properties.items():
+        if isinstance(prop.dtype, Text):
+            prop.hidden = True
```
adp-atea, line 358: I added text prop to hidden, however IDK if this the best place

adp-atea:
There might be a problem for type push chunks, could not write a decent test and for just added skip wrapper for them. I might be wrong but it seems in our case its not possible to create jsonb format,
sa.dialects.postgresql import JSONB for ex.
``@pytest.fixture(scope='module')
def geo_db_for_text():
    with create_sqlite_db({
        'salis': [
            sa.Column('id', sa.Integer, primary_key=True),
            sa.Column('kodas', sa.Text),
            sa.Column('pavadinimas', sa.dialects.postgresql.JSONB),
        ],
        'miestas': [
            sa.Column('id', sa.Integer, primary_key=True),
            sa.Column('pavadinimas', sa.Text),
            sa.Column('salis', sa.dialects.postgresql.JSONB),
        ],
    })
It would return render error for test_text_type_push_chunks and test_type_text_push if something like geo_db_for_text would be created.

sirex:
From given example:

```python
@pytest.fixture(scope='module')
def geo_db_for_text():
    with create_sqlite_db({
        'salis': [
            sa.Column('id', sa.Integer, primary_key=True),
            sa.Column('kodas', sa.Text),
            sa.Column('pavadinimas', sa.dialects.postgresql.JSONB),
        ],
        'miestas': [
            sa.Column('id', sa.Integer, primary_key=True),
            sa.Column('pavadinimas', sa.Text),
            sa.Column('salis', sa.dialects.postgresql.JSONB),
        ],
    })
```

I see, that you are trying to pass PostgreSQL JSONB column type, to an SQLite database, this will not work, you can't use column types from other dialects, here you must use SQLite types.

I'm not sure, what you are trying to achieve here, but if you try to store `text` data to database, you should do this on PostgreSQL database, not on SQLite. SQLite is usually used to simulate an external datasource and we do not store data to external data sources, we only store data to internal backends, and internal backend uses PostgreSQL.

Issue #149, opened 2021-08-04
Push ref dependencies


sirex:
Models with self-references can't be simply pushed by reading whole model data, self references must be automatically parametrized and read in chunk, starting with least dependent objects first.

For example if we have data like this:

| id | parent |
| -- | ------ |
| 1  | null   |
| 2  | 1      | 

Here, we should start with a block, where parent is null, once that done, then we should read all rows, where parent is 1. And so on. This way, we can ensure, that objects are created in correct order and we will not get foreign key constraints.

There might be more complicated cases, like cycles, but I will deal with such cases later.

### Depends on

- #311 
- https://github.com/atviriduomenys/spinta/issues/363

sirex:
Similar issue:

```python
ERROR: Error when sending and receiving data. Model datasets/gov/ivpk/adp/catalog/Dataset,
Server response (status=500):
    {
        'errors': [
            {
                'code': 'IntegrityError',
                'message': '''
                    (psycopg2.errors.ForeignKeyViolation) insert or update on table
                    "datasets/gov/ivpk/adp/catalog/UserLike" violates foreign key constraint
                    "fk_datasets/gov/ivpk/adp/catalog/UserLike_request._id"
                    
                    DETAIL:
                      Key (request._id)=(1a0d39cf-31f9-47d3-8691-761ff9e4814e) is not present in
                      table "datasets/gov/ivpk/adp/catalog/Request".
                    
                    SQL:
                      INSERT INTO "datasets/gov/ivpk/adp/catalog/UserLike":
                        _txn: 'c66bb4ad-b884-4a6b-b86b-eb92a8b64751',
                        _created: TIMEZONE('utc', CURRENT_TIMESTAMP),
                        _id: '152cc978-1f48-4271-ab57-fe3c1ab14327',
                        _revision: 'ee9262cc-f4f6-4b87-9cd8-fb748ddcfc8e',
                        id: 13,
                        created: '2021-03-01T08:52:48.356114',
                        modified: NULL,
                        version: 0,
                        request._id: '1a0d39cf-31f9-47d3-8691-761ff9e4814e'
                ''',
            },
        ],
    }
ERROR: Error while reading data.
Traceback (most recent call last):
  File "spinta/cli/push.py", line 200, in push
    next(rows)
  File "spinta/cli/push.py", line 468, in _save_push_state
    for row in rows:
  File "spinta/cli/push.py", line 268, in _push_to_remote_spinta
    yield from _send_and_receive(
[...]

PR #312, opened 2022-10-13
Retry pushed rows on error

sirex:
- [x] Mark rows failed to push in push state.
- [x] Do insert, patch instead of upsert.
- [x] Do delete if object was removed from source.
- [x] When repeating row with error, first check if it exists on target.
- [x] Repeat push while there are rows with errors left, do this up to specified number of retries.
- [ ] Skip rows with references if referring row is not yet pushed (#149).
- [x] Skip rows if already pushed (when source emits duplicate rows).
- [x] Check if existing state table has newly added columns.


spinta/cli/push.py
```diff
@@ -156,50 +168,153 @@ def push(
...
+    context: Context,
+    client: requests.Session,
+    server: str,
+    models: List[Model],
+    *,
+    state: Optional[_State] = None,
+    chunk_size: Optional[int] = None,
+    dry_run: bool = False,
+    stop_on_error: bool = False,
+):
```
sirex, line 279: Maybe it is possible to incorporate handling of deleted rows into may iterator? Because, ne this is a completely separate flow of code, and things like limiting number of rows is not added.

So instead of separate flow, I would suggest to move `_get_deleted_rows` into `iter_mode_rows` or probably create a separate `_read_rows`, where first we iterate over existing model rows and then deleted rows.

Then everything would fall into a single flow:

- `_read_rows` - first read available rows, then read deleted from state
- `_rows_to_push_rows` - wrap into `_Push row`
- `_add_stop_time` - limit whole push time
- `_check_push_state` - sync with push state
- `slice` - if limit is applied
- `_push_to_remote_spinta` - sync with remote
- `_save_push_state` - update push state

This way we will have a single iterator, and will be able to add more things on top of it, if needed.

tests/test_push.py
```diff
@@ -155,3 +168,342 @@ def test__send_data__json_error(rc: RawConfig, responses: RequestsMock):
...
+        state=state,
+    )
+
+    query = sa.select([table.c.id, table.c.revision, table.c.error, table.c.deleted])
+    assert list(conn.execute(query)) == [(
+        '4d741843-4e94-4890-81d9-5af7c5b5989a',
+        rev_after,
+        False,
+        True,
+    )]
```
sirex, line 509: There are no test for retries, and retries do not work, at least when I tried manually.

sirex (commented):
I think, it is better to incorporate delete functionality into main flow, so that it would be no need for clean up function, but delete actions would be part of the one main flow.

Also there are two issues, I found, when testing:

- When deleted object re-appears, it is not created on remote target.
- Errors are not retries.

I think, there is no need for `deleted` column in push state table, if an object is successfully deleted, it could be deleted from push state. And in order to find, what needs to be deleted, we can check all `pushed is null and error is false`. This way, before each push operation, we reset `pushed` column to NULL, then run over all objects, and those objects who are not ended up with an error, needs to be deleted on remote target.

But before that, objects, marked with error, should be retried.

sirex:
Another option might be, to first update push state, by marking, what needs to be updated, what needs to be deleted, and then, in a second stage, just go through all the rows in push state and and the fetch from source only single row and also update single row in push state.

This way, we will not need to run counting on source. Yes, this is probably the best solution.

adp-atea:
I tried to fix the delete and retry rows, but had to do a lot of changes. One of them was yielding signal rows between model, delete and retry rows. These rows acted as a signal to know when chunk of data should be pushed. I'm not sure if it's okay, but I couldn't find another way for it to work.

Another problem is that now counts are not displayed correctly, but I'm not sure how how they should look like. Should there be different counters for pushing model rows, pushing delete rows and then for retry rows?

spinta/cli/push.py
```diff
@@ -589,23 +806,203 @@ def _save_push_state(
...
+                    checksum=_get_data_checksum(data),
+                    saved=True,
+                    op="patch",
+                    error=True
+                )
+
+        elif resp.status_code == 404:
+            # Was deleted on both - local and target servers
+            if not data:
+                continue
```
sirex, line 998: Shouldn't we update state here?

spinta/cli/push.py
```diff
@@ -245,69 +370,71 @@ def _attach_keymaps(context: Context, store: Store) -> None:
...
+                'name': 'eq',
+                'args': [
+                    {'name': 'bind', 'args': ['_id']},
+                    row.data['_id'],
+                ]
+            }
+            row.data['_where'] = spyna.unparse(where)
+        elif row.op == 'insert':
+            # if _revision is passed insert action gives ManagedProperty error
+            if '_revision' in row.data:
```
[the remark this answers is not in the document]
adp-atea: _revision comes from getall command, I'm not really sure how this could be fixed, because as I understand _revision in only not needed for push, but for other places that use getall command, it is needed.

spinta/cli/push.py
```diff
@@ -589,23 +929,329 @@ def _save_push_state(
     conn = context.get('push.state.conn')
     for row in rows:
         table = metadata.tables[row.data['_type']]
-        if row.saved:
+
+        if row.error and row.op != "delete":
+            data = json.dumps(row.data)
```
sirex, line 934: I tink, we need to call `fix_data_for_json` as in other places, to convert things like dates to json serializable form.

spinta/cli/push.py
```diff
@@ -524,33 +740,139 @@ def _add_stop_time(rows, stop):
...
+            ))
+            need_reflect.append(_add_column(
+                engine,
+                model.name,
+                columns,
+                sa.Column('data', sa.Text)
+            ))
+
+        if any(need_reflect):
+            metadata.reflect(only=[model.name], extend_existing=True)
```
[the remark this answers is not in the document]
adp-atea: Methods for adding and renaming columns only update database, but they don't update metadata object, that is later used in other methods. So reflect is needed to update it, but it is called only if at least one column was actually changed in database. So if table exists and we check if column needs to be renamed or added and if it turns out that it already exists or is renamed, then reflect will not be called and it will go to table.create.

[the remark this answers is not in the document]
sirex: Metadata object does not need to be updated or reflected from database, because we define it:

```python
            table = sa.Table(
                model.name, metadata,
                sa.Column('id', sa.Unicode, primary_key=True),
                sa.Column('checksum', sa.Unicode),
                sa.Column('revision', sa.Unicode),
                sa.Column('pushed', sa.DateTime),
                sa.Column('error', sa.Boolean),
                sa.Column('data', sa.Text)
            )
```

So we already have an up to date metadata, we only need to update the database schema, and for this, we either run `table.create()` or add/rename columns on already created table.

The issue here, that reflect is quite expensive operation and this adds up, when we have a large number of tables. So I would avoid doing completely unnecessary operations, especially expensive operations.

PR #317, opened 2022-11-07
Geometry objects are represented as types. Link coordinates transformed to WGS.

Gediminas-msc:


sirex:
Please fix failing tests.

spinta/backends/postgresql/types/geometry/encode.py
```diff
@@ -42,6 +44,14 @@ def prepare_dtype_for_response(
...
     point = shape.centroid
-    params = urlencode({'mlat': point.x, 'mlon': point.y})
-    link = f'https://www.openstreetmap.org/?{params}#map=19/{point.x}/{point.y}'
-    return Cell(shape.wkt, link=link)
+
+    if WKBElement.srid in [4326, -1]:
+        x, y = point.x, point.y
+    else:
+        proj = Transformer.from_proj(WKBElement.srid, 4326, always_xy=True)
+        x, y = proj.transform(point.x, point.y)
```
sirex, line 52: Please add tests, where this transformation is checked. Test should check at least two cases, when property is in WGS and in LKS. In one case (WGS) transformation is not needed, in other case transformation is needed. Also test this with different shapes, at least with point and polygon.

https://shapely.readthedocs.io/en/stable/manual.html#shapely.ops.transform

spinta/backends/postgresql/types/geometry/encode.py
```diff
@@ -42,6 +44,14 @@ def prepare_dtype_for_response(
 ):
     shape = to_shape(value)
     point = shape.centroid
-    params = urlencode({'mlat': point.x, 'mlon': point.y})
-    link = f'https://www.openstreetmap.org/?{params}#map=19/{point.x}/{point.y}'
-    return Cell(shape.wkt, link=link)
+
+    if WKBElement.srid in [4326, -1]:
```
sirex, line 48: Do not hardcode cryptic values, instead assign these numbers to a constant with self explanatory names.

sirex (changes requested):

tests/dtypes/test_geometry.py
```diff
@@ -135,3 +137,131 @@ def test_geometry_params_srid(
...
+
+    # Write data
+    resp = app.post(f'/{model}', json={
+        'name': "Vilnius",
+        'coordinates': 'SRID=3346;POINT(582710 6061887)',
+    })
+    assert resp.status_code == 201
+
+    # Read data
+    resp = app.get(f'/{model}')
```
sirex, line 260: We need to test if data are converted for html format and wee need to do conversion only for link to OSM map, we don't need to do conversion in other formats.

spinta/backends/postgresql/types/geometry/encode.py
```diff
@@ -40,8 +44,17 @@ def prepare_dtype_for_response(
     action: Action,
     select: dict = None,
 ):
-    shape = to_shape(value)
+    if value.srid not in [4326, -1]:
```
sirex, line 47: Move `4326` to `WGS` constant.

spinta/backends/postgresql/types/geometry/encode.py
```diff
@@ -40,8 +44,17 @@ def prepare_dtype_for_response(
...
+
     point = shape.centroid
-    params = urlencode({'mlat': point.x, 'mlon': point.y})
-    link = f'https://www.openstreetmap.org/?{params}#map=19/{point.x}/{point.y}'
-    return Cell(shape.wkt, link=link)
+    x, y = point.x, point.y
+    params = urlencode({'mlat': x, 'mlon': y})
+    link = f'https://www.openstreetmap.org/?{params}#map=19/{x}/{y}'
+    shape_type = shape.type == 'Point' and shape.wkt or shape.type.upper()
+    return Cell(shape_type, link=link)
```
sirex, line 60:
```suggestion
    if isinstance(shape, Point):
        value = shape.wkt
    else:
        # Shorten possibly long WKT values by reducing it to shape type.
        value = shape.type.upper()
    return Cell(value, link=link)
```

Where `Point` comes from `from shapely.geometry import Point`.

Consider this suggestion, because:

- variable name `shape_type` might be shape_type or wkt value, so better name things correctly,
- avoid relying on raw values like `'Point'` in this case, better use constants and class names, they are more reliable and Python can check if a name `Point` exists.
- avoid cryptic code like `condition and then or else`, just use `if condition: then; else: else` or at least `then if condition else else`.

tests/dtypes/test_geometry.py
```diff
@@ -135,3 +144,113 @@ def test_geometry_params_srid(
...
+        'coordinates': Cell(
+            value='POLYGON',
+            link=None,
+            color=None,
+        ),
+    }
+
+
+@pytest.mark.parametrize('value, cell', [
+    ({'geom_type': 'Point', 'geom_value': [582710, 6061887], 'srid': 4326},
```
sirex, line 187: Use a constant instead of cryptic `4326` value.

tests/dtypes/test_geometry.py
```diff
@@ -135,3 +144,113 @@ def test_geometry_params_srid(
...
+     Cell(value='POINT (582710 6061887)',
+          link='https://www.openstreetmap.org/?mlat=582710.0&mlon=6061887.0#map=19/582710.0/6061887.0')),
+    ({'geom_type': 'Polygon', 'geom_value': [[[502640, 6035299], [502633, 6035297], [502631, 6035305]]], 'srid': 4326},
+     Cell(value='POLYGON', link='https://www.openstreetmap.org/?mlat=502634.6666666667&mlon=6035300.333333333'
+                                '#map=19/502634.6666666667/6035300.333333333')),
+])
```
sirex, line 193: This is quite unreadable and you are testing two different things (1) projection conversion and (2) wkt shortening. Please, could you separate these into separate tests?

sirex (changes requested):

spinta/backends/postgresql/constants.py
```diff
@@ -8,3 +8,6 @@
     'generic',
     'rql',
 ]
+
+WGS84 = 4326
+LKS94 = 3346
```
sirex, line 13: This should go to `spinta.types.geometry.constants`, because these numbers has nothing to do with PostgreSQL it is specific to `geometry` type.

spinta/backends/postgresql/types/geometry/encode.py
```diff
@@ -40,8 +46,21 @@ def prepare_dtype_for_response(
     action: Action,
     select: dict = None,
 ):
-    shape = to_shape(value)
+    if value.srid not in [WGS84, -1]:
+        wgs_crs = CRS(f'EPSG:{WGS84}')
+        value_crs = CRS(f'EPSG:{value.srid}')
+        project = Transformer.from_crs(crs_from=value_crs, crs_to=wgs_crs, always_xy=True).transform
+        shape = transform(project, to_shape(value))
```
sirex, line 53: Transformed value should only be used in the URL to OpenStreetMap, because OpenStreetMap uses WGS, but actual value (link text) should be shown in original projection.

tests/dtypes/test_geometry.py
```diff
@@ -135,3 +145,153 @@ def test_geometry_params_srid(
...
+            value='19e4f199',
+            link='/example/City/19e4f199-93c5-40e5-b04e-a575e81ac373',
+            color=None,
+        ),
+        'name': Cell(
+            value='Vilnius',
+        ),
+        'coordinates': Cell(
+            value='POLYGON',
+            link=None,
```
sirex, line 181: Why there is no link? A link to polygon centroid should be here.

tests/dtypes/test_geometry.py
sirex: Please press „Resove conversation“ if it is resolved, or add a comment otherwise.

tests/dtypes/test_geometry.py
```diff
@@ -135,3 +145,153 @@ def test_geometry_params_srid(
...
     ]
+
+
+def test_geometry_html(rc: RawConfig):
+    context, manifest = load_manifest_and_context(rc, '''
+    d | r | b | m | property                   | type           | ref
+    example                                    |                |
+      |   |   | City                           |                |
+      |   |   |   | name                       | string         |
+      |   |   |   | coordinates                | geometry(4326) |
```
sirex, line 156: Could you add a `description` column with WGS text, in order explain, what 4326 means?

tests/dtypes/test_geometry.py
```diff
@@ -135,3 +145,153 @@ def test_geometry_params_srid(
...
+    ),
+    (
+        {
+            'geom_type': 'Polygon',
+            'geom_value': [[[502640, 6035290], [502630, 6035290], [502630, 6035300]]],
+            'srid': LKS94
+        },
+        Cell(value='POLYGON',
+             link='https://www.openstreetmap.org/?mlat=24.040608716271&mlon=54.454444306004426'
+                  '#map=19/24.040608716271/54.454444306004426')
```
sirex, line 225: I think, it is enough to just add POINT for this test, in order to test transformation, probably tests with POLYGON is redundant?

tests/dtypes/test_geometry.py
```diff
@@ -135,3 +145,153 @@ def test_geometry_params_srid(
...
+    ),
+])
+def test_geometry_coordinate_tansformation(rc: RawConfig, value, cell):
+    context, manifest = load_manifest_and_context(rc, f'''
+        d | r | b | m | property                   | type           | ref   | description
+        example                                    |                |       |
+          |   |   | City                           |                |       |
+          |   |   |   | name                       | string         |       |
+          |   |   |   | coordinates                | geometry       |       | WGS(LKS)
```
sirex, line 234: What does `WGS(LKS)` mean?

tests/dtypes/test_geometry.py
```diff
@@ -135,3 +145,153 @@ def test_geometry_params_srid(
...
+            color=None,
+        ),
+    }
+
+
+@pytest.mark.parametrize('value, cell', [
+    (
+        {
+            'geom_type': 'Point',
+            'geom_value': [582710, 6061887],
```
sirex, line 191: These numbers do not look like WGS, maybe use:

```suggestion
            'geom_value': [25.28277, 54.68661],
```

?

tests/dtypes/test_geometry.py
```diff
@@ -135,3 +145,153 @@ def test_geometry_params_srid(
...
+            'srid': LKS94
+        },
+        Cell(value='POINT (25.282777879597916 54.68661318326901)',
+             link='https://www.openstreetmap.org/?mlat=25.282777879597916&mlon=54.68661318326901'
+                  '#map=19/25.282777879597916/54.68661318326901')
+    ),
+    (
+        {
+            'geom_type': 'Polygon',
+            'geom_value': [[[502640, 6035290], [502630, 6035290], [502630, 6035300]]],
```
sirex, line 210: These numbers do not look like WGS.

tests/dtypes/test_geometry.py
```diff
@@ -135,3 +145,153 @@ def test_geometry_params_srid(
...
+        },
+        Cell(value='POINT (582710 6061887)', link=None)
+    ),
+    (
+        {
+            'geom_type': 'Polygon',
+            'geom_value': [[[502640, 6035290], [502630, 6035290], [502630, 6035300]]],
+            'srid': WGS84
+        },
+        Cell(value='POLYGON', link=None)
```
sirex, line 271: `link` should not be None, it should be a link to OSM map to centroid of polygon.

sirex (changes requested):

tests/dtypes/test_geometry.py
Gediminas-msc: Ambiguous abbreviation from description column removed.

tests/dtypes/test_geometry.py
Gediminas-msc: Values changed to similar in WGS values.

tests/dtypes/test_geometry.py
Gediminas-msc: Values changed to similar in WGS values.

sirex (changes requested):
I did some refactoring to separate different things from a command into separate function. Please finish this refactoring.

See: https://en.wikipedia.org/wiki/Separation_of_concerns

sirex (approved):

PR #327, opened 2022-11-17
Check dataset, model and property names.

Gediminas-msc:


Gediminas-msc:
Could you explain what this means: "Check if dataset, model and property names uses only allowed symbols."?

spinta/datasets/commands/check.py
```diff
@@ -1,14 +1,45 @@
 from spinta import commands
 from spinta.components import Context
 from spinta.datasets.components import Dataset, Resource
+import re
```
sirex, line 4: Standard library imports always goes to the top, see [PEP-8](https://pep8.org/#imports).

spinta/cli/config.py
```diff
@@ -29,8 +29,9 @@ def check(
         "Manifest files to load"
     )),
     mode: Mode = Option('internal', help="Mode of backend operation"),
+    names: bool = Option(False, '--names', help="To check dataset, model and property names"),
```
sirex, line 32: There is not need to add `'--names'`, it is added by default.

spinta/core/config.py
```diff
@@ -579,6 +579,7 @@ def configure_rc(
     manifests: List[Union[str, ManifestPath]] = None,
     *,
     mode: Mode = Mode.internal,
+    names: bool = False,
```
sirex, line 582: Maybe change this to `check_names`? Because `names` is too ambiguous.

spinta/core/config.py
```diff
@@ -642,6 +643,7 @@ def configure_rc(
             'manifest': inline,
         }
         config['manifest'] = 'default'
+        config['names'] = names
```
sirex, line 646: Same here, change to `check_names`?

spinta/datasets/commands/check.py
```diff
@@ -1,14 +1,45 @@
...
 
 
 @commands.check.register()
 def check(context: Context, dataset: Dataset):
     for resource in dataset.resources.values():
         commands.check(context, resource)
 
+    rc = context.get('rc')
+
+    if rc.get('names'):
```
sirex, line 14:
```suggestion
    config: Config = context.get('config')

    if config.check_names:
```

All configuration parameters should be set on top of `config`.

See: https://github.com/atviriduomenys/spinta/blob/master/spinta/types/config.py

spinta/datasets/commands/check.py
```diff
@@ -1,14 +1,45 @@
...
+        name = dataset.name
+        is_lowercase = len([i for i in name if i.islower()] + re.findall('/', name)) == len(name)
```
sirex, line 16: Can you check name with only regexps? A one regex, like this:

    [a-z][a-z0-1_]*(/[a-z][a-z0-1_]*)*

Should check whole dataset name.

spinta/datasets/commands/check.py
```diff
@@ -1,14 +1,45 @@
...
+
+        # Dataset name check
+        if not is_lowercase or name.startswith('/') or name.endswith('/') or re.findall('//', name):
+            raise Exception(f'Dataset name {name} is not correct.')
```
sirex, line 20: Create a custom class in `spinta/exceptions.py` called `InvalidName`, which should extend `UserError`.

spinta/datasets/commands/check.py
```diff
@@ -1,14 +1,45 @@
...
+
+        store = context.get('store')
+        models = store.manifest.models
+
+        for name, model in models.items():
```
sirex, line 25: Model checks should go to `spinta/types/model.py` on `check[Context, Model]` command method.

spinta/datasets/commands/check.py
```diff
@@ -1,14 +1,45 @@
...
+            if not name.startswith('_'):
+                name = name.rsplit('/', 1)[-1]
+                is_lowercase = name.lower() == name
+                is_uppercase = name.upper() == name
```
sirex, line 29: Use regular expression.

spinta/datasets/commands/check.py
```diff
@@ -1,14 +1,45 @@
...
+
+                # Model name check
+                if not re.findall(r'^[A-Za-z]\w*$', name) or is_lowercase or is_uppercase:
+                    raise Exception(f'Model name {name} is not correct.')
```
sirex, line 33: Use `InvalidName` exception from `spinta/exceptions.py`.

spinta/datasets/commands/check.py
```diff
@@ -1,14 +1,45 @@
...
+
+            for prop_name, prop in model.properties.items():
```
sirex, line 35: Properties should be checked in `spinta/types/model.py` on `check[Context, Property]` command method.

spinta/datasets/commands/check.py
```diff
@@ -1,14 +1,45 @@
...
+
+                # Property name check
+                if re.findall(r'^[A-Z]\w*$', prop_name):
```
sirex, line 38: Use precompiled regular expressions and the regular expression is not correct, please add tests to check if it really works.

sirex (changes requested):

spinta/datasets/commands/check.py
sirex: https://regexkit.com/python-regex

spinta/cli/config.py
```diff
@@ -29,8 +29,9 @@ def check(
         "Manifest files to load"
     )),
     mode: Mode = Option('internal', help="Mode of backend operation"),
+    names: bool = Option(False, help="To check dataset, model and property names"),
```
sirex, line 32:
```suggestion
    check_names: bool = Option(False, help="To check dataset, model and property names"),
```

spinta/core/config.py
```diff
@@ -642,6 +643,7 @@ def configure_rc(
             'manifest': inline,
         }
         config['manifest'] = 'default'
+        config['check.names'] = check_names
```
sirex, line 646: Also add this to `config.yml`.

spinta/datasets/commands/check.py
```diff
@@ -1,14 +1,31 @@
...
 @commands.check.register()
 def check(context: Context, dataset: Dataset):
     for resource in dataset.resources.values():
         commands.check(context, resource)
 
+    config: Config = context.get('config')
+
+    if config.check_names:
+        name = dataset.name
+        is_lowercase = re.compile(r'^[a-z0-1_]*(/[a-z0-1_]*)+$')
```
sirex, line 18: Move this to module level. Also, dataset name can't start with `_`, this is for reserved names. There are some reserved names, but these names should be listed and checked against know reserved names.

Also each part separated with `/` in dataset name should start with `[a-z]` and only then, second ant other symbols can be `[a-z0-9_]`. Dataset name can't start with `/`, so in your regexp first block should end with `+` and second with `*`.

spinta/datasets/commands/check.py
```diff
@@ -1,14 +1,31 @@
...
+
+        # Dataset name check
+        if is_lowercase.match(name) is None or name.startswith('/') or name.endswith('/') or name.count('//'):
```
sirex, line 21: All checks should be done only with regexes. There is no need to manually check slashes.

spinta/types/helpers.py
```diff
@@ -29,3 +32,29 @@ def set_dtype_backend(dtype: DataType):
...
     else:
         dtype.backend = dtype.prop.model.backend
+
+
+def check_model_name(context: Context, model: Model):
+    config: Config = context.get('config')
+    if config.check_names:
+        name = model.name
+        if not name.startswith('_'):
+            camel_case = re.compile(r'[A-Z][a-z]+')
```
sirex, line 42: Move compile to module level. Also names like `CityStreet` should be allowed.

spinta/types/helpers.py
```diff
@@ -29,3 +32,29 @@ def set_dtype_backend(dtype: DataType):
...
+                    class_name=model.__class__.__name__,
+                    name=name
+                )
+
+
+def check_property_name(context: Context, prop: Property):
+    config: Config = context.get('config')
+    if config.check_names:
+        name = prop.name
+        if not name.startswith('_'):
```
sirex, line 54: Check all reserved names against a list of allowed reserved names.

spinta/types/helpers.py
```diff
@@ -29,3 +32,29 @@ def set_dtype_backend(dtype: DataType):
...
+            snake_case = re.compile(r'[a-z0-1_]+')
```
sirex, line 55: Property names can't start with `_` or `[0-9]`.

sirex (changes requested):

spinta/config.yml
```diff
@@ -174,6 +174,12 @@ always_show_id:
 default_auth_client:
   type: string
 
+check:
+  type: object
+  items:
+    check_names:
```
sirex, line 180:
```suggestion
    names:
```

spinta/core/config.py
```diff
@@ -579,6 +579,7 @@ def configure_rc(
     manifests: List[Union[str, ManifestPath]] = None,
     *,
     mode: Mode = Mode.internal,
+    check_names: bool = False,
```
sirex, line 582:
```suggestion
    check_names: Optional[bool] = None,
```

This option can be set via environment variables or other means, so we only should set this, if it is given via command line argument.

spinta/core/config.py
sirex, line 646: Do not set this, if `check_names` is `None`.

spinta/cli/config.py
```diff
@@ -29,8 +29,9 @@ def check(
         "Manifest files to load"
     )),
     mode: Mode = Option('internal', help="Mode of backend operation"),
+    check_names: bool = Option(False, help="To check dataset, model and property names"),
```
sirex, line 32: This should be `None` by default.

spinta/types/helpers.py
```diff
@@ -1,13 +1,23 @@
...
+from spinta.components import Model
+from spinta.components import Property
 from spinta.exceptions import BackendNotFound
+from spinta.exceptions import InvalidName
 
 if TYPE_CHECKING:
     from spinta.types.datatype import DataType
 
+upper_camel_case = re.compile(r'^[a-z][a-z0-9]+(/[a-z][a-z0-9]*)+(/([A-Z][a-z0-9]+)*)$')
+snake_case = re.compile(r'^[a-z][a-z0-9]+(_[a-z0-9]+)*$')
```
sirex, line 19: Add tests.

sirex (changes requested):

spinta/types/helpers.py
```diff
@@ -29,3 +39,27 @@ def set_dtype_backend(dtype: DataType):
...
     else:
         dtype.backend = dtype.prop.model.backend
+
+
+def check_model_name(context: Context, model: Model):
+    config: Config = context.get('config')
+    if config.check_names:
+        name = model.name
+        if name not in ['_ns', '_txn', '_schema', '_schema/version']:
+            if upper_camel_case.match(name) is None:
```
sirex, line 49: Check model name and namespace separately. For example:

    a/b/c/Model

Here `a/b/c` is a namespace and should be checked with `check[Context, Namespace]`, `Model` is a model name and should be checked separately with `check[Context, Model]`.

sirex (changes requested):
Please add tests, and test datasets, model, property names and error messages, add test if single letter names are allowed, add tests for invalid cases.

PR #328, opened 2022-11-21
filename option added to check dataset name for matching csv filename…

### LINKED ISSUE #239: Check if filename matches namespace

Gediminas-msc:
…, check more datasets per file.

### DIFF
--- a/spinta/components.py
+++ b/spinta/components.py
@@ -887,6 +887,7 @@ class Config:
     AccessLog: Type[AccessLog]
     exporters: Dict[str, Format]
     check_filename: Optional[bool] = None
+    check_manifests: list = None
 
     def __init__(self):
         self.commands = _CommandsConfig()
--- a/spinta/components.py
+++ b/spinta/components.py
@@ -886,7 +886,7 @@ class Config:
     data_path: pathlib.Path
     AccessLog: Type[AccessLog]
     exporters: Dict[str, Format]
-    check_filename: List[str] = None
+    check_filename: Optional[bool] = None
 
     def __init__(self):
         self.commands = _CommandsConfig()
--- a/spinta/components.py
+++ b/spinta/components.py
@@ -886,6 +886,7 @@ class Config:
     data_path: pathlib.Path
     AccessLog: Type[AccessLog]
     exporters: Dict[str, Format]
+    check_filename: List[str] = None
 
     def __init__(self):
         self.commands = _CommandsConfig()
--- a/spinta/config.yml
+++ b/spinta/config.yml
@@ -175,10 +175,10 @@ default_auth_client:
   type: string
 
 check:
-  type: bool
+  type: object
   items:
-    filename:
-      type: list
+    check_filename:
+      type: boolean
 
 root:
   type: string
--- a/spinta/config.yml
+++ b/spinta/config.yml
@@ -179,6 +179,8 @@ check:
   items:
     filename:
       type: boolean
+    manifests:
+      type: list
 
 root:
   type: string
--- a/spinta/config.yml
+++ b/spinta/config.yml
@@ -177,7 +177,7 @@ default_auth_client:
 check:
   type: object
   items:
-    check_filename:
+    filename:
       type: boolean
 
 root:
--- a/spinta/config.yml
+++ b/spinta/config.yml
@@ -174,6 +174,12 @@ always_show_id:
 default_auth_client:
   type: string
 
+check:
+  type: bool
+  items:
+    filename:
+      type: list
+
 root:
   type: string
 
--- a/spinta/core/config.py
+++ b/spinta/core/config.py
@@ -579,7 +579,7 @@ def configure_rc(
     manifests: List[Union[str, ManifestPath]] = None,
     *,
     mode: Mode = Mode.internal,
-    check_filename: bool = False,
+    check_filename: bool = None,
     backend: str = None,
     resources: List[ResourceTuple] = None,
 ) -> RawConfig:
--- a/spinta/core/config.py
+++ b/spinta/core/config.py
@@ -579,7 +579,7 @@ def configure_rc(
     manifests: List[Union[str, ManifestPath]] = None,
     *,
     mode: Mode = Mode.internal,
-    check_filename: bool = None,
+    check_filename: Optional[bool] = None,
     backend: str = None,
     resources: List[ResourceTuple] = None,
 ) -> RawConfig:
@@ -643,9 +643,10 @@ def configure_rc(
             'manifest': inline,
         }
         config['manifest'] = 'default'
+        config['manifests'] = manifests
 
-        if check_filename:
-            config['check.filename'] = manifests
+        if check_filename is not None:
+            config['check.filename'] = check_filename
 
     if config:
         rc = rc.fork(config)
--- a/spinta/core/config.py
+++ b/spinta/core/config.py
@@ -643,7 +643,8 @@ def configure_rc(
             'manifest': inline,
         }
         config['manifest'] = 'default'
-        config['manifests.filename'] = manifests
+
+        config['check.manifests'] = manifests
 
         if check_filename is not None:
             config['check.filename'] = check_filename
--- a/spinta/core/config.py
+++ b/spinta/core/config.py
@@ -643,7 +643,7 @@ def configure_rc(
             'manifest': inline,
         }
         config['manifest'] = 'default'
-        config['manifests'] = manifests
+        config['manifests.filename'] = manifests
 
         if check_filename is not None:
             config['check.filename'] = check_filename
--- a/spinta/core/config.py
+++ b/spinta/core/config.py
@@ -579,7 +579,7 @@ def configure_rc(
     manifests: List[Union[str, ManifestPath]] = None,
     *,
     mode: Mode = Mode.internal,
-    filename: bool = False,
+    check_filename: bool = False,
     backend: str = None,
     resources: List[ResourceTuple] = None,
 ) -> RawConfig:
@@ -644,10 +644,9 @@ def configure_rc(
         }
         config['manifest'] = 'default'
 
-        if filename:
-            config['filename'] = manifests
+        if check_filename:
+            config['check.filename'] = manifests
 
     if config:
         rc = rc.fork(config)
-
     return rc
--- a/spinta/core/config.py
+++ b/spinta/core/config.py
@@ -579,6 +579,7 @@ def configure_rc(
     manifests: List[Union[str, ManifestPath]] = None,
     *,
     mode: Mode = Mode.internal,
+    filename: bool = False,
     backend: str = None,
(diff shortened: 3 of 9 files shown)


spinta/core/config.py
```diff
@@ -579,6 +579,7 @@ def configure_rc(
     manifests: List[Union[str, ManifestPath]] = None,
     *,
     mode: Mode = Mode.internal,
+    filename: bool = False,
```
sirex, line 582:
```suggestion
    chec_filename: bool = False,
```

spinta/core/config.py
```diff
@@ -643,6 +644,9 @@ def configure_rc(
         }
         config['manifest'] = 'default'
 
+        if filename:
+            config['filename'] = manifests
```
sirex, line 648:
```suggestion
            config['check.filename'] = manifests
```

spinta/datasets/commands/check.py
```diff
@@ -5,6 +5,17 @@
 
 @commands.check.register()
 def check(context: Context, dataset: Dataset):
+    rc = context.get('rc')
+    filename = rc.get('filename')
+    if filename:
```
sirex, line 10:
```suggestion
    config: Config = context.get('config')
    if config.check_filename:
```

spinta/datasets/commands/check.py
```diff
@@ -5,6 +5,17 @@
...
+        if len(context.get('store').manifest.datasets) > 1:
+            raise Exception(f'File contains more than one dataset.')
```
sirex, line 12: Use `MultipleDatasetsError` exception from `spinta/exceptions.py` (create a new class for this).

spinta/datasets/commands/check.py
```diff
@@ -5,6 +5,17 @@
...
+        for name in filename:
+            if 'csv' in name:
+                csv_ns_name = name[:name.find('csv') - 1]
+                if csv_ns_name not in dataset.ns.name:
+                    raise Exception(f'Dataset namespace {dataset.ns.name} not match the csv filename {csv_ns_name}.')
```
sirex, line 17: Raise `InvalidFileName` exception from `spinta/exceptions.py` (create a new class if it is not defined).

tests/test_checks.py
```diff
@@ -93,3 +94,22 @@ def test_enum_type_boolean(tmpdir, rc):
                              |         | false
     ''')
     commands.check(context, manifest)
+
+
+def test_filename(tmpdir, rc):
+    context, manifest = load_manifest_and_context(rc, '''
```
sirex, line 100: Here, you should write manifest to a file and only then load it from file. See example here:

https://github.com/atviriduomenys/spinta/blob/master/tests/manifests/test_tabular.py#L9-L10

tests/test_checks.py
```diff
@@ -93,3 +94,22 @@ def test_enum_type_boolean(tmpdir, rc):
                              |         | false
     ''')
     commands.check(context, manifest)
+
+
+def test_filename(tmpdir, rc):
```
sirex, line 99:
```suggestion
def test_filename(tmp_path: Path, rc: RawConfig):
```

Always tuse `tmp_path`, because `tmpdir` is becoming deprecated. And, please always use type annotations.

sirex (changes requested):

tests/test_checks.py
```diff
@@ -93,3 +98,28 @@ def test_enum_type_boolean(tmpdir, rc):
...
+    with pytest.raises(InvalidFileName) as e:
+        create_tabular_manifest(tmp_path / 'hidrologija.csv', '''
+        d | r | b | m | property | type    | source
+        datasets/gov/example     |         |
+                                 |         |
+          |   |   | Data         |         |
+          |   |   |   | value    | string  |
+        ''')
+
+        context = load_manifest_get_context(rc, tmp_path / 'hidrologija.csv')
```
sirex, line 113:
```suggestion
        context = load_manifest_get_context(rc, tmp_path / 'hidrologija.csv', check_filename=True)
```

tests/test_checks.py
```diff
@@ -93,3 +98,28 @@ def test_enum_type_boolean(tmpdir, rc):
...
+
+        config = context.get('config')
+        config.check_filename = ['hidrologija.csv']
```
sirex, line 116:
```suggestion
```

tests/test_checks.py
```diff
@@ -93,3 +98,28 @@ def test_enum_type_boolean(tmpdir, rc):
...
+
+        store = context.get('store')
+        manifest = store.manifest
+
+        commands.check(context, manifest)
+
+    assert e.value.message == (
+        "Dataset namespace datasets/gov/example not match the csv filename hidrologija."
+    )
```
sirex, line 125: Add another test when check is successful.

sirex (changes requested):

spinta/config.yml
```diff
@@ -174,6 +174,12 @@ always_show_id:
 default_auth_client:
   type: string
 
+check:
+  type: object
+  items:
+    check_filename:
```
sirex, line 180:
```suggestion
    filename:
```

spinta/core/config.py
```diff
@@ -579,6 +579,7 @@ def configure_rc(
     manifests: List[Union[str, ManifestPath]] = None,
     *,
     mode: Mode = Mode.internal,
+    check_filename: bool = False,
```
sirex, line 582: Should be `None`.

spinta/datasets/commands/check.py
```diff
@@ -1,10 +1,30 @@
...
 @commands.check.register()
 def check(context: Context, dataset: Dataset):
+    config: Config = context.get('config')
+    store: Store = context.get('store')
+
+    filename = config.check_filename
+
+    if filename:
+        if len(store.manifest.datasets) > 1:
+            raise MultipleDatasetsError()
```
sirex, line 18: Maybe pass `store.manifest` as error context?

spinta/datasets/commands/check.py
```diff
@@ -1,10 +1,30 @@
...
+        for name in filename:
+            if 'csv' in name:
```
sirex, line 20: Maybe just drop extension?

https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.with_suffix

sirex (changes requested):

spinta/core/config.py
```diff
@@ -643,7 +644,9 @@ def configure_rc(
         }
         config['manifest'] = 'default'
 
+        if check_filename:
+            config['check.filename'] = manifests
```
sirex, line 648:
```suggestion
        if check_filename is not None:
            config['check.filename'] = check_filename
```

spinta/core/config.py
```diff
@@ -579,6 +579,7 @@ def configure_rc(
     manifests: List[Union[str, ManifestPath]] = None,
     *,
     mode: Mode = Mode.internal,
+    check_filename: bool = None,
```
sirex, line 582:
```suggestion
    check_filename: Optional[bool] = None,
```

spinta/components.py
```diff
@@ -878,6 +878,7 @@ class Config:
     data_path: pathlib.Path
     AccessLog: Type[AccessLog]
     exporters: Dict[str, Format]
+    check_filename: List[str] = None
```
sirex, line 881:
```suggestion
    check_filename: Optional[bool] = None
```

This should be a boolean flag.

spinta/datasets/commands/check.py
```diff
@@ -1,10 +1,31 @@
...
 
 @commands.check.register()
 def check(context: Context, dataset: Dataset):
+    config: Config = context.get('config')
+    store: Store = context.get('store')
+
+    filename = config.check_filename
+    if filename:
+        if len(store.manifest.datasets) > 1:
+            raise MultipleDatasetsError(manifest=store.manifest)
```
sirex, line 18: It looks, that this check must be done on `load` command, because multiple dataset files passed via command line arguments are combined into a single in-memory manifest. So there are two options:

1. Refactor code base, to work on multiple manifests, instead of a single one. This might be difficult to do, but a more correct solution.
2. Do the checking on `load` command, when files are read one by one. This is a less correct, but easier to do.

So probably, for this task second option is a better choice.

spinta/datasets/commands/check.py
```diff
@@ -1,10 +1,31 @@
...
+        for name in filename:
+            if name.endswith('.csv'):
```
sirex, line 20: We don't want to depend on a specific extension.

tests/test_checks.py
```diff
@@ -93,3 +98,45 @@ def test_enum_type_boolean(tmpdir, rc):
...
+def test_check_filename_correct_name(tmp_path: Path, rc: RawConfig):
+    create_tabular_manifest(tmp_path / 'hidrologija.csv', '''
+    d | r | b | m | property | type    | source
+    datasets/gov/hidrologija |         |
+                             |         |
+      |   |   | Data         |         |
+      |   |   |   | value    | string  |
+    ''')
+
+    context = load_manifest_get_context(rc, tmp_path / 'hidrologija.csv')
```
sirex, line 134: There are many tests without `check_filename`, but there is not tests with `check_filename=False`. We should try to avoid duplicating tests.

sirex (changes requested):

tests/test_checks.py
```diff
@@ -93,3 +98,45 @@ def test_enum_type_boolean(tmpdir, rc):
...
+      |   |   |   | value    | string  |
+    ''')
+
+    context = load_manifest_get_context(rc, tmp_path / 'hidrologija.csv', check_filename=True)
+
+    store = context.get('store')
+    manifest = store.manifest
+
+    with pytest.raises(InvalidFileName) as e:
+        commands.check(context, manifest)
```
sirex, line 118: Also would be nice to have a test without error.

spinta/core/config.py
```diff
@@ -643,7 +644,11 @@ def configure_rc(
         }
         config['manifest'] = 'default'
 
+        config['check.manifests'] = manifests
+
+        if check_filename is not None:
+            config['check.filename'] = check_filename
```
sirex, line 650: This adds configuration file only if manifest file is specified via command line arguments, but this should be enabled in all cases, no matter how manifest is configured.

Also, we probably need to set root directory of manifest files, this could be done by adding `-C, --directory` (see [git](https://git-scm.com/docs/git#Documentation/git.txt--Cltpathgt), [tar](https://www.mankier.com/1/tar#-C)) flag to all commands, similarly how it is done with `-o` in `spinta/cli/main.py`.

This way command should look like this:

    spinta -C path/to/root check --check-filename

Or same thing can be set via configuration parameters:

    spinta -o directory=path/to/root check --check-filename

sirex (changes requested):

SauliusPetraitis:
per senas

PR #333, opened 2022-12-28
308 add export to RDF support

adp-atea:


spinta/formats/rdf/commands.py
```diff
@@ -0,0 +1,388 @@
...
+    *,
+    data: Dict[str, Any],
+    action: Action,
+    select: dict = None,
+    prefix: str = None
+):
+    super_ = commands.prepare_dtype_for_response[Context, Format, Ref, dict]
+    data = super_(context, fmt, dtype, value, data=data, action=action, select=select)
+    data['prefix'] = prefix
+    return data
```
sirex, line 342: Ref properties should be rendered as

```xml
<prefix:name rdf:resource="{value['_id']}" />
```

If only `_id` is given (normalized form), but for denormalized form, we need to do full value traversal:

```xml
<prefix:name>
  <prefix:RefModel rdf:about="{value['_id']}">
    <prefix:prop ...>
  </prefix:RefModel>
</prefix:name>
```

spinta/formats/rdf/commands.py
```diff
@@ -0,0 +1,388 @@
...
+        ), prefix=prefix)
+    if value['_content_type'] is None:
+        _content_type = Cell('', prefix=prefix)
+    else:
+        _content_type = Cell(value['_content_type'], prefix=prefix)
+    return {
+        '_id': _id,
+        '_content_type': _content_type,
+        'prefix': prefix
+    }
```
sirex, line 388: Files should be included as links `rdf:aboult`, unless `_content` is explicitly requested.

spinta/formats/rdf/components.py
```diff
@@ -0,0 +1,31 @@
...
+class Rdf(Format):
+    content_type = 'application/rdf+xml'
+    accept_types = {
+        'application/rdf+xml',
+    }
+    params = {}
+
+
+@dataclasses.dataclass
+class Cell:
```
sirex, line 16: I would change this name to `Elem`, or something like this, because RDF does not have things like cells, while HTML tables do have cells. Maybe instead of having a custom class, we could just use [etree.Element](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element)?

tests/formats/test_rdf.py
```diff
@@ -0,0 +1,149 @@
...
+    })
+    city2 = pushdata(app, '/example/rdf/City', {
+        'name': 'Kaunas',
+        'country': {'_id': country['_id']},
+    })
+
+    res = app.get("/example/rdf/City/:format/rdf").text
+    assert res == f'<Rdf>\n'\
+                  f'  <City>\n'\
+                  f'    <_type>{city1["_type"]}</_type>\n'\
```
sirex, line 39: Type should be shown as `rdf:type`.

tests/formats/test_rdf.py
```diff
@@ -0,0 +1,149 @@
...
+                  f'    <_id about="https://testserver/example/rdf/City/{city1["_id"]}">{city1["_id"]}</_id>\n'\
```
sirex, line 40: ID should be shown as `rdf:about`.

tests/formats/test_rdf.py
```diff
@@ -0,0 +1,149 @@
...
+        'country': {'_id': country['_id']},
+    })
+    city2 = pushdata(app, '/example/rdf/City', {
+        'name': 'Kaunas',
+        'country': {'_id': country['_id']},
+    })
+
+    res = app.get("/example/rdf/City/:format/rdf").text
+    assert res == f'<Rdf>\n'\
+                  f'  <City>\n'\
```
sirex, line 38: If `uri` is given, use it here, in this case it should be `dcat:city`, if `uri` is not given, a `<Description rdf:about="example/rdf/City">` form should be used.

tests/formats/test_rdf.py
```diff
@@ -0,0 +1,149 @@
...
+                  f'    <_type>{city1["_type"]}</_type>\n'\
+                  f'    <_id about="https://testserver/example/rdf/City/{city1["_id"]}">{city1["_id"]}</_id>\n'\
+                  f'    <_revision>{city1["_revision"]}</_revision>\n'\
+                  f'    <name>{city1["name"]}</name>\n'\
+                  f'    <country>\n'\
+                  f'      <_id about="https://testserver/example/rdf/Country/{country["_id"]}">{country["_id"]}</_id>\n'\
```
sirex, line 44: `https://testserver/` bit should be moved into `xml:base`, and here a relative form `example/rdf/Country/...` should be used.

tests/formats/test_rdf.py
```diff
@@ -0,0 +1,149 @@
...
+        'country': {'_id': country['_id']},
+    })
+
+    res = app.get("/example/rdf/City/:format/rdf").text
+    assert res == f'<Rdf>\n'\
+                  f'  <City>\n'\
+                  f'    <_type>{city1["_type"]}</_type>\n'\
+                  f'    <_id about="https://testserver/example/rdf/City/{city1["_id"]}">{city1["_id"]}</_id>\n'\
+                  f'    <_revision>{city1["_revision"]}</_revision>\n'\
+                  f'    <name>{city1["name"]}</name>\n'\
```
sirex, line 42: `dct:name` should be used here.

sirex (changes requested):

spinta/formats/rdf/commands.py
```diff
@@ -0,0 +1,460 @@
...
+        text=str(value)
+    )
+
+
+@commands.prepare_dtype_for_response.register(Context, Rdf, Ref, (dict, str, type(None)))
+def prepare_dtype_for_response(
+    context: Context,
+    fmt: Rdf,
+    dtype: File,
+    value,
```
sirex, line 355: Do not create empty elements if value is None.

sirex (changes requested):

spinta/formats/rdf/commands.py
```diff
@@ -0,0 +1,483 @@
...
+            elem = _create_model_element(model, row)
+            root.append(elem)
+
+    for key, val in root.nsmap.items():
+        namespaces.append(f'xmlns:{key}="{val}"')
+    if root.base:
+        namespaces.append(f'xml:base="{root.base}"')
+    if isinstance(root_name, QName):
+        root_name = f"{root.prefix}:{root_name.localname}"
+    namespaces = "\n".join(namespaces)
```
sirex, line 212:
```suggestion
    namespaces = "\n  ".join(namespaces)
```

spinta/formats/rdf/commands.py
```diff
@@ -0,0 +1,460 @@
...
+        base=str(request.base_url),
+        nsmap=prefixes
+    )
+    if action == Action.GETONE:
+        elem = _create_model_element(model, data)
+        root.append(elem)
+    else:
+        for row in data:
+            elem = _create_model_element(model, row)
+            root.append(elem)
```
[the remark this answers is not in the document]
sirex: Could you explain, how this is resolved, I don't see any fix and thing this is still loads all data into memory instead of streaming it to the client, row by row.

sirex (changes requested):

sirex (approved):

PR #340, opened 2023-01-09
Push command accepts --no-progress-bar option to skip counting rows

### LINKED ISSUE #332: Option to not count rows on  push
Add a command line option to not count rows, this will disable progress bar, but will speed up push operation.

The idea is to disable this function:
https://github.com/atviriduomenys/spinta/blob/ee68619225fc878030e6d75f71f480ffd7fcaa0a/spinta/cli/push.py#L152-L157

which is required by progress bar, disabling counts. should also disable progress bar.

Since counts are only needed by progress bar, so command line flag could be called `--no-progress-bar`.

marpetr:
Fixes #332.

### DIFF
--- a/tests/cli/test_push.py
+++ b/tests/cli/test_push.py
@@ -1,0 +1,107 @@
+import pytest
+import sqlalchemy as sa
+from spinta.manifests.tabular.helpers import striptable
+from spinta.testing.cli import SpintaCliRunner
+from spinta.testing.datasets import create_sqlite_db
+from spinta.testing.tabular import create_tabular_manifest
+from tests.datasets.test_sql import configure_remote_server, create_rc
+
+
+@pytest.fixture(scope='module')
+def geodb():
+    with create_sqlite_db({
+        'salis': [
+            sa.Column('id', sa.Integer, primary_key=True),
+            sa.Column('kodas', sa.Text),
+            sa.Column('pavadinimas', sa.Text),
+        ]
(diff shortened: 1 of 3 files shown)


sirex (changes requested):
Looks good, but there should also be a test. Here is an example:

https://github.com/atviriduomenys/spinta/blob/master/tests/datasets/test_sql.py#L551

But the test probably should go here:

https://github.com/atviriduomenys/spinta/tree/master/tests/cli

tests/cli/test_push.py
```diff
@@ -0,0 +1,107 @@
...
+        '-d', 'datasets/gov/example',
+        '-o', remote.url,
+        '--credentials', remote.credsfile,
+        '--no-progress-bar',
+    ])
+    assert result.exit_code == 0
+    assert "Count rows:   0%" not in result.stderr
+    assert "PUSH:   0%|          | 0/3" not in result.stderr
+    assert "PUSH:  33%|###3      | 1/3" not in result.stderr
+    assert "PUSH: 100%|##########| 3/3" not in result.stderr
```
sirex, line 107: That is not a good test, because progress bar is still shown, only in a slightly different form. Please add a better test and fix the code, to truly disable progress bar.

sirex (changes requested):

sirex:
Replaced by https://github.com/atviriduomenys/spinta/pull/384

PR #342, opened 2023-01-16
216 references from denormalized data

adp-atea:


spinta/commands/read.py
```diff
@@ -101,6 +101,7 @@ async def getall(
             action,
             select_tree,
             reserved=['_type', '_id', '_revision'],
+            include_denorm_props=False,
```
sirex, line 104: Denormalized properties, should be included in the result, but lets leave it like this, and fix it with a new task.

sirex (approved):

Issue #148, opened 2021-08-04
Unique constraint in tabular manifests


sirex:
Add support for unique constrains in tabular manifests. It could look like this:

| model | property | type   | ref           |
| ----- | -------- | ------ | ------------- |
| City  |          |        |               |
|       |          | unique | name, country |
|       | name     | string |               |
|       | country  | ref    | Country       |

All table constraints should be specified within model or property context, if property column can be used to set unique constraint name, if not specified, a random one will be generated.

A shorter version for a single property can be used like this:

| model | property | type   | ref           |
| ----- | -------- | ------ | ------------- |
| City  |          |        |               |
|       | name     | string unique |               |
|       | country  | ref    | Country       |

PR #343, opened 2023-01-17
#148 unique constraint in tabular manifests

adp-atea:


spinta/manifests/tabular/helpers.py
```diff
@@ -488,15 +481,17 @@ def read(self, row: Dict[str, str]) -> None:
...
                 'name': row['source'],
                 'prepare': self.data.pop('prepare'),
             }
 
         self.state.model.data['properties'][row['property']] = self.data
-
+        # Decide if unique constrains for column are needed
+        # Needs to have in column name _unique
+        if '_unique' in row['type']:
+            DataType.schema['unique']['default'] = True
```
sirex, line 494: You should never mutate global variables. And proper places for loading `unique` should be here:

https://github.com/atviriduomenys/spinta/blob/master/spinta/types/model.py#L183-L190

But instead of returning `str, list[str]` you should return `str, list[str], dict[str, str]` and add content of third argument into `data` which then will be used to set `prop.unique = data['unique']`.

spinta/backends/postgresql/commands/init.py
```diff
@@ -78,13 +76,15 @@ def prepare(context: Context, backend: PostgreSQL, dtype: DataType):
...
             f"Unknown type {dtype.name!r} for property {prop.place!r}."
         )
-
-    column_type = types[dtype.name]
-    return sa.Column(name, column_type, unique=dtype.unique)
+    column_type = types[dtype.name.split("_unique")[0]]
+    if '_unique' in dtype.name:
+        return sa.Column(name, column_type, unique=dtype.unique)
+    else:
+        return sa.Column(name, column_type)
```
sirex, line 87:
```suggestion
    column_type = types[dtype.name]
    return sa.Column(name, column_type, unique=dtype.prop.unique)
```

sirex (changes requested):

sirex:
A good place to use as an example: https://github.com/atviriduomenys/spinta/pull/344/files

tests/manifests/test_tabular.py
```diff
@@ -226,6 +226,46 @@ def test_time_type(tmp_path, rc):
...
+                             |                    |         |
+      | data                 | postgresql         | default |
+                             |                    |         |
+      |   |   | Country      |                    | code    |
+      |   |   |   | code     | string             |         |
+      |   |   |   | name     | string             |         | locn:geographicName
+                             |                    |         |
+      |   |   | City         |                    | name    |
+      |   |   |   | name     | string unique      |         | locn:geographicName
+      |   |   |   | country  | ref unique         | Country |
```
sirex, line 266:
```suggestion
      |   |   |   |          | unique             | name, country |
      |   |   |   | name     | string             |         | locn:geographicName
      |   |   |   | country  | ref                | Country |
```

Unique together, should be added on both `name` and `country` together, not separately.

sirex (changes requested):

spinta/nodes.py
```diff
@@ -100,12 +100,15 @@ def get_node(
...
-        )
+        if ctype not in [_type+'_unique' for _type in config.components[group]]:
+            raise exceptions.InvalidManifestFile(
+                parent,
+                manifest=manifest.name,
+                eid=eid,
+                error=error,
+            )
+        else:
+            ctype = str(ctype.split("_unique")[0])
```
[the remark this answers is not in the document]
sirex: Still same thing stands. Unique handling should not be here, this is a low level function. Unique should be handled in a command and in a tabular helper. See https://github.com/atviriduomenys/spinta/pull/344/files#diff-ef698dfa7530dce56bc8139a272f35940c59a01449c3c23bfe5fbb538effe5cc

spinta/manifests/tabular/helpers.py
```diff
@@ -859,6 +859,46 @@ def leave(self) -> None:
...
+
+    def read(self, row: ManifestRow) -> None:
+        reader = self.state.stack[-1]
+
+        if not isinstance(reader, (
+            DatasetReader,
+            ResourceReader,
+            BaseReader,
+            ModelReader,
+            PropertyReader,
```
sirex, line 873: Unique should be supporter only on 'model', so here we only need `ModelReader`.

spinta/manifests/tabular/helpers.py
```diff
@@ -859,6 +859,46 @@ def leave(self) -> None:
...
+        )):
+            return
```
sirex, line 875: If `unique` is added not on a `model`, then we should emit an error before return.

spinta/manifests/tabular/helpers.py
```diff
@@ -859,6 +859,46 @@ def leave(self) -> None:
...
+
+        if self.type not in reader.data:
+            reader.data['unique'] = {}
+
+        unique = reader.data['unique']
+
+        self.name = '_'.join(row[REF].split(', '))
+
+        unique[self.type] = {
```
sirex, line 884: Unique values should be added on a name constructed from row[REF], for example if ref is `name, country`, then unique name should be `name_country`.

spinta/manifests/tabular/helpers.py
```diff
@@ -859,6 +859,46 @@ def leave(self) -> None:
...
+            PropertyReader,
+        )):
+            return
+
+        if self.type not in reader.data:
+            reader.data['unique'] = {}
+
+        unique = reader.data['unique']
+
+        self.name = '_'.join(row[REF].split(', '))
```
sirex, line 882: Split just by ',' and then strip spaces. Because `ref` could be given as `name,country` or `  name,     country  `.

sirex (changes requested):

spinta/nodes.py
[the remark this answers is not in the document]
adp-atea: Are you sure you did not look to outdated commit? It should be removed already.

spinta/manifests/tabular/helpers.py
adp-atea: I think it already is

spinta/cli/main.py
adp-atea, line 1: This line
pathlib.PosixPath = pathlib.WindowsPath

Is needed to run project on windows, I accidently left it, but on this commit remove it, `cause not sure if it would be okay to have this on master.

sirex: Monkey patching is definitely a bad idea, and cross platform compatibility is always a good thing. I'm not sure, what is the issue here, but it should not be only posix or only windows specific. Python has a platform independent class `Path`, where both `PosixPath` and ` WindowsPath` are subclassess of.

https://docs.python.org/3/library/pathlib.html

spinta/backends/postgresql/types/ref/init.py
```diff
@@ -29,10 +29,12 @@ def get_pg_foreign_key(
...
     table_name: str,
     model_name: str,
-    column_type: TypeEngine,
-) -> List[Union[sa.Column, sa.Constraint]]:
+    column_type: TypeEngine) -> List[Union[sa.Column, sa.Constraint]]:
     column_name = get_column_name(prop) + '._id'
     nullable = not prop.dtype.required
+    if prop.model.unique:
+        prop.model.unique = [list(map(lambda val: val.replace(
+            get_column_name(prop), column_name), val)) for val in prop.model.unique]
```
sirex, line 37: We can't modify `model` here, this must be done in `load` and in load, we should deal with properties as instances of `Property` class. Here, we can only convert property names into backend naming without changing the existing metadata.

sirex (changes requested):

spinta/manifests/tabular/helpers.py
```diff
@@ -1098,7 +1128,7 @@ def _read_csv_manifest(
         for i, row in enumerate(rows, 1):
             yield str(i), row
     else:
-        with pathlib.Path(path).open(encoding='utf-8-sig') as f:
+        with pathlib.Path(path).open(encoding='utf-8') as f:
```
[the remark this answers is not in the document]
adp-atea: My bad

notes/metadata/unique.sh
```diff
@@ -0,0 +1,181 @@
...
+#|                     '_id': UUID('05a5a1a9-5109-4bcb-a259-115ffb92ba32'),
+#|                     '_revision': '2e1fb767-c052-4b7c-9bad-3de95ed7690a',
+#|                     'name': 'Vilnius',
+#|                     'country': 'Lithuania'
+#|                 }
+#|             "
+#|         }
+#|     ]
+#| }
+# FIXME: Should be 400 error.
```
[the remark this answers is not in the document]
adp-atea: This should be fixed in this tsak
https://github.com/atviriduomenys/spinta/pull/421
Later I think we`ll merge both.

sirex (approved):

PR #344, opened 2023-01-17
259 required not null constraint

adp-atea:


tests/dtypes/test_required.py
```diff
@@ -0,0 +1,224 @@
...
+
+    resp = app.post('/example/City', json={
+        'name': 'City',
+        'description': 'Test city',
+    })
+    city = resp.json()
+
+    resp = app.patch(f'/example/City/{city["_id"]}', json={
+        '_revision': city["_revision"],
+        'description': 'Test city updated',
```
sirex, line 129:
```suggestion
        'name': None,
        'description': 'Test city updated',
```

This would be a better test to check required.

spinta/backends/__init__.py
```diff
@@ -275,6 +276,8 @@ def complex_model_properties_check(
                     data.backend,
                     given,
                 )
+            if data.action == Action.UPSERT and data.saved is NA:
+                check_type_value(prop.dtype, given)
```
sirex, line 280: I think this should go to simple data check?

spinta/backends/__init__.py
```diff
@@ -192,7 +192,8 @@ def simple_data_check(
     backend: Backend,
     value: object,
 ) -> None:
-    check_type_value(dtype, value)
+    if data.action in (Action.UPDATE, Action.INSERT):
```
sirex, line 195: Maybe include UPSET too?

spinta/backends/__init__.py
```diff
@@ -192,7 +192,8 @@ def simple_data_check(
...
+        check_type_value(dtype, value)
```
sirex, line 196: Could you please change `raise Exception` in `check_type_value` into a new exception from `spinta/exceptions.py`, extended from `UserError`? These exceptions provides a lot more information to the client.

spinta/commands/write.py
```diff
@@ -621,6 +622,8 @@ async def validate_data(
                 if '_revision' in data.given:
                     raise exceptions.ManagedProperty(data.model, property='_revision')
             if data.prop:
+                if data.action == Action.UPSERT and data.saved.get(data.prop.name) is NA:
+                    check_type_value(data.prop.dtype, data.given.get(data.prop.name))
```
sirex, line 626: This is already in `complex_data_check`, so I think these two lines should be removed. And I think, this should go into simple check.

spinta/manifests/tabular/helpers.py
```diff
@@ -468,6 +468,11 @@ def read(self, row: Dict[str, str]) -> None:
                 f"defined for this {self.state.model.name!r} model."
             )
 
+        required = False
+        if 'required' in row['type']:
+            required = True
+            row['type'] = row['type'].replace('required', '').strip()
```
sirex, line 474: Could you please move this:

https://github.com/atviriduomenys/spinta/blob/373b654f8369a4b218ee7badba58cc847e71045e/spinta/types/model.py#L183-L190

Out of `load` into here and parse `type` string in one place, because type can be given in following forms:

```python
"string" -> {
  "type": "string",
}
"geometry(point, 3542)" -> {
  "type": "geometry",
  "type_args": ["point", "3542"],
}
"geometry(point, 3542) unique required" -> {
  "type": "geometry",
  "type_args": ["point", "3542"],
  "unique": True,
  "required": True,
}
```

sirex: And please collaborate on this with https://github.com/atviriduomenys/spinta/pull/343

tests/manifests/test_tabular.py
```diff
@@ -216,6 +216,16 @@ def test_prop_type_not_given(tmp_path, rc):
     )
 
 
+def test_prop_type_required(tmp_path, rc):
+    check(tmp_path, rc, '''
+    d | r | b | m | property | type
+    example                  |
+                             |
+      |   |   | City         |
+      |   |   |   | name     | string required
```
sirex, line 225:
```suggestion
      |   |   |   | name     | string required
      |   |   |   | place    | geometry(point) required
```

I think, this addition will not work.

sirex (changes requested):

sirex:
Also `null` or `not null` constraint on database should be added.

spinta/types/model.py
```diff
@@ -180,14 +180,28 @@ def link(context: Context, base: Base):
...
+def _parse_dtype_string(dtype: str) -> dict:
+    args = []
+    required = unique = False
+
+    if 'required' in dtype:
+        required = True
+        dtype = dtype.replace('required', '').strip()
+    if 'unique' in dtype:
+        unique = True
+        dtype = dtype.replace('unique', '').strip()
```
sirex, line 192: That is not a reliable approach to parse dtype. Could you add a test for `type(required) unique` and see what happens.

First we need to handle parentheses if they are present and what goes after `)`, handle separately by splitting by space.

Also you need raise an error, if we get something unrecognized after parenthesis.

And please add tests for this function alone with various inputs.

spinta/types/model.py
```diff
@@ -210,11 +224,8 @@ def load(
     prop.lang = load_lang_data(context, prop.lang)
     prop.comments = load_comments(prop, prop.comments)
 
-    # Parse dtype like geometry(point, 3346)
     if data['type'] is None:
         raise UnknownPropertyType(prop, type=data['type'])
-    dtype_type, dtype_args = _parse_dtype_string(data['type'])
```
sirex, line 216: If this function is not used here anymore, then it should be moved to tabular helpers.

spinta/backends/postgresql/commands/init.py
```diff
@@ -84,7 +84,8 @@ def prepare(context: Context, backend: PostgreSQL, dtype: DataType):
         )
 
     column_type = types[dtype.name]
-    return sa.Column(name, column_type, unique=dtype.unique)
+    nullable = not dtype.required
+    return sa.Column(name, column_type, unique=dtype.unique, nullable=nullable)
```
sirex, line 88: You need to add this to all other types tool. Because, currently unique constraint won't be added to types like ref, geometry and others.

Unfortunately we can't handle this at the property level, because, different types, can have different specifics, so we need to repeat this for each dtype separately.

sirex (changes requested):

PR #347, opened 2023-01-18
Fix error with keymap on non primary key

sirex:


tests/datasets/sql/test_query.py
```diff
@@ -211,3 +213,21 @@ def test_group_2(rc: RawConfig):
...
+      |   |   | Country        |         | id            | COUNTRY    |
+      |   |   |   | id         | integer |               | ID         | open
+      |   |   |   | code       | string  |               | NAME       | open
+      |   |   | City           |         | name          | CITY       |
+      |   |   |   | name       | string  |               | NAME       | open
+      |   |   |   | country    | ref     | Country[code] | COUNTRY_ID | open
+    ''', 'example/City') == '''
+    SELECT
+      "CITY"."NAME",
+      "CITY"."COUNTRY_ID"
```
sirex, line 231: Here, we also need to join `Country` and get `Country.id`, in order to be able to get key for `Country.code` and associate it with `Country.id`.

tests/datasets/test_id.py
```diff
@@ -8,16 +8,20 @@ def test_scalar():
...
     with kmap:
-        val = 42
-        key = kmap.encode('test', val)
-        assert kmap.decode('test', key) == val
-
+        # val = 42
+        # key = kmap.encode('test', val, None)
+        # assert kmap.decode('test', key) == 42
+        pkey = kmap.encode('datasets/gov/example/Country', [1, 'lt'], None, None)
+        xkey = kmap.encode('datasets/gov/example/Country.code', 1, 'lt', 'datasets/gov/example/Country')
```
sirex, line 15:
```suggestion
        xkey = kmap.encode('datasets/gov/example/Country.code', pkey, 'lt')
```

tests/datasets/test_id.py
```diff
@@ -8,16 +8,20 @@ def test_scalar():
...
+        assert kmap.decode('datasets/gov/example/Country.code', pkey) == 'lt'
 
 def test_list():
     context = Context('test')
     kmap = KeyMap('sqlite:///:memory:')
     commands.prepare(context, kmap)
     with kmap:
         val = [42, 'foo', 'bar']
-        key = kmap.encode('test', val)
+        key = kmap.encode('test', val, None, None)
```
sirex, line 26: Existing keymap API should be backwards compatible.

tests/datasets/test_sql.py
```diff
@@ -275,34 +275,32 @@ def test_filter_join_ne_array_value(rc, tmp_path, geodb):
     ]
 
 
-@pytest.mark.skip('todo')
+#@pytest.mark.skip('todo')
```
sirex, line 278: Do not leave commented out code.

tests/datasets/test_sql.py
```diff
@@ -275,34 +275,31 @@ def test_filter_join_ne_array_value(rc, tmp_path, geodb):
...
-       |   |   |                  |             |                    |        |               |       |        |     |         |
-       |   |   |   | city         | miestas     | country.code!='ee' |        | name          |       |        |     | City    |
-       |   |   |   |   | name     | pavadinimas |                    | string |               | 3     | open   |     | Name    |
-       |   |   |   |   | country  | salis       |                    | ref    | country[code] | 4     | open   |     | Country |
+    id | d | r | b | m | property | source      | prepare            | type    | ref            | level | access | uri | title   | description
+       | datasets/gov/example     |             |                    |         |                |       |        |     | Example |
+       |   | data                 |             |                    | sql     |                |       |        |     | Data    |
+       |   |   |                  |             |                    |         |                |       |        |     |         |
+       |   |   |   | Country1     | salis       |                    |         | id,code        |       |        |     | Country |
```
sirex, line 284: Change `datasets/gov/example` to `keymap` (or something like that) instead of changing model name.

tests/datasets/test_sql.py
```diff
@@ -275,34 +275,31 @@ def test_filter_join_ne_array_value(rc, tmp_path, geodb):
...
-
-    resp = app.get('/datasets/gov/example/city?sort(name)')
+    resp = app.get('/datasets/gov/example/City1?sort(name)')
     data = listdata(resp, 'country._id', 'name', sort='name')
     data = [(codes.get(country), city) for country, city in data]
     assert data == [
-        ('lv', 'Ryga'),
-        ('lt', 'Vilnius'),
+         ('lv', 'Ryga'),
+         ('lt', 'Vilnius'),
```
sirex, line 302: It looks, that lines are over indented.

tests/datasets/test_id.py
```diff
@@ -8,10 +8,14 @@ def test_scalar():
     kmap = KeyMap('sqlite:///:memory:')
     commands.prepare(context, kmap)
     with kmap:
-        val = 42
-        key = kmap.encode('test', val)
-        assert kmap.decode('test', key) == val
-
+        # val = 42
+        # key = kmap.encode('test', val, None)
+        # assert kmap.decode('test', key) == 42
```
sirex, line 13: Please do not leave commented out code.

spinta/datasets/keymaps/sqlalchemy.py
```diff
@@ -57,31 +57,48 @@ def encode(self, name: str, value: object) -> Optional[str]:
...
-            })
+                })
 
         return key
 
     def decode(self, name: str, key: str) -> object:
         table = self.get_table(name)
         query = sa.select([table.c.value]).where(table.c.key == key)
         value = self.conn.execute(query).scalar()
+        print(value, "????")
```
sirex, line 101: Please do not leave debugging code.

spinta/datasets/keymaps/sqlalchemy.py
```diff
@@ -46,8 +47,7 @@ def get_table(self, name):
             )
             table.create(checkfirst=True)
         return self.metadata.tables[name]
-
-    def encode(self, name: str, value: object) -> Optional[str]:
+    def encode(self, name: str, value: object, primary_key=None) -> Optional[str]:
```
sirex, line 50:
```suggestion
    def encode(self, name: str, value: Any, primary_key: Optional[str] = None) -> Optional[str]:
```

spinta/datasets/keymaps/sqlalchemy.py
```diff
@@ -57,31 +57,48 @@ def encode(self, name: str, value: object) -> Optional[str]:
...
         hash = hashlib.sha1(value).hexdigest()
-
-        # Try to find key by value hash.
-        table = self.get_table(name)
         query = sa.select([table.c.key]).where(table.c.hash == hash)
         key = self.conn.execute(query).scalar()
-        if key is None:
+        if primary_key is not None:
+            parent_table = name.split(".")
+            _query_parent_table = sa.select((self.metadata.tables[parent_table[0]]))
```
sirex, line 67:
```suggestion
            parent_table, = name.split(".", 1)
            _query_parent_table = sa.select((self.metadata.tables[parent_table]))
```

spinta/datasets/keymaps/sqlalchemy.py
```diff
@@ -57,31 +57,48 @@ def encode(self, name: str, value: object) -> Optional[str]:
...
+            result = self.conn.execute(_query_parent_table)
+            for n in result:
```
sirex, line 69: Use [one_or_none](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Result.one_or_none) instead. And query pkey by value hash, instead of looping through all keys.

spinta/datasets/backends/sql/commands/read.py
```diff
@@ -105,10 +102,18 @@ def getall(
                     if isinstance(sel.prop.dtype, PrimaryKey):
                         val = keymap.encode(sel.prop.model.model_type(), val)
                     elif isinstance(sel.prop.dtype, Ref):
-                        val = keymap.encode(sel.prop.dtype.model.model_type(), val)
+                        if sel.prop.dtype.refprops == sel.prop.dtype.model.external.pkeys or len(
+                            sel.prop.dtype.model.external.pkeys) == 1:
```
sirex, line 106: Why do you need this second condition? I think this is the reason why I get incorrect key on my testing notes.

spinta/datasets/backends/sql/commands/read.py
```diff
@@ -105,10 +102,18 @@ def getall(
...
+                            val = keymap.encode(sel.prop.dtype.model.model_type(), val)
+                        else:
+                            pkey = keymap.encode(sel.prop.model.model_type(),
+                                                 sel.prop.dtype.model.external.pkeys[-1].name)
+                            val = keymap.encode(
+                                f'{sel.prop.dtype.model.model_type()}.{sel.prop.dtype.model.external.pkeys[-1].name}',
```
sirex, line 112: Instead of using only last pkey, we should use all keys. And I think we need a helper function here, that would generate a correct keymap name, from given model and optionally prop.dtype.refprops.

sirex (commented):
Key is generated incorrectly, see testing nodes.

spinta/datasets/backends/sql/commands/read.py
adp-atea: It because of this test
tests/datasets/test_sql.py::test_composite_ref_keys

it seems it should be 

but condition is not equal
sel.prop.dtype.refprops == sel.prop.dtype.model.external.pkeys

so i added additional condition
sel.prop.dtype.refprops == sel.prop.dtype.model.external.pkeys or len(
                            sel.prop.dtype.refprops) > len(sel.prop.dtype.model.external.pkeys):

sirex (commented):
Something is off, because I still get incorrect keys, when referenced property is specified explicitly, see my testing notes.

spinta/datasets/backends/sql/commands/read.py
```diff
@@ -105,10 +102,22 @@ def getall(
                     if isinstance(sel.prop.dtype, PrimaryKey):
                         val = keymap.encode(sel.prop.model.model_type(), val)
                     elif isinstance(sel.prop.dtype, Ref):
-                        val = keymap.encode(sel.prop.dtype.model.model_type(), val)
+                        if sel.prop.dtype.model.external.pkeys == sel.prop.dtype.refprops or len(
+                            sel.prop.dtype.refprops) > len(sel.prop.dtype.model.external.pkeys):
+                            val = keymap.encode(sel.prop.dtype.model.name, val)
+                        else:
+                            table = backend.get_table(model, str(sel.prop.dtype.model.external.name))
+                            pk = conn.execute(_get_params_values(table, val)).scalar()
```
sirex, line 110: This generates multiple queries and we don't want to do this, because it slows down read operation significantly, `conn.execute` is very expensive. We have to retrieve primary key from main query, by doing a join in the main query.

spinta/backends/__init__.py
```diff
@@ -264,19 +264,6 @@ def simple_data_check(
...
-def simple_data_check(
-    context: Context,
-    data: DataItem,
-    dtype: Ref,
-    prop: Property,
-    backend: Backend,
-    value: object,
-) -> None:
-    if value and not isinstance(value, dict):
-        raise exceptions.InvalidRefValue(prop, value=value)
```
sirex, line 277: Why this check was deleted? And how this change is related to this task?

tests/datasets/sql/test_query.py
```diff
@@ -115,14 +117,42 @@ def test_unresolved_getattr(rc: RawConfig):
...
-      "CITY"."COUNTRY"
+      "COUNTRY_1"."NAME" AS "NAME_1"
     FROM "CITY"
     LEFT OUTER JOIN "COUNTRY" AS "COUNTRY_1" ON "CITY"."COUNTRY" = "COUNTRY_1"."NAME"
     WHERE "COUNTRY_1"."DEMOCRATIC"
     '''
 
 
 def test_join_aliases(rc: RawConfig):
+    print(_build(rc, '''
```
sirex, line 128: Tests should not print anything.

sirex (commented):
Mostly ok, there is only few questions.

spinta/backends/__init__.py
adp-atea: replaced, going to check if all tests pass

Issue #216, opened 2022-03-30
References from denormalized data


sirex:
Add possibility to create references from denormalized data.

Example:

    d | r | b | m | property     | type | ref
    example                      |      |
      |   |   | City             |      |
      |   |   |   | country      | ref  | Country
      |   |   |   | country.name |      |

Here, `country.name` does not have `type`, because this is a denormalized property, which originated from `Country` model.

Internally we still store `country.name` as given, in  other words, we duplicate country name in two  places.

And when someone queries `City` model, we will give `country.name` even if it was not explicitly requested.

### Links

- https://atviriduomenys.readthedocs.io/dsa/ref.html#denormalizuoti-duomenys
- https://github.com/atviriduomenys/manifest/pull/1387#discussion_r843160851
- https://github.com/atviriduomenys/spinta/blob/master/tests/manifests/test_tabular.py

PR #350, opened 2023-01-24
282 check if denormalized property has a ref

adp-atea:


spinta/types/model.py
```diff
@@ -190,6 +191,45 @@ def _parse_dtype_string(dtype: str) -> Tuple[str, List[str]]:
...
+            model = models[ref_prop.dtype.model]
+            _check_denorm_prop(name_parts[1], prop, model, manifest)
+    else:
+        if name not in properties:
+            raise UndefinedPropertyInDenormProperty(
+                denorm_prop=prop.name,
+                denorm_model=prop.model.name,
+                ref_prop=name,
+                ref_model=model.name,
+            )
```
sirex, line 230:
```suggestion
            raise ReferencedPropertyNotFound(
                prop,
                ref={'property': name, 'model': model.name},
            )
```

Always pass a node as a first argument to exception classes. Also, `ReferencedPropertyNotFound` should be subclass of `PropertyNotFound`, because, basically this is the same thing, but slightly in a different context.

Same thing applies to other exceptions in this function.

spinta/types/model.py
```diff
@@ -210,6 +250,9 @@ def load(
     prop.lang = load_lang_data(context, prop.lang)
     prop.comments = load_comments(prop, prop.comments)
 
+    if data['type'] == 'denorm':
+        _check_denorm_prop(prop.name, prop, prop.model, manifest)
```
sirex, line 254: This should be done in `check` command, because, during `load` not all models might be loaded yet.

sirex (changes requested):

spinta/types/model.py
adp-atea: I saw in testing notes that generated datatypes in database should be fixed, so that they would be the same as in ref model. For that I needed to add rel_prop to Denorm datatype and assign it during link command. And since it was doing pretty much the same thing as _check_denorm_prop, I also moved this check to link.

spinta/types/denorm/link.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+
+def _get_denorm_prop(
+    name: str,
+    prop: Property,
+    model: Model,
+) -> Property:
+    models = model.manifest.models
+    name_parts = name.split('.', 1)
```
sirex, line 25: I just realized, that support for denormalized data was implemented incorrectly:

https://github.com/atviriduomenys/spinta/blob/086fdf1c8a435ede2578d64b5d146679ef757346/spinta/manifests/tabular/helpers.py#L498-L500

So I think we need to fix how denormalized data are read first.

When reading property name that contains a dot in name, for example:

    a.b.c.d  type=string

Tabular reader should produce following metadata:

```python
{
    'type': None,
    'properties': {
        'b': {
            'type': None,
            'properties': {
                'c': {
                    'type': None,
                    'properties': {
                        'd': {
                            'type': 'string',
                        },
                    },
                },
            },
        },
    },
}
```

Then we need to merge this into `self.state.model.data['properties'][row['property']]`

Then we will have a proper flow of processing nested properties, instead of splitting dots. Because, there can be multiple situations, how these nested denormalized properties can work, they can point to other denormalized data, they can point to arrays, base model properties, etc.

So a better way to handle this complexity is to use visitor pattern, where we handle each nested property separately without knowing or caring about other nodes or its place in the structure.

Sorry, somehow I missed that in https://github.com/atviriduomenys/spinta/issues/216, but we still need to fix that, before going any further.

sirex:
Found another related places, that needs to be fixed:

https://github.com/atviriduomenys/spinta/pull/370#discussion_r1137298609

sirex (approved):

PR #351, opened 2023-02-01
325 add support for base metadata in dsa

adp-atea:


spinta/manifests/tabular/helpers.py
```diff
@@ -1446,6 +1457,22 @@ def _resource_to_tabular(
...
 
+def _base_to_tabular(
+    base: Base,
+    *,
+    end_of_base: bool = False
+) -> Iterator[ManifestRow]:
+    if end_of_base:
+        yield torow(DATASET, {
+            'base': '/',
+        })
```
sirex, line 1468: This is not related to base, since any dimension can have end marker. So it would be more correct to move this to a separate function `_end_marker(name)`, which would produce end marker for any given dimension by name.

spinta/manifests/tabular/helpers.py
```diff
@@ -1623,6 +1651,12 @@ def datasets_to_tabular(
         else:
             separator = False
 
+        if model.base and (not base or model.base.name != base.name):
+            base = model.base
+            yield from _base_to_tabular(model.base)
+        elif base and not model.base:
+            base = None
+            yield from _base_to_tabular(model.base, end_of_base=True)
```
sirex, line 1659: Please add more universal recognition of dimension end marker. End marker should be added for any node, which is not in context of previous node and is in a lower dimension.

For example if model comes after base, resource, dataset and is not in context of previous base, resource and dataset, then before model, and end marker should be added to the last node, that this model is not part of.

For example:

```
dataset
  resource
    base
/
      model
```
Here model does not belong to any previous nodes, so before adding model, and end marker should be added.

Probably a separate function could be added to detect that.

sirex (changes requested):

spinta/manifests/tabular/helpers.py
```diff
@@ -1798,7 +1832,7 @@ def _write_csv(
     rows: Iterator[ManifestRow],
     cols: List[ManifestColumn],
 ) -> None:
-    with path.open('w') as f:
+    with path.open('w', newline='') as f:
```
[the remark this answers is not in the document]
adp-atea: Pushed it on accident. It was needed for some tests to pass on Windows.

spinta/manifests/tabular/helpers.py
```diff
@@ -398,7 +402,11 @@ def read(self, row: Dict[str, str]) -> None:
             'type': 'model',
             'id': row['id'],
             'name': name,
-            'base': base.name if base else None,
+            'base': {
+                'name': base.name,
+                'parent': base.data['model'],
+                'pk': base.data['pk'],
+            } if base and base.data else None,
```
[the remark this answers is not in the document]
adp-atea: Bases are loaded specifically for models. As I understand it's similar to model's external (Entity) attribute. So bases are not in manifest.objects. For example, if we have City and Country and they both inherit from Location, two Base objects are created, one for City and one for Country, but with the same base name. I can try to add bases to objects, but is it correct, if Base is specific for model?

spinta/manifests/tabular/helpers.py
adp-atea: Couldn't find universal way to do this, because you need to check different conditions with different dimensions. For now I added option to add "/" to database, resource and base. If this is not a good way, maybe there should be a separate task, because this one already took longer than expected.

sirex (approved):

PR #354, opened 2023-02-02
330 add srid to geometry types on write

adp-atea:


spinta/backends/postgresql/commands/write.py
```diff
@@ -22,7 +22,11 @@ async def insert(
     table = backend.get_table(model)
     async for data in dstream:
         patch = commands.before_write(context, model, backend, data=data)
-
+        if 'coordinates' in patch.keys():
+            if 'SRID' not in patch['coordinates']:
+                coordinates = cast(sa.Column, backend.tables[model.name].columns['coordinates'])
+                srid = coordinates.type.srid
+                patch['coordinates'] = 'SRID='+str(srid)+';'+str(patch['coordinates'])
```
sirex, line 29: This should not be gere, all geometry related code should go under geometry related command methods.

spinta/backends/postgresql/commands/write.py
```diff
@@ -22,7 +22,11 @@ async def insert(
     table = backend.get_table(model)
     async for data in dstream:
         patch = commands.before_write(context, model, backend, data=data)
-
+        if 'coordinates' in patch.keys():
```
sirex, line 25: What if a property is not named `coordinates`? See my testing notes, I named it `point`.

tests/dtypes/test_geometry.py
```diff
@@ -150,6 +149,25 @@ def test_geometry_params_srid(
...
+    assert listdata(resp, full=True) == [
+        {
+            'name': "Vilnius",
+            'coordinates': 'POINT (582710 6061887)',
+        },
+        {
+            'name': "Vilnius",
+            'coordinates': 'POINT (582710 6061887)',
+        }
+    ]
```
sirex, line 169: Please, move it to another test, that tests case when SRID is given in geometry itself.

sirex (changes requested):

spinta/backends/postgresql/commands/write.py
```diff
@@ -133,3 +132,22 @@ def after_write(
...
+def before_write(
+    context: Context,
+    dtype: Geometry,
+    backend: PostgreSQL,
+    *,
+    data: DataSubItem):
+    patch = {}
+    if 'SRID' not in data.patch and dtype.srid is not None:
+        value = {dtype.prop.name: 'SRID='+str(dtype.srid)+';'+str(data.patch)}
+        patch.update(value)
```
sirex, line 147:
```suggestion
        patch = {dtype.prop.name: 'SRID='+str(dtype.srid)+';'+str(data.patch)}
```

spinta/backends/postgresql/commands/write.py
```diff
@@ -133,3 +132,22 @@ def after_write(
...
+    else:
+        value = {dtype.prop.name: str(data.patch)}
+        patch.update(value)
```
sirex, line 150:
```suggestion
        patch = {dtype.prop.name: str(data.patch)}
```

sirex (changes requested):

tests/dtypes/test_geometry.py
```diff
@@ -287,3 +277,46 @@ def test_loading(tmp_path: Path, rc: RawConfig):
...
+):
+    context = bootstrap_manifest(rc, f'''
+        d | r | b | m | property                | type           | access
+        backends/postgres/dtypes/geometry/srid  |                | 
+          |   |   | Point                       |                | 
+          |   |   |   | point                   | geometry       | open
+          |   |   | PointLKS94                  |                | 
+          |   |   |   | point                   | geometry(3346) | open
+          |   |   | PointWGS84                  |                | 
+          |   |   |   | point                   | geometry(4326) | open
```
sirex, line 295: Why you need `Point` and `PointWGS84` if you only test with `PointLKS94`?

tests/dtypes/test_geometry.py
```diff
@@ -287,3 +277,46 @@ def test_loading(tmp_path: Path, rc: RawConfig):
...
+
+    resp = app.post(f'/{model}', json={
+        'point': 'POINT(582710 6061887)',
+    })
+    assert resp.status_code == 201
+
+    resp = app.post(f'/{model}', json={
+        'point': 'POINT(582710 6061887)',
+    })
+    assert resp.status_code == 201
```
sirex, line 320: Why you need to add exactly same thing, two times?

sirex (changes requested):

tests/dtypes/test_geometry.py
adp-atea: Missed that.

sirex (approved):

PR #370, opened 2023-02-22
208 add possibility to reference external model

adp-atea:


spinta/types/model.py
```diff
@@ -381,11 +384,30 @@ def check(context: Context, model: Model):
...
+def check(context: Context, dtype: Ref):
+    if dtype.prop.level:
+        level = int(dtype.prop.level)
+        ref_external = dtype.model.external
+        external = dtype.prop.model.external
+        if ref_external and external and \
+                ref_external.resource and \
+                ref_external.resource != external.resource and \
+                level > 3:
+            raise exceptions.InvalidPropertyLevel(dtype.prop.model, prop=dtype.prop.name)
```
sirex, line 408:
```suggestion
            raise exceptions.InvalidPropertyLevel(dtype)
```

Exceptions should always use deepest node possible, higher nodes will be populated to error context automagically.

spinta/types/model.py
```diff
@@ -381,11 +384,30 @@ def check(context: Context, model: Model):
...
+@check.register(Context, Ref)
+def check(context: Context, dtype: Ref):
+    if dtype.prop.level:
+        level = int(dtype.prop.level)
+        ref_external = dtype.model.external
+        external = dtype.prop.model.external
+        if ref_external and external and \
+                ref_external.resource and \
+                ref_external.resource != external.resource and \
+                level > 3:
```
sirex, line 407: Now I realized, that checking external resource is a bad idea. Instead we should check that ref property level is not higher than referenced models pkeys. Basically we should check `zip(dtype.refprops, dtype.model.external.pkeys)` and each a.level <= b.level.

sirex (changes requested):

spinta/utils/data.py
```diff
@@ -71,6 +74,8 @@ def take(keys, *args):
     for k in keys:
         for v in args:
             for x in k:
+                if v and not isinstance(v, Iterable):
+                    raise InvalidNestedValue(value=v)
```
sirex, line 78: We should not use higher level components in a lower level function. Exceptions from spinta.exceptions are high level exceptions and spinta.utils is a set of low level functions.

Since this is probably related to case when `ref` type receives a scalar value, then this kind of check should be done in a `simple_data_check` command.

See the command flow below:

image

sirex (changes requested):

spinta/backends/__init__.py
```diff
@@ -229,6 +231,43 @@ def simple_data_check(
...
+@commands.simple_data_check.register(Context, DataItem, Ref, Property, Backend, object)
+def simple_data_check(
+    context: Context,
+    data: DataItem,
+    dtype: Ref,
+    prop: Property,
+    backend: Backend,
+    value: object,
+) -> None:
+    if value and not isinstance(value, (dict, str)):
```
sirex, line 267: Why `str` is allowed too?

sirex (approved):

spinta/backends/__init__.py
adp-atea: The test that fails without str is [`test_manifest_loading`](https://github.com/atviriduomenys/spinta/blob/95dbf1c2949b995deae2a7141453fad57a006bc8/tests/test_manifests.py#L28). From what I understand, during `cli.invoke(rc, ['migrate'])`, [`update_schema_version`](https://github.com/atviriduomenys/spinta/blob/95dbf1c2949b995deae2a7141453fad57a006bc8/spinta/manifests/backend/helpers.py#L232) method is called and `version` attribute is passed as string.

sirex: In this case we definately should fix the `update_schema_version` and here we should expect `value` to be `dict`.

PR #376, opened 2023-02-27
359 rewrite ascii table formatter

adp-atea:


spinta/formats/ascii/helpers.py
```diff
@@ -1,11 +1,207 @@
...
+            break
+        else:
+            for col in cols:
+                value = row.get(col)
+                if value:
+                    value = str(value)
+                    if len(value) > max_value_length:
+                        value = value[:max_value_length] + "..."
+
+                    if '\n' in value:
```
sirex, line 28: Use `str.splitlines`, there are multiple line separators in different platforms.

tests/formats/test_ascii.py
```diff
@@ -108,28 +109,30 @@ async def test_export_multiple_types(rc: RawConfig):
...
-        'example/C  ∅    ∅          1\n'
-        'example/C  ∅    ∅          2\n'
-        'example/C  ∅    ∅          3\n'
-        '---------  ---  ---------  -----'
+        '---------  ---  ---------  -----  \n'
+        '_type      _id  _revision  value  \n'
+        'example/C  ∅    ∅          1    \n'
+        'example/C  ∅    ∅          2    \n'
+        'example/C  ∅    ∅          3    \n'
+        '---------  ---  ---------  -----  \n'
```
sirex, line 135: Maybe it is possible to avoid spaces at the end of line? This will save some bytes transferred data.

spinta/formats/ascii/commands.py
```diff
@@ -47,18 +47,30 @@ def _render(
     status_code,
     headers,
 ):
-    # Format params ar given in RQL query `?format(width(1),colwidth(1))`.
+    # Format params ar given in RQL query `?format(width(1),max_col_width(1))`.
     width = params.formatparams.get('width')
-    colwidth = params.formatparams.get('colwidth')
+    max_col_width = params.formatparams.get('max_col_width')
```
sirex, line 52:
```suggestion
    max_col_width = params.formatparams.get('colwidth')
```

This is user facing API and should be backwards compatible.

spinta/formats/ascii/commands.py
```diff
@@ -47,18 +47,30 @@ def _render(
...
+    max_value_length = params.formatparams.get('max_value_length', 100)
```
sirex, line 53:
```suggestion
    max_value_length = params.formatparams.get('vlen', 100)
```

This is user facing api and should use similar naming convention as in other places.

sirex (approved):

spinta/formats/ascii/commands.py
```diff
@@ -47,18 +47,30 @@ def _render(
...
+    rows_to_read = params.formatparams.get('rows_to_read', 200)
```
sirex, line 54: This parameter should not be exposed to api, because some bad actors could use this for memory attacks.

sirex (approved):

PR #402, opened 2023-03-22
366 paginated push

adp-atea:


adp-atea:
Added row reading with pages, but there is a problem. If we use `push` command one time and `page` values are saved in state, then next time when `push` command is called, we only read rows with values that are greater than saved `page` value. The problem with that is that in the beginning we reset all `pushed` values to `None`, to know which rows were deleted. And because now we read rows starting from `page` value and skip the previous ones, those rows are left with `pushed = None` and are added to delete list.

sirex (changes requested):
A good start, but in order to fix the issue with deletes and also to fix backwards compatibility and other issues, few changes have to be made.

First, pagination must be exposed not only at commands level, but also pagination should be available at API level. For example, I should be able to do:

    http get /example/City?page(5)

And this, should only return a single page, starting at 5.

In order to avoid clashed with http query `page`, we need to load `page` from `prepare` at `load` time, using a separate `LoadBuidler` instead of `*QueryBuilder`.

`LoadBuider` should set `model.page.by` and `model.page.size`.

Similarly we should resolve query parameters (`?page(5)`), in `_prepare_urlparams_from_path`:

https://github.com/atviriduomenys/spinta/blob/f56b5a56ba88be00f995cc6a55f8b63064b9f947/spinta/urlparams.py#L92

Except instead of setting `model.page`, it should set `params.page` and instead of setting `page.by`, should set `page.value`, then we can do `model.page.by > params.page.value`.

`params.page.value` should be passed to `getall` as `page` argument in:

https://github.com/atviriduomenys/spinta/blob/f56b5a56ba88be00f995cc6a55f8b63064b9f947/spinta/commands/read.py#L62

And I guess, we can handle pagination form main `getall` command. If `params.page` is given, then we just pass the given page, but if `params.page` is not given, then we can do some thing like this:

```python
if model.page:
    rows = paginate(context, model, backend, params.page, expr, params.limit)
else:
    rows = commands.getall(context, model, backend, query=expr)
```

Where `paginate` would do something like this:

```python
def paginate(
    context: Context,
    model: Model,
    backend: Backend,
    page: Page,
    expr: Expr | None,
    limit: int | None,
) > Iterator[ObjectData]:
    size = page.size or model.page.size
    value = FIRST_PAGE
    while True:
        rows = commands.getall(context, model, backend, query=expr, page=value, limit=size)
        for row in rows:
            value = row.pop('_page') # _page must be always added
            yield row
        # Also we should add a calculated limit, instead of relying on
        # backend limit functionality.
```

Keep in mind, that pagination can be done by single or multiple values. So we need to construct query correctly. Also page fields, can be ordered, for example `a,-b`, where `a` is sorted in ascending order and `b` in descending. Pagination fields, can be of different types, not all types can be supported.

Also, probably it would be a good idea, to move query builder out of backend `getall` commands, so that we could reuse already prebuilt query for each page and only modify it, to add pagination.

On internal backends, I think we can always use `_id` as `page.by`, ignoring what is specified on `model.prepare` and only use `model.prepare` on external backends. Also on external backends, if `page` is not given on `model.prepare`, then we should use `model.ref` as `page.by`.

if `sort` is given, then we should paginate by sort keys + `page.by`. Also, for sort keys, we should respect sort direction, for sort keys in ascending order we should do `key > value` and for descending order we should do `key < value`.


Once we have a working pagination at backend level, then for push, first instead of adding `--full`, we should add `-i, --incremental`, because currently by default we do full push, so we should leave that for backwards compatibility.

Also, it would be nice to have a `push  --model example/CIty --page VALUE` flag, to push only specific page.

Now regarding push state, since we always read data sordet by `page.by`, it is easy to compare, what we have in stead and what is on the data source, without cleaning push state.

We need to read from both, data source and from push state and compare rows side by side, so that we have something like this:

| source._page | state._page | action |
| ------------ | ----------- | ------ |
| 1            | 1           | patch  |
| 2            | -           | insert |
| -            | 3           | delete |

We just need to compare both sides and if page values are not equal on both sides, take row from side having lower value and skip reading rows from other side. And as shown in the table above, you can easily decide if you need to delete things or not.

For this to work, we need to save page value  in push state. Pages values in push state could be stored as json serialized string. If page has more than one filed, then we should serialize values as json array.

In order to avoid storing all data if `model.ref` is not specified, page values should be truncated up to 40 symbols.

Also, we should raise and error, if we receive duplicate `source._page`, because that breaks all the idea.

sirex:
```python
{
  'name': 'and', 'args': [
    {'name': 'ge', 'args': [
      {'name': 'bind', 'args': ['a']},
      42
    ]}
  ]
}
```

adp-atea:
There are still quite a few failing tests. Most of them are from test_sql.py. The problems that I found:

1. If we use SQL backend and model.page.by property doesn't have prop.external.name, we get NoExternalName 
error. Error comes from get_column method which is used in sort. 

2. Another problem is when property in model.page.by is private, because then it gets filtered out and it's 
not returned in the row, so we cannot assign it to page.value.

3. Also if we sort by two or more properties, they need to ascend or descent together.
For example if model.page.by = [id1, id2] and we have data:
id1 | id2
 1   |  1
 2   |  3
 3   |  9
 5   | 10
This would work. But if we have data:
id1 | id2
 1   |  20
 2   |  3
 3   |  6
 5   | 10
This wont work, because the first compare query would be (id1 > 1 & id2 > 20) and it wouldn't return
any more rows.

4. Also for --page VALUE flag, for now I made it to work, when VALUE and model.page.by is only one
attribute. I'm not sure how this should be solved with multiple attributes.

sirex:
> If we use SQL backend and model.page.by property doesn't have prop.external.name, we get NoExternalName
error. Error comes from get_column method which is used in sort.

Could you give a link to a test where this error appears?

> Another problem is when property in model.page.by is private, because then it gets filtered out and it's
not returned in the row, so we cannot assign it to page.value.

 I'm thinking, that probably, best way, to do pagination is to stick to sorting and pagination by `_id`. In case of external backends, we still accept `_id` for pagination, bet `_id` is resolved to `model.ref` properties, via keymap. So even if som `model.ref` properties are private, we still should be able to do sorting and pagination, without exposing any private values, because via API, we should still see only `_id`, mapped to `model.ref` properties via keymap.

So private properties should not be an issue.

> Also if we sort by two or more properties, they need to ascend or descent together.
For example if model.page.by = [id1, id2] and we have data:
id1 | id2
1 | 1
2 | 3
3 | 9
5 | 10
This would work. But if we have data:
id1 | id2
1 | 20
2 | 3
3 | 6
5 | 10
This wont work, because the first compare query would be (id1 > 1 & id2 > 20) and it wouldn't return
any more rows.

For now, in case if we have a user request sort, just do pagination, bet first sort key. There might be cases when first sort key returns very large pages


In cases, when we have a user requested sort, we should always start pagination by first sort key, when we retrieve first page from backend and we have same pagination key in whole page, then we add another sort key like this `id1 = 1 & id2 > 20`. And we should repeat this, for all sort keys.

For example, if we have following data:

a | b
---: | ---:
1 | 20
1 | 21
1 | 22
1 | 23
2 | 1
2 | 2
2 | 3

If page size is set to 2, the we do:

- `ORDER BY a, b, _id LIMIT 3` - get first page, use only `a` as page key, since this returns `[(1, 20), (1, 21), (1, 22)]`, where page key `a` always the same, we consider this as incomplete page and add next sort key, to our page key. Even if we have page size set to 2, we do `LIMIT 3`, by adding extra row, to know if we need get back so single key page or not.
- `WHERE a = 1 AND b > 21 ORDER BY a, b, _id LIMIT 3` - get next page, now this returns `[(1, 22), (1, 23)]`, since we got less than 3 rows, we know, that `a` key is exhausted, and we can continue with single page key.
- `WHERE a > 1 ORDER BY a, b, _id LIMIT 3` - get next page, where we receive `[(2, 1), (2, 2), (2, 3)]`, we see again, that `a` has same value, then we add next sort key to our page key and continue.
- ...

By adding extra row to the page, we know, then we need to add extra key to our page key, but if at least one value is different in our page key, we know, that whole page key fits into one page and we can operate only using first sort key.

> Also for --page VALUE flag, for now I made it to work, when VALUE and model.page.by is only one
attribute. I'm not sure how this should be solved with multiple attributes.

For multiple values, we should use `--page VALUE1 --page VALUE2 --page VALUE3`.

https://typer.tiangolo.com/tutorial/multiple-values/multiple-options/

notes/push/page.sh
```diff
@@ -0,0 +1,292 @@
...
+#| WzgsICJNYVx1MDE3ZWVpa2lhaSIsICJhYzFlNGQyNC1iNmZmLTRmNjMtOGNjMy03Mjc3NjA1NjM5NDEiXQ==      8   Mažeikiai
+#| WzksICJKb25hdmEiLCAiMzc0Yjg2MmYtZDAxZS00MmFjLWJjYWQtMGNhOGY5N2FmZjE1Il0=                  9   Jonava
+#| WzEwLCAiVXRlbmEiLCAiNzczMjdjOTktNmMwNi00NDc2LWEwYTUtNGVmM2ZmMGQxZmI3Il0=                  10  Utena
+
+
+echo "WzEsICJWaWxuaXVzIiwgImRjZjI5MWNlLTgzNWQtNDUxNy04NWNiLTBlZjAyZTJiYmEwYyJd" | base64 --decode
+#| [1, "Vilnius", "dcf291ce-835d-4517-85cb-0ef02e2bba0c"]
+# XXX: Here I run Spinta with internal mode, whre page key should always be
+#      _id, but here I get more than _id. Shouldn't this be change to use only
+#      _id?
```
sirex, line 248: Shouldn't internal backend use only _id as page key, if sort is not used?

adp-atea: it is done this way, so that internal backend would still abide by given manifest page()

PR #404, opened 2023-03-24
364 replace frictionless-py with sqlachemy

adp-atea:


adp-atea:
There are failing tests, that I'm not sure how to fix. I commented them out for now. 

One problem is that some use sqldump as resource and we try to get manifest type from this resource ([here](https://github.com/atviriduomenys/spinta/blob/61cc3d321938770c903bf7d282d7c1d8111907f8/spinta/cli/inspect.py#L85)), but there is no sqldump manifest, there's only sqldump backend.

Another problem is when we have existing manifest and there's database related to it and added to the context ([example](https://github.com/atviriduomenys/spinta/blob/61cc3d321938770c903bf7d282d7c1d8111907f8/tests/test_inspect.py#L352))
but it is not added as resource with -r to command. In this case when manifest is loaded, it's loaded only from the file and related database is not taken into account.

sirex (changes requested):
When I run

    poetry run spinta --tb=native inspect -r sql sqlite:///$BASEDIR/db.sqlite -o $BASEDIR/manifest.csv

I get following error:

```python
  File "spinta/cli/inspect.py", line 65, in inspect
    store = load_manifest(context, ensure_config_dir=True)
  File "spinta/cli/helpers/store.py", line 99, in load_manifest
    store = load_store(
  File "spinta/cli/helpers/store.py", line 85, in load_store
    commands.load(context, store)
  File "spinta/types/store.py", line 50, in load
    manifest = create_manifest(context, store, manifest)
  File "spinta/manifests/helpers.py", line 68, in create_manifest
    Manifest_ = config.components['manifests'][mtype]
KeyError: 'memory'
```

You need to also add `memory` manifest, which would be responsible only for loading other manifests.

See: https://spinta.readthedocs.io/en/latest/contrib/manifests.html#adding-new-manifest-type

For `load` command you need to only add:

```python
    if load_internal:
         target = into or manifest
         if '_schema' not in target.models:
             store = context.get('store')
             commands.load(context, store.internal, into=target)

     for source in manifest.sync:
         commands.load(
             context, source,
             into=into or manifest,
             freezed=freezed,
             rename_duplicates=rename_duplicates,
             load_internal=load_internal,
         )
```

notes/cli/inspect.sh
```diff
@@ -0,0 +1,194 @@
...
+#|    |   |   |   |   | name       | string  |         | name                                          |         |       |        |     |       |
+#|    |   |   |   |   | continent  | integer |         | continent_id                                  |         |       |        |     |       |
+#|    |   |   |   |   | population | integer |         | population                                    |         |       |        |     |       |
+#|    |                            |         |         |                                               |         |       |        |     |       |
+#|    |   |   |   | Continent      |         | id      | continents                                    |         |       |        |     |       |
+#|    |   |   |   |   | id         | integer |         | id                                            |         |       |        |     |       |
+#|    |   |   |   |   | name       | string  |         | name                                          |         |       |        |     |       |
+# FIXME: `continent` type was lost, types specified in manifest table should be preserved.
```
sirex, line 164: Property types specified in manifest file, should be preserved even if type in data source has changed. For example here, foreign key was not specified in data source, but it was specified in manifest file, so after updating manifest from source, we should preserve foreign key in manifest file.

sirex (changes requested):

spinta/cli/inspect.py
```diff
@@ -37,33 +57,767 @@ def inspect(
     auth: Optional[str] = Option(None, '-a', '--auth', help=(
         "Authorize as a client"
     )),
+    priority: str = Option('manifest', '-p', '--priority', help=(
+        "Merge property priority ('manifest' or 'external')"
+    ))
 ):
     """Update manifest schema from an external data source"""
+    if priority not in ['manifest', 'external']:
+        raise Exception(f"\'{priority}\' priority doesnt not exist, there can only be \'manifest\' or \'external\'")
```
sirex, line 66: I think, this should not be an exception, but a simple error message, printed with `print`.

spinta/cli/inspect.py
```diff
@@ -37,33 +57,767 @@ def inspect(
...
+    if resource.external:
+        result = resource.external
+    else:
+        if manifest.backends:
+            if resource.backend.name in manifest.backends:
+                result = manifest.backends[resource.backend.name].config['dsn']
+        elif manifest.store.backends:
+            if resource.backend.name in manifest.store.backends:
+                result = manifest.store.backends[resource.backend.name].config['dsn']
+    if result.__contains__("@"):
```
sirex, line 807:
```suggestion
    if "@" in result:
```

spinta/cli/inspect.py
```diff
@@ -37,33 +57,767 @@ def inspect(
...
+        result = result.split("@")[1]
+    return result
+
+
+def _backend_dsn(backend: ExternalBackend) -> str:
+    dsn = backend.config['dsn']
+    if dsn.__contains__("@"):
```
sirex, line 814:
```suggestion
    if "@" in dsn:
```

sirex (approved):

PR #415, opened 2023-03-30
149 push ref dependencies

### LINKED ISSUE #149: Push ref dependencies

adp-atea:


spinta/commands/write.py
```diff
@@ -864,6 +865,68 @@ async def prepare_data_for_write(
...
+    async for data in dstream:
+        payload = data.payload
+        filtered_properties = [payload.get(key) for key in self_reference_properties]
+        if all(value is None for value in filtered_properties) or action not in [Action.INSERT, Action.UPSERT]:
+            yield data
+        else:
+            data_with_self_reference[payload.get("_id")] = {
+                "data": data,
+                "self_reference": filtered_properties
+            }
```
sirex, line 891: Unfortunately this won't work with large datasets, because as I understand, here all table data is loaded into memory and then sorted?

But of course this would be a good solution if all datasets would fit into memory.

If data would be sorted using SQL query this would work. But in order to sort data using SQL we need a recursive CTE, see https://stackoverflow.com/questions/59137122/sql-query-for-topological-sort

sirex (changes requested):

SauliusPetraitis:
blogs sprendimas, atmesta senai, bet pr likes

PR #419, opened 2023-04-05
411 add support for older SQLite versions

adp-atea:


sirex (changes requested):
I think, current approach will soon get out of hands, since more changes to state table are coming soon, so we need a better solution, to handle schema changes.

Moved sqlite migrations to a separate module and instead of using raw sql, switched to alembic instead.

And for older sqlite version, I think better solution, to just copy whole table, instead of single column, because drop column feature is also quite new in sqlite.

Please, finish what I started and add tests for all cases, we currently need.

tests/utils/test_sqlite.py
```diff
@@ -0,0 +1,149 @@
+from typing import Dict
+from typing import List
+
+import sqlalchemy as sa
+from sqlalchemy.engine.reflection import Inspector
+
+from spinta.utils.sqlite import migrate_table
+
+
+def test_create_table():
```
sirex, line 10: Please add parametrization to all tests, to test both ways, with alter table and with copy table. For this to work, `migrate_table` should accept a `copy` argument if `copy=True`, then `_migrate_with_insert_from_select` is used, if `copy=False`, then `_migrate_with_alter_table` is used.

sirex (changes requested):

tests/utils/test_sqlite.py
sirex: Please use parameterization:

https://docs.pytest.org/en/6.2.x/parametrize.html

Instead of adding multiple copies of same tests.

sirex (changes requested):

spinta/utils/sqlite.py
```diff
@@ -0,0 +1,116 @@
...
+    renames = renames or {}
+
+    with engine.begin() as conn:
+        ctx = MigrationContext.configure(conn)
+        op = Operations(ctx)
+
+        if (
+            isinstance(engine.dialect, SQLiteDialect) and
+            engine.dialect.server_version_info < (3, 36) and
+            copy
```
sirex, line 37: Should be `or copy`.

sirex (changes requested):

spinta/utils/sqlite.py
```diff
@@ -0,0 +1,116 @@
...
+        return
+
+    renames = renames or {}
+
+    with engine.begin() as conn:
+        ctx = MigrationContext.configure(conn)
+        op = Operations(ctx)
+
+        if (
+            renames and
```
sirex, line 35: Why renames is included here?

sirex (approved):

PR #420, opened 2023-04-05
363 handle foreign key constraint errors

adp-atea:


notes/types/ref/constraints.sh
```diff
@@ -0,0 +1,121 @@
...
+#|                 "property": "twin",
+#|                 "schema": "8"
+#|             },
+#|             "message": "Unknown property 'twin'.",
+#|             "template": "Unknown property {property!r}.",
+#|             "type": "model"
+#|         }
+#|     ]
+#| }
+# FIXME: This should not be an error.
```
sirex, line 121: This should not be an error.

sirex (changes requested):

spinta/backends/postgresql/commands/write.py
```diff
@@ -20,21 +23,50 @@ async def insert(
...
+                _revision=patch['_revision'],
+                _txn=transaction.id,
+                _created=utcnow(),
+            )
+            connection.execute(qry, patch)
+            commands.after_write(context, model, backend, data=data)
+        except exc.IntegrityError as error:
+            rollback_full = True
+            savepoint.rollback()
+            if type(error.orig) is psycopg2.errors.ForeignKeyViolation:
```
sirex, line 48:
```suggestion
            if isinstance(error.orig, psycopg2.errors.ForeignKeyViolation):
```

spinta/backends/postgresql/commands/write.py
```diff
@@ -20,21 +23,50 @@ async def insert(
...
+                error_message = error.orig.diag.message_detail
+                error_property_name = re.search('Key \((.*?)\)', error_message).group(1).split(".")[0]
+                error_ref_id = re.search('\)=\((.*?)\)', error_message).group(1)
+                error_property = model.properties.get(error_property_name)
```
sirex, line 52: Please move this and other similar cases of extracting property name and value into separate function and add tests with example error messages.

Error messages are likely to change, so it will be easier to maintain if such changes do occur.

spinta/backends/postgresql/commands/write.py
```diff
@@ -20,21 +23,50 @@ async def insert(
...
+            )
+            connection.execute(qry, patch)
+            commands.after_write(context, model, backend, data=data)
+        except exc.IntegrityError as error:
+            rollback_full = True
+            savepoint.rollback()
+            if type(error.orig) is psycopg2.errors.ForeignKeyViolation:
+                error_message = error.orig.diag.message_detail
+                error_property_name = re.search('Key \((.*?)\)', error_message).group(1).split(".")[0]
+                error_ref_id = re.search('\)=\((.*?)\)', error_message).group(1)
```
sirex, line 51: Use precompiled regular expressions (`re.compile`). Parsing regular expression string is pretty expensive operation.

sirex (approved):

Issue #363, opened 2023-02-20
Handle foreign key constraint errors


sirex:
Currently foreign key constraints are handled only at postgresql level and if there is an error, client receives a 500 error, but instead a 400 error should be returned with a proper error message.

### Related

- https://github.com/atviriduomenys/spinta/issues/149
- https://github.com/atviriduomenys/spinta/issues/311
- https://github.com/atviriduomenys/spinta/issues/148

PR #421, opened 2023-04-06
371 Added UniqueConstraint for external pkeys

adp-atea:
This is not finished, still need to add Error message when triggering IntegrityError. IntegrityError handle has been started in task 363: https://github.com/atviriduomenys/spinta/issues/363

There I have created try except around insert, update and delete. It checks for IntegrityError, if it finds it, it adds it to the MultipleErrors exception. Also if it finds any errors it reverts the transaction. The upside of that implementation, is that it checks all of the data or until it find certain amount of errors, so user gets more informative response.

With current UniqueConstraint error implementation user only gets 1 response.

### DIFF
--- a/spinta/backends/postgresql/commands/init.py
+++ b/spinta/backends/postgresql/commands/init.py
@@ -37,7 +37,10 @@ def prepare(context: Context, backend: PostgreSQL, model: Model):
             columns.extend(column)
         elif column is not None:
             columns.append(column)
-
+    if model.external is not None and model.external.unknown_primary_key is False:
+        if len(model.external.pkeys) == 1:
+            model.external.pkeys[0].dtype.unique = True
+        columns.append(sa.UniqueConstraint(*[key.name for key in model.external.pkeys]))
     # Create main table.
     main_table_name = get_pg_name(get_table_name(model))
     pkey_type = commands.get_primary_key_type(context, backend)
@@ -49,7 +52,6 @@ def prepare(context: Context, backend: PostgreSQL, model: Model):
         *columns,
     )
     backend.add_table(main_table, model)
-
     # Create changes table.
     changelog_table = get_changes_table(context, backend, model)
     backend.add_table(changelog_table, model, TableType.CHANGELOG)


spinta/manifests/tabular/helpers.py
```diff
@@ -1550,6 +1550,9 @@ def _property_to_tabular(
     if prop.access < access:
         return
 
+    remember_unique_state = prop.dtype.unique
+    if len(prop.model.external.pkeys) == 1:
+        prop.dtype.unique = False
```
sirex, line 1555: Please move this into `dtype.get_type_repr`, because this method is responsible for producing correct  type string. Also it would be nice, to move `get_type_repr` out of `dtype`, because, this representation only needed for tabular manifests, so it should be part of tabular manifest, not a high level method.

tests/backends/postgresql/commands/test_init.py
```diff
@@ -32,3 +32,30 @@ def test_prepare(rc: RawConfig):
...
+    assert any(
+        [table.c['id']] == list(constraint.columns) for constraint in table.constraints if
+        type(constraint).__name__ == 'UniqueConstraint')
+
+    model_multiple_unique = manifest.models['example/Country']
+    commands.prepare(context, backend, model_multiple_unique)
+    table = backend.get_table(model_multiple_unique)
+    assert any(
+        [table.c['id'], table.c['name']] == list(constraint.columns) for constraint in table.constraints if
+        type(constraint).__name__ == 'UniqueConstraint')
```
sirex, line 61: Please, add tests for handling unique constraint errors, when trying to write a duplicate value on a column or columns, that have unique constraints.

sirex (changes requested):

notes/backends/postgresql/pkeys/unique.sh
```diff
@@ -0,0 +1,180 @@
...
+#|                     '_revision': '8de45de1-577d-4ef7-ba50-7480b5d7934f',
+#|                     'id': 1,
+#|                     'code': '01',
+#|                     'title': 'T01'
+#|                 }
+#|             "
+#|         }
+#|     ]
+#| }
+# FIXME: This should be a 400 error.
```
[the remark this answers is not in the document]
sirex: Same error here: https://github.com/atviriduomenys/spinta/pull/343#discussion_r1176076447

spinta/manifests/tabular/helpers.py
sirex: Same issue here: https://github.com/atviriduomenys/spinta/pull/343#discussion_r1176129421

sirex:
Please, do not move pull request to review, if there are unresolved comments.

sirex (approved):

PR #422, opened 2023-04-11
400 external ref to model without primary key

adp-atea:


tests/dtypes/test_external_ref.py
```diff
@@ -75,3 +75,51 @@ def test_external_ref(
...
+    d | r | b | m | property | type   | ref                           | source       | level | access
+    datasets/externalref     |        |                               |              |       |
+      | external             | sql    |                               | sqlite://    |       |
+      |   |   | Country      |        |                               |              |       |
+      |   |   |   | code     | string |                               |              |       | open
+      |   |   |   | name     | string |                               |              |       | open
+    datasets/internal/pk     |        |                               |              |       |
+      |   |   | City         |        |                               |              |       |
+      |   |   |   | name     | string |                               |              |       | open
+      |   |   |   | country  | ref    | /datasets/externalref/Country |              | 3     | open
```
sirex, line 95: Please add tests with `ref: Country[id]`. See my testing notes, with `CityExplicit`. If referenced model does not have `model.ref` set, but referenced column is given in `[]`, then we should used referenced columns in `[]`.

sirex (changes requested):

sirex (approved):

Issue #262, opened 2022-06-02
Write multiple objects on a single request


sirex:
Add support for writing multiple objects on a single request. For example:

| model | property | type   | ref           |
| ----- | -------- | ------ | ------------- |
| Person  |          |        |               |
|       | name     | string |               |
| City  |          |        |               |
|       | name     | string |               |
|       | citizen[]  | ref    | Person       |

```json
{
  "_type": "City",
  "name": "Vilnius",
  "citizen": [
    {"_type": "Person", "name": "Jonas"},
    {"_type": "Person", "name": "Petras"},
  ]
}
```
This way, multiple objects will be created.

sirex:
For updates, each foreign property must contain `_revision`. Also, `_base` model can be updates using this functionality, passing `_base._revision`.

PR #425, opened 2023-04-13
205 writing models with base

adp-atea:
This is not finished, need to wait for #262 in order to add Base modification from Target request. Also need #363 for some of the exception handle. Currently this implementation will disallow users to reference Inherit type properties while inserting or updating objects, also for now when updating or inserting objects that has Inherit type properties will not be shown in response.


sirex (changes requested):

sirex:
Please, resolve all comments if issue raised in comment is solved or add a comment otherwise.

sirex (approved):

PR #430, opened 2023-04-18
423 return error exit code on error for spinta push

adp-atea:


spinta/cli/push.py
```diff
@@ -202,6 +202,9 @@ def push(
             error_counter=error_counter,
         )
 
+        if error_counter.count < max_error_count:
```
sirex, line 205: Maybe, we should raise  error `if error_counter.count > 0`?

sirex (changes requested):
I get 0 exit code, when error comes from

```python
Traceback (most recent call last):t/s]
  File "spinta/cli/helpers/data.py", line 80, in _read_model_data
    stream = peek(stream)
  File "spinta/utils/itertools.py", line 40, in peek
    peek = list(islice(it, 1))
  File "spinta/datasets/backends/sql/commands/read.py", line 90, in getall
    table = backend.get_table(model, table)
  File "spinta/datasets/backends/sql/components.py", line 37, in get_table
    sa.Table(name, self.schema, autoload_with=self.engine)
  <...>
  File "sqlalchemy/engine/reflection.py", line 788, in reflect_table
    raise exc.NoSuchTableError(table.name)
sqlalchemy.exc.NoSuchTableError: cities
```

Exist code should be 1.

sirex (approved):

PR #441, opened 2023-05-09
Added resource check for push

adp-atea:


notes/push/includes.sh
```diff
@@ -0,0 +1,109 @@
...
+poetry run spinta push $BASEDIR/manifest.csv -o test@localhost
+#| 2023-05-12 13:56:41,105 ERROR: Error when sending and receiving data. Model push/includes/City, data:                                                                     
+#|  {
+#|     '_id': '1cacb9c9-4350-4df2-8b5b-678c5669d731',
+#|     '_op': 'insert',
+#|     '_type': 'push/includes/City',
+#|     'country': {'_id': '6fbef53b-efa6-4c8c-b13c-37e78197e089'},
+# FIXME: This should be: 'country': {'code': '6fbef53b-efa6-4c8c-b13c-37e78197e089'}
+#        See: https://github.com/atviriduomenys/spinta/issues/208
+#             https://github.com/atviriduomenys/spinta/blob/66d8696ce5a889ceaad9ad9fe4b8def5e6f16ad8/notes/types/ref/external.sh#L111
```
sirex, line 82: Please, fix this, when pushing a ref with level < 4, we should use different property name. Also, add a test for this case or modify existing one.

sirex (changes requested):

notes/push/includes.sh
```diff
@@ -0,0 +1,155 @@
...
+
+poetry run spinta push $BASEDIR/manifest.csv -o test@localhost
+#| 2023-05-12 13:56:41,105 ERROR: Error when sending and receiving data. Model push/includes/City, data:                                                                     
+#|  {
+#|     '_id': '1cacb9c9-4350-4df2-8b5b-678c5669d731',
+#|     '_op': 'insert',
+#|     '_type': 'push/includes/City',
+#|     'country': {'_id': '6fbef53b-efa6-4c8c-b13c-37e78197e089'},
+# FIXME: This should be: 'country': {'code': '6fbef53b-efa6-4c8c-b13c-37e78197e089'}
+#        See: https://github.com/atviriduomenys/spinta/issues/208
```
sirex, line 90: We can't query a model in external mode, that does not have a source set. This should return 400 error.

notes/push/includes.sh
```diff
@@ -0,0 +1,155 @@
...
+#|     return func(*args, **kwargs)
+#|   File "spinta/backends/postgresql/commands/query.py", line 285, in select
+#|     selected = env.call('select', arg)
+#|   File "spinta/core/ufuncs.py", line 218, in call
+#|     raise UnknownMethod(expr=str(Expr(name, *args, **kwargs)), name=name)
+#| spinta.exceptions.UnknownMethod: Unknown method 'select' with args select(null).
+#|   Context:
+#|     expr: select(null)
+#|     name: select
+# FIXME:
```
sirex, line 155: Push does not work.

sirex (changes requested):

notes/push/includes.sh
```diff
@@ -0,0 +1,146 @@
...
+#|                         push/includes/countries/Country._id,
+#|                         push/includes/countries/Country._revision,
+#|                         push/includes/countries/Country.code,
+#|                         push/includes/countries/Country.name
+#|                     FROM push/includes/countries/Country
+#|             "
+#|         }
+#|     ]
+#| }
+# FIXME: In external mode, this should return 400 error, because source is not set, so we can't get data.
```
sirex, line 83: This should return 400 error with a nice explanation, instead of 500 internal server error.

notes/push/includes.sh
```diff
@@ -0,0 +1,146 @@
...
+#|     selected = env.call('select', arg)
+#|                ^^^^^^^^^^^^^^^^^^^^^^^
+#|   File "/home/sirex/dev/data/spinta/spinta/core/ufuncs.py", line 218, in call
+#|     raise UnknownMethod(expr=str(Expr(name, *args, **kwargs)), name=name)
+#| spinta.exceptions.UnknownMethod: Unknown method 'select' with args select(null).
+#|   Context:
+#|     expr: select(null)
+#|     name: select
+#| 
+#| PUSH: 100%|#############| 1/1 [00:00<00:00, 308.38it/s]
```
sirex, line 146: Please fix the error above.

sirex (changes requested):

tests/cli/test_push.py
```diff
@@ -1,11 +1,13 @@
 import pytest
 import sqlalchemy as sa
+from spinta.core.config import RawConfig
 from spinta.manifests.tabular.helpers import striptable
 from spinta.testing.cli import SpintaCliRunner
 from spinta.testing.data import listdata
 from spinta.testing.datasets import create_sqlite_db
 from spinta.testing.tabular import create_tabular_manifest
-from tests.datasets.test_sql import configure_remote_server, create_rc
+from tests.datasets.test_sql import configure_remote_server, create_rc, create_client
```
sirex, line 9: Do not import functions from other tests. If you need to do that, move function into `spinta.testing`.

tests/types/renamed__init__.py
sirex, line 1: Why do we need this file?

spinta/backends/nobackend/commands/changes.py
```diff
@@ -0,0 +1,18 @@
...
+def create_changelog_entry(
+    context: Context,
+    model: (Model, Property),
+    backend: NoBackend,
+    *,
+    dstream: AsyncIterator[DataItem],
+) -> AsyncIterator[DataItem]:
+    # FileSystem properties don't have change log, change log entry will be
+    # created on property model backend.
+    return dstream
```
sirex, line 18: This command is called from write operations, we don't need to define it from `NoBackend`, because we can't do any write operations if the first place.

spinta/backends/nobackend/commands/wipe.py
```diff
@@ -0,0 +1,10 @@
+from spinta import commands
+from spinta.components import Context
+from spinta.components import Model
+from spinta.backends.nobackend.components import NoBackend
+
+
+@commands.wipe.register(Context, Model, NoBackend)
+def wipe(context: Context, model: Model, backend: NoBackend):
+    if model.name in backend.data:
+        backend.data[model.name] = {}
```
sirex, line 10: Just `pass` and do nothing.

spinta/backends/nobackend/commands/write.py
```diff
@@ -0,0 +1,70 @@
...
+@commands.after_write.register(Context, Model, NoBackend)
+def after_write(
+    context: Context,
+    model: Model,
+    backend: NoBackend,
+    *,
+    data: DataSubItem,
+) -> dict:
+    table = get_table_name(model)
+    raise TableHasNoSource(table)
```
sirex, line 70: `before_write` and `after_write` are called from write operations, bute write operations should raise error, so we don't need to define methods of these commands from `NoBackend`.

spinta/types/namespace.py
```diff
@@ -238,11 +238,16 @@ def traverse_ns_models(
...
     *,
     internal: bool = False,
+    source_check: bool = False
 ):
     models = (ns.models or {})
     for model in models.values():
         if _model_matches_params(context, model, action, dataset_, resource, internal):
-            yield model
+            if source_check:
+                if model.external:
```
sirex, line 247:
```suggestion
                if model.backend and not isinstance(model.backend, NoBacend):
```

sirex (changes requested):
Regarding https://github.com/atviriduomenys/spinta/actions/runs/5832636800/job/15818455159#step:9:184

Just create `push` command method for `NoBackend` and raise error there.

https://github.com/atviriduomenys/spinta/blob/f4756ab3f2c09c11d41be1bf16e9b3729e9b3156/spinta/commands/write.py#L56-L64

PR #443, opened 2023-05-10
336 import rdfs schema

adp-atea:


spinta/components.py
```diff
@@ -516,6 +516,7 @@ class Model(MetaData):
         'params': {'type': 'object'},
         'comments': {},
         'uri': {'type': 'string'},
+        'schema_type': {'type': 'string'},
```
sirex, line 519: There should be no such ting, as `schema_type`, this is only used as metadata to know schema type, when loading, but `load` commands should not receive it.

spinta/components.py
```diff
@@ -592,6 +593,7 @@ class Property(Node):
         'units': {'type': 'string'},
         'lang': {'type': 'object'},
         'comments': {},
+        'schema_type': {'type': 'string'},
```
sirex, line 596: Same gere, there should be no such thing as `schema_type`.

notes/manifests/rdf.sh
```diff
@@ -106,3 +106,37 @@ where {
...
+#|   File "spinta/types/model.py", line 214, in load
+#|     raise UnknownPropertyType(prop, type=data['type'])
+#| spinta.exceptions.UnknownPropertyType: Unknown property type 'None'.
+#|   Context:
+#|     component: spinta.components.Property
+#|     manifest: default
+#|     schema: 2
+#|     model: dcat2/Distribution
+#|     property: accessurl
+#|     type: None
```
sirex, line 142: If type is not given, use `string`. And in general, we should be able to load data structure, that we generate.

sirex (changes requested):

notes/manifests/rdf.sh
```diff
@@ -106,3 +106,9 @@ where {
     bind(str(?prop_title_rdfs) as ?prop_title_rdfs)
 }
 ''')
+
+poetry install
+unset SPINTA_CONFIG
+poetry run spinta inspect -r rdf https://www.w3.org/ns/dcat2.ttl
+#| id | d | r | b | m | property | type | ref | source | prepare | level | access | uri | title | description
+#|    | dcat2                    |      |     |        |         |       |        |     |       |
```
sirex, line 114: Nothing is loaded.

sirex (changes requested):

spinta/manifests/rdf/helpers.py
```diff
@@ -0,0 +1,464 @@
...
+            bind("rdfs" as ?schema)
+        }
+
+        union
+
+        {
+            ?property a rdf:Property .
+            optional { ?property rdfs:isDefinedBy ?description }
+            bind("rdfs" as ?schema)
+        }
```
sirex, line 70: We need to define same thing for `owl:ObjectProperty`:

```ttl
        union
        {
            ?property a owl:ObjectProperty .
            optional { ?property rdfs:label ?title }
            optional { ?property rdfs:comment ?description }
            optional { ?property rdfs:domain ?type }
            bind("owl" as ?schema)
        }
```

sirex (changes requested):

sirex (changes requested):
See my testing notes comments.

sirex (changes requested):

sirex (approved):

PR #445, opened 2023-05-11
373 fix tabular export for changes

adp-atea:


adp-atea:
Could not reproduce the problem. Added a test for now.

sirex (approved):

PR #447, opened 2023-05-12
Added test to replicate the bug

adp-atea:


notes/dimensions/base.sh
```diff
@@ -390,3 +397,48 @@ http GET "$SERVER/$DATASET/City?select(_id,id,name,population,type,koord)&format
...
+#|     expr = env.resolve(query)
+#|   File "spinta/backends/postgresql/commands/query.py", line 296, in select
+#|     selected = env.call('select', arg)
+#|   File "spinta/backends/postgresql/commands/query.py", line 320, in select
+#|     return env.call('select', prop.dtype)
+#|   File "spinta/backends/postgresql/commands/query.py", line 426, in select
+#|     column = table.c[dtype.prop.name]
+#|   File "sqlalchemy/sql/base.py", line 1192, in __getitem__
+#|     return self._index[key]
+#| KeyError: 'name'
```
sirex, line 444: I still get the same error.

sirex (changes requested):

notes/dimensions/base.sh
```diff
@@ -390,3 +324,9 @@ http GET "$SERVER/$DATASET/City?select(_id,id,name,population,type,koord)&format
 #       join.
 #
 #       https://github.com/atviriduomenys/spinta/issues/395
+
+http GET "$SERVER/$DATASET/City?select(id,name,country.name)&format(ascii)"
+#| id  name     country.name
+#| --  -------  ------------
+#| 24  Vilnius  Vilnius
+# FIXME: `country.name` should be Lithuania.
```
sirex, line 332: Something is not right, because `country.name` should be `Lithuania`.

sirex (changes requested):

notes/dimensions/base.sh
```diff
@@ -145,48 +153,63 @@ http POST "$SERVER/$DATASET/Place" $AUTH \
...
+#|                 "schema": "10"
+#|             },
+#|             "message": "Ability to indirectly modify base parameters is not implemented yet.",
+#|             "template": "{feature} is not implemented yet.",
+#|             "type": "property"
+#|         }
+#|     ]
 #| }
+# FIXME: This should not be an error, because we do not modify base, we just use
+#        _base.name as identifier, which exists in Place.
```
sirex, line 198: This is not exactly related with this task, bet could you please fix it, because I can't do further testing.

The error probably comes from https://github.com/atviriduomenys/spinta/pull/425

adp-atea: Before writing, added a check,, that compares all inherited variables if they were given.

sirex (approved):

PR #448, opened 2023-05-16
98 add inspect for json api params

adp-atea:


adp-atea:
JSON manifest was added, but still need ability to use params and need to adjust inspect, to map with source ".."

sirex (changes requested):

sirex (approved):

PR #478, opened 2023-06-13
display headers inline

adp-atea:


spinta/templates/data.html
```diff
@@ -26,26 +26,30 @@
 {% if empty %}
     <div>No data.</div>
 {% elif params.pk %}
```
sirex, line 28: Now `/$ID/:changes` are shown correctly, but `/$ID` now is shown incorrectly. `/$ID` should be left unchanged and should be shown as it was before, this fix is only needed for `/$ID/:changes`. So you need to check if `params.pk and not params.changes` or something like that.

This is strange, that this test did not broke any tests.

sirex (changes requested):

sirex (changes requested):
Please add tests.

sirex (approved):

PR #484, opened 2023-06-21
122 Auth clients self service

adp-atea:


notes/api/auth/clients.sh
```diff
@@ -0,0 +1,58 @@
...
+#| ERROR: Authorization server error: invalid_scope: 
+#| Traceback (most recent call last):
+#|   File "authlib/oauth2/rfc6749/authorization_server.py", line 185, in create_token_response
+#|     grant.validate_token_request()
+#|   File "authlib/oauth2/rfc6749/grants/client_credentials.py", line 72, in validate_token_request
+#|     self.validate_requested_scope(client)
+#|   File "authlib/oauth2/rfc6749/grants/base.py", line 92, in validate_requested_scope
+#|     raise InvalidScopeError(state=self.request.state)
+#| authlib.oauth2.rfc6749.errors.InvalidScopeError: invalid_scope: 
+#| INFO: "POST /auth/token HTTP/1.1" 400 Bad Request
```
sirex, line 58: For some reason, I no long can get access token.

sirex (changes requested):

notes/api/auth/clients.sh
adp-atea: I have tried to get access token with postman and it worked 
also I have tried with 
windows curl:
`curl -X POST -H "Content-Type: application/x-www-form-urlencoded" -d "grant_type=client_credentials&scope=spinta_getone spinta_insert" -u test:secret localhost:8000/auth/token`
it also worked, created auth client with 2 scopes.

then I tried Ubuntu WSL, with Httpie library:
`http -a test:secret -f :8000/auth/token grant_type=client_credentials scope="spinta_getone spinta_getall"`
It also worked, created auth client with 2 scopes.

Tried the same with older version of spinta it also worked there. Spinta seperates scopes with " " spacing, so that could have been the issue.

sirex: The error appears, when I request `spinta_auth_clients` scope, which is not recognized, but needed for client managment.

sirex: Created another issue, to show better error message: https://github.com/atviriduomenys/spinta/issues/537.

sirex (approved):

PR #489, opened 2023-06-29
Rewrote all dependencies according to notes given

### LINKED ISSUE #78: Move most deps to extras
Spinta should be very lightweight package and most of dependencies should go to extras.

adp-atea:


spinta/formats/ascii/commands.py
```diff
@@ -3,14 +3,15 @@
...
-from starlette.requests import Request
-from starlette.responses import StreamingResponse
-
 from spinta.formats.ascii.components import Ascii
 from spinta.components import Context, Action, UrlParams, Model
 from spinta import commands
+from spinta.utils.imports import use
 from spinta.utils.response import aiter, peek_and_stream
 
+Request = use('http', 'starlette.requests', 'Request').Request
```
sirex, line 12:
```suggestion
Request = use('http', 'starlette.requests', 'Request')
```

`Request` class is already given as a third argument, so there is no need to repeat it as an attribute.

Also, external classes should be accessed inside functions, where they are needed, not at modele scope. Spinta scans all modules for commands, so there is no point having optional dependencies if they are required if module is just imported.

All souch `use` classes should happint inside functions.

sirex (changes requested):

SauliusPetraitis:
Nepavyko, CCT bandymas naujesnis buvo

PR #492, opened 2023-07-14
397 fix denormalized properties

adp-atea:


sirex (approved):
At least looking at tests, looks good. Maybe would be nice to have tests for incorrect cases, like `a.b.c: string`, where `a` is declared as `integer` and similar.

Also, everytime I see `deepcopy` it raises a suspicion. I usually prefer explicit copying, to have better control and understanding on what was copied.

And another, thing, I'm not sure if it is worth introducing `partial` and `partial_array` types? Partials are tabular specific and should be resolved when parsing tabular manifests. Outside of tabular parser, we should not see any artifacts, from a specific manifest format, that does not apply to other formats.

These are just some cocerns, but if tests are green, the I think it is good enough.

PR #494, opened 2023-07-19
Support for raw

Faustasm:


spinta/manifests/sql/helpers.py
```diff
@@ -173,6 +173,7 @@ def _read_props(
     (postgresql.UUID, 'string'),
     (Geometry, 'geometry'),
     (oracle.ROWID, 'string'),
+    (oracle.RAW, 'string'),
```
sirex, line 176: `oracle.RAW` is a `'binary'` type, see:

- https://docs.sqlalchemy.org/en/14/dialects/oracle.html#sqlalchemy.dialects.oracle.RAW
- https://docs.oracle.com/cd/A58617_01/server.804/a58234/datatype.htm

sirex (changes requested):

spinta/manifests/sql/helpers.py
Faustasm: fixed

sirex (approved):

PR #497, opened 2023-07-26
454 check geometry boundaries

adp-atea:


spinta/backends/postgresql/types/geometry/helpers.py
```diff
@@ -27,10 +28,19 @@ def get_osm_link(value: WKBElement, srid: Optional[int]) -> Optional[str]:
...
-
     centroid = shape.centroid
     lat, lon = centroid.x, centroid.y
     params = urlencode({'mlat': lat, 'mlon': lon})
+    if srid and srid != WGS84:
+        if 'bbox' in transformer.to_json_dict():
+            south_lat = transformer.to_json_dict().get('bbox')['south_latitude']
+            north_lat = transformer.to_json_dict().get('bbox')['north_latitude']
+            east_lon = transformer.to_json_dict().get('bbox')['east_longitude']
+            west_lon = transformer.to_json_dict().get('bbox')['west_longitude']
```
sirex, line 39: Do not parse json multiple time, instead assign `transfer.to_json_dict().get('bbox')` to a variable and then, work with that variable.

spinta/backends/postgresql/types/geometry/helpers.py
```diff
@@ -27,10 +28,19 @@ def get_osm_link(value: WKBElement, srid: Optional[int]) -> Optional[str]:
...
+            if not south_lat < lat < north_lat:
+                raise LatitudeOutOfRange
```
sirex, line 41: Instead of just raising bare exception, pass property, that has the error. And the check should be done in `check` command. We don't want to check for errors, when reading data, we check errors, when writing data.

spinta/backends/postgresql/types/geometry/helpers.py
```diff
@@ -27,10 +28,19 @@ def get_osm_link(value: WKBElement, srid: Optional[int]) -> Optional[str]:
...
+            if not west_lon < lon < east_lon:
+                raise LongitudeOutOfRange
```
sirex, line 43: Same, hare, provide context for the error, so that client, could understand, where is the error.

sirex (changes requested):
Errors should be check before write, not on read.

sirex (approved):

PR #498, opened 2023-07-31
475 added acceptable content types

### LINKED ISSUE #475: Direct data upload
POST /datasets/gov/example/City

    Content-Type: text/csv

    csv data as http body

When a POST request is given, with a non `application/json` content type, then data should be imported into database in a similar way as with `spinta push`, respecting `source` and `model.ref` for id mapping. If `model.ref` is not specified, souch data uploads should not be allowed, because we cannot reliably map local and global identifiers.

Also, optionally POST with a `multipart/form-data` should be recognized, to handle data posted via [HTML forms](https://developer.mozilla.org/en-US/docs/Learn/Forms/Sending_and_retrieving_form_data#a_special_case_sending_files). When data are uploaded vie HTML forms, then form must contain `<input name="data" type="file">` field, which will be used to read data file from. Backend to read data with, should be recognized, by file extension of `file` field from [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) header.

All existing external dataset formats should be supported.

If there are errors, then errors should be displayed in format asked by the client via `Accept` header. Currently we only need to support JSON and HTML formats. JSON is already supported, so only HTML for errors needs to be added.

### Epic

- https://github.com/atviriduomenys/spinta/issues/247

### Related

- https://github.com/atviriduomenys/katalogas/issues/343

adp-atea:


spinta/commands/write.py
```diff
@@ -73,10 +77,15 @@ async def push(
         scope = scope.prop
 
     stop_on_error = not params.fault_tolerant
-    if is_streaming_request(request):
+    content_type = get_content_type_from_request(request)
+    if is_streaming_request(content_type):
         stream = _read_request_stream(
             context, request, scope, action, stop_on_error,
         )
+    elif is_not_streaming_request(request):
```
sirex, line 85:
```suggestion
    elif is_direct_upload(request):
```

spinta/commands/write.py
```diff
@@ -212,15 +221,26 @@ def _stream_group_key(data: DataItem):
...
-def is_streaming_request(request: Request):
+def get_content_type_from_request(request: Request):
     content_type = request.headers.get('content-type')
     if content_type:
         content_type = cgi.parse_header(content_type)[0]
+    return content_type
+
+
+def is_streaming_request(content_type):
+    content_type = cgi.parse_header(content_type)[0]
```
sirex, line 232: `parse_header` is already don in `get_content_type_from_request`?

spinta/commands/write.py
```diff
@@ -212,15 +221,26 @@ def _stream_group_key(data: DataItem):
...
     return content_type in STREAMING_CONTENT_TYPES
 
 
+def is_not_streaming_request(content_type):
+    content_type = cgi.parse_header(content_type)[0]
+    return content_type not in STREAMING_CONTENT_TYPES
```
sirex, line 238: Maybe use `detect_backend_from_content_type`? If backend could not be detected, then proceed to other request processing options?

spinta/datasets/backends/helpers.py
```diff
@@ -24,3 +27,29 @@ def handle_ref_key_assignment(keymap: KeyMap, value: Any, prop: Property) -> dic
             val = keymap.encode(prop.dtype.model.model_type(), value)
             val = {'_id': val}
     return val
+
+
+def detect_backend_from_content_type(context, content_type):
+    config = context.get('config')
+    backends = config.components['backends']
+    for backend in backends.values():
+        if issubclass(backend, ExternalBackend) and backend != BackendNotImplemented:
```
sirex, line 36:
```suggestion
        if issubclass(backend, ExternalBackend) and not issubclass(backend, BackendNotImplemented):
```

Always use `issubclass` or `isinstance`, when comparing classes, to take into account class inheritance.

spinta/datasets/backends/sql/components.py
```diff
@@ -12,6 +12,9 @@
 
 class Sql(ExternalBackend):
     type: str = 'sql'
+    accept_types = {
+        'text/xml',
```
sirex, line 16: XML for SQL backend? I think, currently `Sql` backend does not support any direct upload formats, because this backends can only read data from database, and you can't upload a database, you could upload a database dump, bud `Sql` backend does not support that.

spinta/datasets/components.py
```diff
@@ -94,6 +94,7 @@ def __init__(self):
 class ExternalBackend(Backend):
     engine: Engine = None
     schema: sa.MetaData = None
+    accept_types: set = None
```
sirex, line 97: Maybe set this to an empty set in `__init__`, in order to avoid dealing with `None`? It is always good idea, to avoid `None` if possible and in this case there is no need for `None`.

spinta/formats/helpers.py
```diff
@@ -154,3 +154,11 @@ def get_model_tabular_header(
...
         header = list(_get_model_header(model, names, select, reserved))
     return header
+
+
+def get_format_by_content_type(context, content_type):
+    config = context.get('config')
+    formats = config.components['exporters']
+    for fmt in formats.values():
+        if content_type in fmt.accept_types:
+            return fmt
```
sirex, line 164: Maybe use https://github.com/atviriduomenys/spinta/blob/3a57a4eedada898981b322c8dbe74816c5dce832/spinta/urlparams.py#L287 ?

tests/commands/test_direct_data_upload.py
```diff
@@ -0,0 +1,78 @@
...
+        response = app.post('/example/direct/data/City',
+                            headers=headers,
+                            files={'file': ('data.csv', file, 'text/csv')})
+
+    if os.path.exists(file_path):
+        os.remove(file_path)
+    assert response.status_code == 200
+
+
+def test_direct_data_upload_csv_2(
```
sirex, line 52:
```suggestion
def test_direct_data_upload_csv_via_body(
```

tests/commands/test_direct_data_upload.py
```diff
@@ -0,0 +1,78 @@
...
+
+from pathlib import Path
+from spinta.core.config import RawConfig
+from spinta.components import Mode
+from spinta.testing.pytest import rc
+from spinta.testing.client import create_test_client
+from spinta.testing.manifest import prepare_manifest
+
+
+def test_direct_data_upload_csv(
```
sirex, line 12:
```suggestion
def test_direct_data_upload_csv_via_form(
```

tests/commands/test_direct_data_upload.py
```diff
@@ -0,0 +1,78 @@
...
+    app.authmodel('example/direct/data/City', ['getall', 'insert', 'post'])
+    headers = {'Content-type': 'text/csv'}
+
+    with open(file_path, "rb") as file:
+        response = app.post('/example/direct/data/City',
+                            headers=headers,
+                            files={'file': ('data.csv', file, 'text/csv')})
+
+    if os.path.exists(file_path):
+        os.remove(file_path)
```
sirex, line 48: There is no need to remove file, because `tmp_path` fixture does that for you.

tests/commands/test_direct_data_upload.py
```diff
@@ -0,0 +1,78 @@
...
+
+    data = ('id, name, country\n'
+            '0, Vilnius, Lietuva\n'
+            '1, Kaunas, Lietuva\n'
+            '2, Ryga, Latvija\n'
+            '3, Berlynas, Vokietija'
+            )
+    headers = {'Content-type': 'text/csv'}
+    response = app.post('/example/direct/data/City', content=data, headers=headers)
+    print()
```
sirex, line 77: Should not be here.

sirex (changes requested):

spinta/datasets/backends/helpers.py
```diff
@@ -80,3 +85,31 @@ def extract_values_from_row(row: Any, model: Model, keys: list):
...
+def detect_backend_from_content_type(context, content_type):
+    config = context.get('config')
+    backends = config.components['backends']
+    for backend in backends.values():
+        if issubclass(backend, ExternalBackend) and not issubclass(backend, BackendNotImplemented):
+            if backend.accept_types and content_type in backend.accept_types:
+                return backend()
+    raise ValueError(
+        f"Can't find a matching external backend for the given content type {content_type!r}"
+    )
```
sirex, line 99: Raise `UserError`.

spinta/commands/write.py
```diff
@@ -49,6 +52,7 @@
 STREAMING_CONTENT_TYPES = [
     'application/x-jsonlines',
     'application/x-ndjson',
+    'application/json'
```
sirex, line 55: `application/json` should not be here, this is not a streaming content type.

SauliusPetraitis:
Uzduotis atsaukta

PR #501, opened 2023-08-01
Hot fix error with unique constraints

sirex:
Tests and probably better fix is needed.


adp-atea:
Changes in https://github.com/atviriduomenys/spinta/pull/503

adp-atea:
Why this task returned?

PR #505, opened 2023-08-16
477 HTTP API for inspect

adp-atea:
…dded inspect api


sirex (changes requested):
Can't get it to work, see my testing notes.

adp-atea:
There are multiple issues with the notes.

Main reason it did not work, was that in order to create response, spinta checks for backend. Since in the notes spinta was not ran with any config file, it did not have backend.

Another issue is that, currently `spinta inspect`, works by converting given `resource` to the `manifest` of it's type and then merging all of the manifests together. Tabular manifest types, unlike others, requires manifest to be given directly, so `inspect` with `resource` of type `csv` currently will not work with raw data, since it expects the given data to be manifest structure.

sirex (approved):

PR #509, opened 2023-09-13
508 Add swap user function

adp-atea:


notes/ufuncs/swap.sh
```diff
@@ -0,0 +1,50 @@
...
+#|   109 │   │   chosen_start = self._verify_start(start)
+#| 
+#| ╭──────────────────────────────────────── locals ────────────────────────────────────────╮
+#| │         text = 'swap(Vėlnius, "Vilnius").swap(Vilna, "Vilnius").swap(Kauna, "Kaunas")' │
+#| ╰────────────────────────────────────────────────────────────────────────────────────────╯
+#| UnexpectedCharacters: No terminal matches 'ė' in the current parser context, at line 1 col 7
+#| 
+#| swap(Vėlnius, "Vilnius").swap(Vilna, "Vilnius"
+#|       ^
+# FIXME: Values coming from `source` must be enclosed with quoates.
```
sirex, line 48: If value from source comes as string, it should be added with quotes.

sirex (changes requested):

sirex (approved):

PR #533, opened 2023-10-26
532 unauthorized keymap sync error

adp-atea:


notes/push/sync/private.sh
```diff
@@ -0,0 +1,104 @@
...
+#| Missing one of scopes:
+#| - spinta_push_sync_private_external_country_getall
+# TODO: there should be list of scopes:
+#       - spinta_push_sync_private_external_country_getall
+#       - spinta_push_sync_private_external_getall
+#       - spinta_push_sync_private_getall
+#       - spinta_push_sync_getall
+#       - spinta_push_getall
+#       - spinta_getall
+#       But only one scope is considered.
```
sirex, line 96: There should be no error here. I will create another issue for that.

sirex (approved):

notes/push/sync/private.sh
sirex: https://github.com/atviriduomenys/spinta/issues/534

PR #570, opened 2024-02-15
217,  256 parametrization

adp-atea:


spinta/dimensions/param/ufuncs.py
```diff
@@ -1,10 +1,140 @@
...
+            return True
+    return False
+
+
+@ufunc.resolver(ParamLoader, object)
+def select(env: ParamLoader, any_: object):
+    return True
+
+
+@ufunc.resolver(ParamLoader, Expr)
```
sirex, line 133: `Expr` is used, when you need to handle functions with dynamic arguments, for example `func(*args, **kwargs)`. But in this case we only have a single argument.

Instead `@ufunc.resolver(ParamLoader, Expr)` ir should be `@ufunc.resolver(paramLoader, types.GeneratorType)`.

And `loop` function implementation should look like this:

```python
@ufunc.resolver(ParamLoader, types.GeneratorType)
def loop(env: ParamLoader, rows: Iterable) -> Iterable:
    empty = False
    while not empty:
        empty = True
        for row in rows:
            empty = False
            yield row
```

spinta/dimensions/param/ufuncs.py
```diff
@@ -1,10 +1,140 @@
...
+    args, kwargs = expr.resolve(env)
+    for arg in args:
+        yield _Loop(expr, arg)
+
+
+@ufunc.resolver(ParamBuilder, Generator, Bind)
+def loop(env: ParamBuilder, generator: Generator, bind: Bind):
+    for item in generator:
+        data = env.call("getattr", item, bind)
+        yield data
```
sirex, line 80: There should be no such function like this. Here you are trying to resolve `loop(rows, attr)`. But instead it should be `getattr(loop(rows), attr)`.

`getattr` should expect generator type, iterate over it and call getattr again with a single row, for example:

```python
def getattr(env, rows, bind):
    for row in rows:
        env.call('getattr', row, bind)
```

spinta/dimensions/param/ufuncs.py
```diff
@@ -1,10 +1,140 @@
...
+    yield from resolved
+
+
+@ufunc.resolver(ParamBuilder, Generator, object, name="getattr")
+def getattr_(env: ParamBuilder, generator: Generator, obj: object):
+    for item in generator:
+        yield from env.call("getattr", item, obj)
+
+
+@ufunc.resolver(ParamBuilder, _Loop, Bind, name="getattr")
```
sirex, line 48: This should work not only with `_Loop`, but with any kind of `GeneratorType`.

spinta/dimensions/param/ufuncs.py
```diff
@@ -1,10 +1,140 @@
...
+        self.expr = expr
 
 
 @ufunc.resolver(Env, Bind)
 def param(env: Env, bind: Bind) -> Any:
     return env.params[bind.name]
+
+
+@ufunc.resolver(ParamBuilder, Expr)
+def select(env: ParamBuilder, expr: Expr) -> Any:
```
sirex, line 29: Can we change `select` to `read`? I just realized, that `select` is already used and does not have a first argument as model. So to avoid using same name with different semantic, I suggest, to change `select` here to `load`.

The full example would be:

```python
read('Page').loop().more
```

Which would be read into following AST:

```python
getattr(loop(read('Page')), bind('more'))
```

Also you could do this:

```python
read('Page').select(more).loop().more
```

For now we only need `read` ufunc.

sirex (changes requested):
See my inline comments. And I would like to see tests for the `iterparams` function.

spinta/dimensions/param/ufuncs.py
adp-atea: This is part of `ParamLoader`, it's main goal is validate and parse params at `link` time, to see if empty `read()` exists in the expression and if it does assume that `source` is a model and try to resolve it. `ParamBuilder` is used for looping through data.

spinta/dimensions/param/ufuncs.py
adp-atea: Loop is no longer part of the implementation

spinta/dimensions/param/ufuncs.py
adp-atea: loop is no longer part of the implementation

spinta/datasets/backends/dataframe/commands/read.py
```diff
@@ -362,14 +388,18 @@ def getall(
...
-        df = dask.bag.from_sequence([base]).map(
-            _get_data_json,
-            source=model.external.name,
-            model_props=props
-        ).flatten().to_dataframe(meta=meta)
-        yield from _dask_get_all(context, query, df, backend, model, builder)
+    model_formatted_bases = []
+    for formatted_base in bases:
+        for params in iterparams(context, model.manifest, model.params):
+            model_formatted_bases.append(formatted_base.format(params))
```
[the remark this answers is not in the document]
adp-atea: bases are now generators, but it only works for `xml` and `json`, since it has custom file read, while `csv` uses `dask` `read_csv` which only support `path` or `list`, so for `csv`, bases is converted to list.

spinta/datasets/backends/dataframe/commands/read.py
```diff
@@ -340,16 +342,40 @@ def _get_dask_dataframe_meta(model: Model):
     return dask_meta
 
 
+PARAM_EXTRACTION_REGEX = re.compile("{\\w*}")
+
+
+def requires_parametrization(context: Context, base: str):
+    matched_values = PARAM_EXTRACTION_REGEX.findall(base)
```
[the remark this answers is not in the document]
adp-atea: This has been reworked, regex is now only used at load time to get all required params for the resource in order to determine which params does the resource depend on.

sirex (approved):

PR #578, opened 2024-03-06
549 language support in rdf

adp-atea:


tests/formats/test_ascii.py
```diff
@@ -434,6 +434,146 @@ async def test_ascii_check_last_page(
...
+            'C': 'UK'
+        }
+    })
+
+    res = app.get("/example/ascii/text/lang/Country/:format/ascii?lang(*)&select(id,name)", headers=Headers(headers={
+        'accept-language': 'lt'
+    })).text
+    assert res == (
+        '--  ------  ---------  -------\n'
+        'id  name.C  name.en    name.lt\n'
```
sirex, line 550: Would it be difficult to change column names into:

```
id name@C name@en name@lt
```

To match `property` values in manifest table? Not only for ASCII, but for all tabular formats?

sirex (approved):
Minor issues regarding @C and tabular header naming.

tests/formats/test_ascii.py
adp-atea: Currently headers are tightly coupled with mapping logic, in most cases data is given as `dict` format (with no information about types). This `dict` is then flattened and mapped with header names. Changing header names to `name@en` would be easy, but then we get issues with mapping, since flattening function only accepts one type of separator and by default its `.`. In order for it to be able to flatten based on property type, we would need to change how `flatten` works.

tests/dtypes/test_text.py
```diff
@@ -466,9 +466,7 @@ def test_text_unknown_language(
     assert select_by_prop.status_code == 200
     assert len(select_by_prop.json()['_data']) == 1
     assert select_by_prop.json()['_data'] == [{
-        'name': {
-            'C': 'LT',
-        }
+        'name': 'LT'
```
sirex, line 469: If explicit form is requested `name@C` we must always return data in explicit form to, here we should get:

```python
'name': {
    '': 'LT',
}
```

tests/formats/test_ascii.py
```diff
@@ -434,6 +434,146 @@ async def test_ascii_check_last_page(
...
+    )
+    app = create_test_client(context)
+    app.authmodel('example/ascii', ['insert', 'getall', 'search'])
+
+    pushdata(app, f'/example/ascii/text/Country', {
+        'id': 0,
+        'name': {
+            'lt': 'Lietuva',
+            'en': 'Lithuania',
+            'C': 'LT'
```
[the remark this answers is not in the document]
sirex: Why this is marked as resolved, if `'C': 'LT'` is still returned?

tests/formats/test_ascii.py
sirex: Moved to a separate issue:

- https://github.com/atviriduomenys/spinta/issues/581

PR #765, opened 2024-09-05
752 xsd to dsa remove duplicate models1 into 0.1

karina-klinkeviciute:


spinta/manifests/xsd/helpers.py
```diff
@@ -112,6 +112,13 @@ def __init__(self, xsd: 'XSDReader', node: _Element = None):
         self.parent_model: XSDModel | None = None
         self.is_root_model: bool | None = None
 
+    def __eq__(self, other: XSDModel) -> bool:
+        if self.external["name"] != other.external["name"]:
```
ievabraskute, line 116: Maybe it's worth adding a type check?
`if not isinstance(other, XSDModel):
        return False
`

spinta/manifests/xsd/helpers.py
```diff
@@ -855,10 +862,31 @@ def _add_refs_for_backrefs(self):
...
         for model in self.models.values():
             model.properties = dict(sorted(model.properties.items()))
 
+    def _remove_duplicate_models(self):
+        """removes models that are exactly the same"""
+
+        model_pairs = {}
+
+        for model_name, model in self.models.items():
+            for another_model_name, another_model in self.models.items():
```
ievabraskute, line 875: Is it only possible for 2 models to be identical and not 3 or more?

tests/manifests/xsd/test_xsd.py
```diff
@@ -189,18 +186,26 @@ def test_xsd(rc: RawConfig, tmp_path: Path):
...
+    |   |   |   |   | iki             | date required    |     | IKI/text()                        |         |       |        |     |       |
+    |   |   |   |   | nuo             | date required    |     | NUO/text()                        |         |       |        |     |       |
+    |   |   |   |   | tipas           | string required  |     | TIPAS/text()                      |         |       |        |     |       |
+    |   |   |   |   | tipo_santrumpa  | string required  |     | TIPO_SANTRUMPA/text()             |         |       |        |     |       |
+    |   |   |   |   | vardas_k        | string required  |     | VARDAS_K/text()                   |         |       |        |     |       |
+    |   |   |   |   | vardas_k_lot    | string required  |     | VARDAS_K_LOT/text()               |         |       |        |     |       |
+    |   |   |   |   | vardas_v        | string required  |     | VARDAS_V/text()                   |         |       |        |     |       |
+    |   |   |   |   | vardas_v_lot    | string required  |     | VARDAS_V_LOT/text()               |         |       |        |     |       |
```
sirex, line 202: In tests, we don't need to repeat same properties, multiple times, we need at least won example of each case. Test data should be small enough to just cover the cases, which is tested.

Now there is too many moving parts and I'm not sure which are important for this particular test.

tests/manifests/xsd/test_xsd.py
```diff
@@ -1089,3 +1081,79 @@ def test_xsd_enumeration(rc: RawConfig, tmp_path: Path):
...
+                    <xs:element minOccurs="0" maxOccurs="1" name="extract" type="Extract" />
+                </xs:choice>
+            </xs:sequence>
+        </xs:complexType>
+        <xs:complexType name="getDocumentsByWagonSearchParams">
+            <xs:sequence>
+                <xs:element minOccurs="0" maxOccurs="1" name="searchType" type="xs:string" />
+                <xs:element minOccurs="0" maxOccurs="1" name="code" type="xs:string" />
+                <xs:element minOccurs="0" maxOccurs="1" name="RegDateFrom" type="xs:string" />
+                <xs:element minOccurs="0" maxOccurs="1" name="RegDateTo" type="xs:string" />
```
sirex, line 1104: In this test case, we need at least one element, because we are testing duplicate model removal, so there elements are not needed in this test.

sirex (commented):
Looks good, just the implementation, looks a bit inefficient, written in C brute force style, python has better tools for duplication removal.

Also, test cases are too verbose involving many unnecessary complexities. Code would be much better maintainable, if test cases would use [Minimal reproducible examples](https://en.wikipedia.org/wiki/Minimal_reproducible_example), instead of copy-past blobs.

spinta/manifests/xsd/helpers.py
karina-klinkeviciute: it is possible, yes. And this should catch them all. Or at least I think so. 

`model` is the model that it's being compared against, and `another_model` is the one that's being compared. There might be many cases of `another_model` being equal to `model` and it's added to the `model_pairs" every time. I used `another_model_name` as a key for this reason. So if we have model `Person`, then an identical model `Person1` and `Person2`, the model_pairs will look like this: `{"Person1": "Person", "Person2": "Person"}` and then the references will be reassigned for each of those models.

karina-klinkeviciute: > Is it only possible for 2 models to be identical and not 3 or more?

I can be wrong though, if you think that it won't remove more than one duplicate, could you suggest what I should do differently?

spinta/manifests/xsd/helpers.py
```diff
@@ -112,6 +112,13 @@ def __init__(self, xsd: 'XSDReader', node: _Element = None):
         self.parent_model: XSDModel | None = None
         self.is_root_model: bool | None = None
```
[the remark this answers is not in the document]
karina-klinkeviciute: PyCharm does understand this. And some of the variables are being initialized here with values passed to the __init__ so I think it makes sense to initialize the other ones with None values at the same place, too. 

Or is there any reason to declare them at the class level?

spinta/manifests/xsd/helpers.py
```diff
@@ -855,10 +862,31 @@ def _add_refs_for_backrefs(self):
...
+    def _remove_duplicate_models(self):
+        """removes models that are exactly the same"""
+
+        model_pairs = {}
+
+        for model_name, model in self.models.items():
+            for another_model_name, another_model in self.models.items():
+                if (model is not another_model) and (model == another_model):
+                    if another_model_name not in model_pairs.values():
+                        model_pairs[another_model_name] = model_name
```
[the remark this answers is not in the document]
karina-klinkeviciute: I will explore those options, thanks

tests/manifests/xsd/test_xsd.py
karina-klinkeviciute: I will minimize tests, thanks

spinta/manifests/xsd/helpers.py
ievabraskute: You're right, my mistake!

spinta/manifests/xsd/helpers.py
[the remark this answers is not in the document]
karina-klinkeviciute: > I thin running `collections.Counter` on `self.models.values().external['name']` should be much more efficient and probably simpler solution?
There are cases, when the model is split by `choice`, for example `FizinisAsmuo` and `JuridinisAsmuo`, where external['name'] can be the same, but properties might be different, in this case it's not an exact duplicate and we can't remove it. We still need to check if properties are equal.

tests/manifests/xsd/test_xsd.py
```diff
@@ -1089,3 +1081,79 @@ def test_xsd_enumeration(rc: RawConfig, tmp_path: Path):
...
+    |   |   |   |   | extract.last_update_time         | datetime required |                  | extract/lastUpdateTime/text()                 |         |       |        |     |       |
+    |   |   |   |   | search_parameters                | ref               | SearchParameters | searchParameters                              |         |       |        |     |       |
+    |   |   |   |   | search_parameters.code           | string            |                  | searchParameters/code/text()                  |         |       |        |     |       |
+    |   |   |   |   | search_parameters.reg_date_from  | string            |                  | searchParameters/RegDateFrom/text()           |         |       |        |     |       |
+    |   |   |   |   | search_parameters.reg_date_to    | string            |                  | searchParameters/RegDateTo/text()             |         |       |        |     |       |
+    |   |   |   |   | search_parameters.search_type    | string            |                  | searchParameters/searchType/text()            |         |       |        |     |       |
```
sirex, line 1146: Ar teisingai suprantu, kad dublikatai nepasinaikino?

karina-klinkeviciute: Pasinaikino tik tie dublikatai, kur viskas (source ir visos savybės) sutapo, o pavadinimai skyrėsi.

PR #908, opened 2024-10-21
860 implement post_process_refs

### LINKED ISSUE #860: XSD -> DSA: post_process_refs
When building a model tree, the top level `element` and `complexType` nodes are being turned into models, but the references on properties on other models are not being created. this is being done to avoid too much logic happening at once. 

Because of this, we need to revisit each property and create those links as needed. If property has a link to a model, which hasn't been properly processed yet, it will have `xsd_ref_to` if it is connected through `ref` and `xsd_type_to` if it's connected through `type`. Models that have to be checked, have `is_root_model` set to True and they have `source`. 

To search models quicker, models that can be referenced, will be added to `XSDReader`     `top_level_element_models` or  `top_level_complex_type_models` dictionaries depending on if it's created from `element` or from `complexType`. 

- [ ] tests
- [ ] documentation
- [ ] entry in CHANGES.rst

ievabraskute:
closes #860

### DIFF
--- a/spinta/manifests/xsd2/helpers.py
+++ b/spinta/manifests/xsd2/helpers.py
@@ -343,10 +343,13 @@ class XSDReader:
         """
         for model in self.models:
             for prop in model.properties.values():
+                
                 if prop.xsd_ref_to:
-                    target_model: XSDModel = self.top_level_element_models.get(prop.xsd_ref_to) or self.top_level_complex_type_models.get(prop.xsd_ref_to)
-                    if not target_model:
+                    try:
+                        target_model: XSDModel = self.top_level_element_models[prop.xsd_ref_to]
+                    except KeyError:
                         raise KeyError("Reference to a non-existing model")
+                    
                 elif prop.xsd_type_to:
                     try:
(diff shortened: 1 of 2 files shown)


spinta/manifests/xsd2/helpers.py
```diff
@@ -338,7 +338,22 @@ def _post_process_resource_model(self):
...
 
     def _post_process_refs(self):
-        pass
+        """
+        Links properties in all models to their target models based on xsd_ref_to and xsd_type_to.
+        """
+        for model in self.models:
+            for prop in model.properties.values():
+                if prop.xsd_ref_to:
+                    target_model: XSDModel = self.top_level_element_models.get(prop.xsd_ref_to) or self.top_level_complex_type_models.get(prop.xsd_ref_to)
```
karina-klinkeviciute, line 347: In this case it should look only in `top_level_element_models`. If it doesn't find the model there, it means that there's some bug in adding models to those lists.

karina-klinkeviciute (approved):

PR #1023, opened 2024-12-08
987 error message for ref without model

karina-klinkeviciute:


spinta/exceptions.py
```diff
@@ -998,3 +998,7 @@ class BackendUnavailable(BaseError):
 
 class InvalidClientFileFormat(UserError):
     template = "File {client_file} data must be a dictionary, not a {client_file_type}."
+
+
+class MissingRefModel(UserError):
+    template = "Property of type `ref` or `backref` should have a model in the `ref` column"
```
sirex, line 1004: Add explicitly names of property, that has the error.

spinta/types/ref/link.py
```diff
@@ -14,6 +14,8 @@ def link(context: Context, dtype: Ref) -> None:
 
     # XXX: https://github.com/atviriduomenys/spinta/issues/44
     rmodel: str = dtype.model
+    if rmodel is None:
+        raise MissingRefModel
```
sirex, line 18:
```suggestion
        raise MissingRefModel(dtype=dtype)
```

spinta/exceptions.py
```diff
@@ -1001,4 +1001,4 @@ class InvalidClientFileFormat(UserError):
 
 
 class MissingRefModel(UserError):
-    template = "Property of type `ref` or `backref` should have a model in the `ref` column"
+    template = 'Property "{property_name}" of type "{property_type}" in the model "{model_name}" should have a model name in the `ref` column, to which it refers.'
```
sirex, line 1004:
```suggestion
    template = 'Property "{dtype.prop.place}" of type "{dtype.name}" in the model "{dtype.prop.model.name}" should have a model name in the `ref` column, to which it refers.'
```

karina-klinkeviciute: It actually was a bit different, through the use of `context`, but I changed it.

sirex (approved):

PR #1028, opened 2024-12-12
582 text type as ref (#1004)

ievabraskute:
582 text type as ref


ievabraskute:
decided not to merge to 0.1, only master

PR #1118, opened 2025-01-30
[#1019] Additional checks when property.type is not explicitly defined

Struchid:
This Pull Request fixes a couple of bugs that were raised for the following DSA:

```
    d | r | b | m | property     | type    | ref
    example                      |         |
      |   |   | Area             |         |
      |   |   |   | name         | string  |
      |   |   |   | population   | integer |
      |   | Area                 |         |
      |   |   | Country          |         |
      |   |   |   | name         |         |
      |   |   |   | code         |         |
      |   |   |   | population   | integer |
```

1. `_get_type_repr()` function if breaking the flow, due to `property.type` being accessed with in operator, and the `property.type` value being `None`, causing: `TypeError: argument of type 'NoneType' is not iterable`.

**Fixed by returning a default `inherit` value if the passed argument `dtype` has a value of `None`.**

2. The DSA itself, should not pass, because property `code` has no defined type AND the `Country` base model `Area` does not have a code `type` which could be inherited by `country.code`.

**Fixed by implementing `check[Inherit]` function and adding the required logic there.**

#### Notes:
* Replaced `test_prop_type_not_given` with `test_property_type_undefined_no_base_model` since they are both checking the same thing and the example in the latter one is more detailed.
* Also added `type` attribute to `PropertyGiven` following a comment from Mantas: 
image

#### Issue: https://github.com/atviriduomenys/spinta/issues/1019

### DIFF
--- a/spinta/components.py
+++ b/spinta/components.py
@@ -751,11 +751,12 @@ class Model(MetaData):
 
 
 class PropertyGiven:
-    access: str = None
-    enum: str = None
-    unit: str = None
-    name: str = None
+    access: Optional[str] = None
+    enum: Optional[str] = None
+    unit: Optional[str] = None
+    name: Optional[str] = None
     explicit: bool = True
+    type: Optional[str] = None
     prepare: List[PrepareGiven] = []
 
 
--- a/spinta/components.py
+++ b/spinta/components.py
@@ -751,12 +751,12 @@ class Model(MetaData):
 
 
 class PropertyGiven:
-    access: Optional[str] = None
-    enum: Optional[str] = None
-    unit: Optional[str] = None
-    name: Optional[str] = None
+    access: str | None = None
+    enum: str | None = None
+    unit: str | None = None
+    name: str | None = None
     explicit: bool = True
-    type: Optional[str] = None
+    type: str | None = None
     prepare: List[PrepareGiven] = []
 
 
--- a/spinta/exceptions.py
+++ b/spinta/exceptions.py
@@ -106,7 +106,7 @@ class BaseError(Exception):
             this = args[0]
         else:
             this = None
-            log.error("Only one positional argument is alowed, but %d was given.", len(args), stack_info=True)
+            log.error("Only one positional argument is allowed, but %d was given.", len(args), stack_info=True)
 
         self.type = this.type if this and hasattr(this, 'type') else 'system'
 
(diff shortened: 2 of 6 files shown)


spinta/components.py
```diff
@@ -751,11 +751,12 @@ def get_from_flatprops(self, prop: str) -> Property:
 
 
 class PropertyGiven:
-    access: str = None
-    enum: str = None
-    unit: str = None
-    name: str = None
+    access: Optional[str] = None
```
karina-klinkeviciute, line 754: siūlyčiau rašyti ne `Optional[str]` bet `str | None`, t.y. naudoti python in-built tipų aprašymo priemones, nes visos `spinta` palaikomos versijos tai palaiko.

Struchid: Spinta palaiko Python 3.9 ir aukštesnes versijas:
https://github.com/atviriduomenys/spinta/blob/3d1bfee2d00bb7d0fa1d02769bde75e87f7e137f/pyproject.toml#L36-L37

Built-in python tipai pristatyti tik Python 3.10 versijoje:
> Changed in version 3.10: Optional can now be written as X | None.

* https://docs.python.org/3/whatsnew/3.10.html#pep-604-new-type-union-operator.
* https://docs.python.org/3/library/typing.html#typing.Optional

karina-klinkeviciute: tada gal geriau naudoti `from __future__ import annotations`. Bet aš jau pradedu maišytis `spinta` palaikomose versijose - man atrodo, kad nors čia ir parašyta `3.9`, bet ji jau nebepalaikoma paties `Python` (tiksliau, deprecated ir greitai bus visai nebvepalaikoma), tai nuspręsta, kad ir `spinta` jos turėtų nebepalaikyti. 

Dabar nerandu, kur visa tai nurodyta, bet bent jau čia rašoma, kad reikalinga Python 3.10 ar aukštesnė versija: https://spinta.readthedocs.io/en/latest/contrib/setup.html

karina-klinkeviciute: aha, matau ir github actions nurodoma, kad žemiausia versija 3.9 - tai gal vis dėlto ji dar palaikoma. 

Bet vis tiek, man atrodo, kad pridėjus `from__future__ import annotations` turėtų ir 3.9 versijoje leisti `str | None`, tai mano nuomone, čia būtų patogiausias variantas.

Struchid: [3a7015d](https://github.com/atviriduomenys/spinta/pull/1118/commits/3a7015df51c4edd21ccffb5c237ffbbbc352107b)

Issue #576, opened 2024-03-04
Do not allow NULL for model.ref


sirex:
Raise error if column values of `model.ref` contains NULL values.

If `model.ref` column values contains NULL values, we can't use it as unique identifier.

PR #1156, opened 2025-02-12
[#1015] Throw an error for array type primary keys

Struchid:
`Array` type object can not be a PK for a model. With this pull request the logic is adjusted to throw a nice error message to the user specifying just that.

#### Notes:
- Added two tests, one, that proves that the error is thrown on the specific situation where PK has an array type object and another, that proves that backref works correctly, when the primary key is of allowed type. Not sure how much the second test is required, maybe someone with more experience on this project can tell me.

### DIFF
--- a/spinta/datasets/commands/load.py
+++ b/spinta/datasets/commands/load.py
@@ -130,7 +130,7 @@ def load(context: Context, entity: Entity, data: dict, manifest: Manifest):
     else:
         entity.unknown_primary_key = True
         entity.pkeys = sorted(
-            take(entity.model.flatprops).values(),
+            [pk for pk in take(entity.model.flatprops).values() if not pk.list],
             key=lambda p: p.place,
         )
 
--- a/spinta/datasets/commands/load.py
+++ b/spinta/datasets/commands/load.py
@@ -125,12 +125,12 @@ def load(context: Context, entity: Entity, data: dict, manifest: Manifest):
     # XXX: https://gitlab.com/atviriduomenys/spinta/-/issues/44
     pkeys: List[str] = entity.pkeys or []
     if pkeys:
-        _check_unknown_keys(entity.model, pkeys, entity.model.properties)
-        entity.pkeys = [entity.model.properties[k] for k in pkeys]
+        _check_unknown_keys(entity.model, pkeys, entity.model.flatprops)
+        entity.pkeys = [entity.model.flatprops[k] for k in pkeys]
     else:
         entity.unknown_primary_key = True
         entity.pkeys = sorted(
-            take(entity.model.properties).values(),
+            take(entity.model.flatprops).values(),
             key=lambda p: p.place,
         )
 
--- a/spinta/datasets/components.py
+++ b/spinta/datasets/components.py
@@ -197,7 +197,7 @@ class Entity(External):
     resource: Optional[Resource]    # resource
     model: Model                    # model
     pkeys: List[Property]           # model.ref
-    # This is set to True if primary key is not given. Anf if primary key is not
+    # This is set to True if primary key is not given. And if primary key is not
     # given, then `pkeys` will be set to all properties.
     unknown_primary_key: bool = False
     name: str                       # model.source
--- a/spinta/exceptions.py
+++ b/spinta/exceptions.py
@@ -766,7 +766,7 @@ class NoReferencesFound(UserError):
 
 
 class PrimaryKeyArrayTypeError(UserError):
-    template = "Array type object can not be used as a model primary key."
+    template = "Array type property can not be used as a model primary key."
 
 
 class MultipleBackRefReferencesFound(UserError):
--- a/spinta/exceptions.py
+++ b/spinta/exceptions.py
@@ -765,6 +765,10 @@ class NoReferencesFound(UserError):
     template = "Property {prop_name!r} of type 'Ref' was not found."
 
 
+class PrimaryKeyArrayTypeError(UserError):
+    template = "Array type object can not be used as a model primary key."
+
+
 class MultipleBackRefReferencesFound(UserError):
     template = "Model {model!r} contains multiple references to backref, please specify which one to use."
 
--- a/spinta/types/datatype.py
+++ b/spinta/types/datatype.py
@@ -477,30 +477,21 @@ def load(context: Context, dtype: Array, data: dict, manifest: Manifest) -> Data
 
 @load.register(Context, ArrayBackRef, dict, Manifest)
 def load(context: Context, dtype: ArrayBackRef, data: dict, manifest: Manifest) -> DataType:
-    if dtype.items:
-        assert isinstance(dtype.items, dict), type(dtype.items)
-        prop: Property = dtype.prop.__class__()
-        prop.name = dtype.prop.name
-        prop.place = dtype.prop.place
-        prop.parent = dtype.prop
-        prop.model = dtype.prop.model
-        commands.load(context, prop, dtype.items, manifest)
-        dtype.items = prop
-    else:
-        dtype.items = None
-    return dtype
+    return _load_array(context, dtype, manifest)
 
 
 def _load_array(context: Context, dtype: Array | ArrayBackRef, manifest: Manifest) -> DataType:
     if dtype.items:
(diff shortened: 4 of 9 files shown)


Struchid:
The test is failing for no apparent reason, seems to be running just OK locally:
image

karina-klinkeviciute:
> The test is failing for no apparent reason, seems to be running just OK locally

I checked out the branch locally and the tests are failing just like on the server. It happens while running all tests, but also when I run that test individually

Struchid:
> > The test is failing for no apparent reason, seems to be running just OK locally
> 
> I checked out the branch locally and the tests are failing just like on the server. It happens while running all tests, but also when I run that test individually

I guess that just proves that "The morning is wiser than the evening", will be working on the fix today:-)

Struchid:
The exception that other tests are raising is the exception that has been implemented in this Pull Request. The problem arises because of this reason:

https://github.com/atviriduomenys/spinta/blob/a0c084b7f03ebc8159fb806a412622a7ad9ca2ba/spinta/datasets/components.py#L199-L202

The `context` fixture loads a dataset without an explicitly set primary-key, which makes it so the Primary key is a composite of all the properties for the model. Which includes the array type properties as well (e.g. `notes[]`, `operating_licenses[]`):
image.

This does not seem to be a problem with the changes, if we are not allowing array type properties as PK's.

sirex:
> The `context` fixture loads a dataset without an explicitly set primary-key, which makes it so the Primary key is a composite of all the properties for the model. Which includes the array type properties as well (e.g. `notes[]`, `operating_licenses[]`)
> 
> This does not seem to be a problem with the changes, if we are not allowing array type properties as PK's.

Yes, this behavior when primary keys is not specified an all properties are used as primary key, was problematic in the past.

I think, instead of using all properties for primary key, we should do a more intelligent primary key detection if it is not specified.

Currently this issue is being solved here:

- https://github.com/atviriduomenys/spinta/issues/1159

For now, I suggest to make sure, that arrays are excluded from primary key, if it is not specified.

Later we might add more logic what can and what can't go to primary key, for example:

- https://github.com/atviriduomenys/spinta/issues/576

Struchid:
Additional questions after adding [prop.list assignment](https://github.com/atviriduomenys/spinta/pull/1156/files#diff-f32b8ded1dca8498ccae6304aa4319dc546a677bfde8c345661b6a66221056c0R488). Now, we have a number of tests breaking, because the added assignment causes the creation of an intermediary table:

```python
FAILED tests/dtypes/test_backref.py::test_backref_many_to_one_level_4[csv] - KeyError: 'example/dtypes/backref/mto/level4/Language/:list/country'
FAILED tests/dtypes/test_backref.py::test_backref_many_to_one_level_4[internal_sql] - KeyError: 'example/dtypes/backref/mto/level4/Language/:list/country'
FAILED tests/dtypes/test_backref.py::test_backref_many_to_one_level_3[csv] - KeyError: 'example/dtypes/backref/mto/level3/Language/:list/country'
FAILED tests/dtypes/test_backref.py::test_backref_many_to_one_level_3[internal_sql] - KeyError: 'example/dtypes/backref/mto/level3/Language/:list/country'
FAILED tests/dtypes/test_backref.py::test_backref_many_to_many_level_4[csv] - KeyError: 'example/dtypes/backref/mtm/level4/Language/:list/country'
FAILED tests/dtypes/test_backref.py::test_backref_many_to_many_level_4[internal_sql] - KeyError: 'example/dtypes/backref/mtm/level4/Language/:list/country'
FAILED tests/dtypes/test_backref.py::test_backref_many_to_many_level_3[csv] - KeyError: 'example/dtypes/backref/mtm/level3/Language/:list/country'
FAILED tests/dtypes/test_backref.py::test_backref_many_to_many_level_3[internal_sql] - KeyError: 'example/dtypes/backref/mtm/level3/Language/:list/country'
FAILED tests/dtypes/test_backref.py::test_backref_x_to_many_expand[csv] - KeyError: 'example/dtypes/backref/xtm/expand/Language/:list/country'
FAILED tests/dtypes/test_backref.py::test_backref_x_to_many_expand[internal_sql] - KeyError: 'example/dtypes/backref/xtm/expand/Language/:list/country'
FAILED tests/dtypes/test_backref.py::test_backref_multiple_same[csv] - KeyError: 'example/dtypes/backref/multiple/same/Language/:list/country_primary'
FAILED tests/dtypes/test_backref.py::test_backref_multiple_same[internal_sql] - KeyError: 'example/dtypes/backref/multiple/same/Language/:list/country_primary'
FAILED tests/dtypes/test_backref.py::test_backref_multiple_all_types[csv] - KeyError: 'example/dtypes/backref/multiple/all/Language/:list/country_1'
FAILED tests/dtypes/test_backref.py::test_backref_multiple_all_types[internal_sql] - KeyError: 'example/dtypes/backref/multiple/all/Language/:list/country_1'
```

E.g. of the error:
```python
...
spinta/backends/postgresql/ufuncs/query/ufuncs.py:304: in select
    selected = _select_backref(env, dtype.items.dtype, is_array=True)
spinta/backends/postgresql/ufuncs/query/ufuncs.py:331: in _select_backref
    selector = env.generate_backref_select(dtype.prop, dtype.refprop, required_columns, is_array)
spinta/backends/postgresql/ufuncs/query/components.py:222: in generate_backref_select
    left_keys = self.backend.get_column(left_table, left)
spinta/backends/postgresql/components.py:121: in get_column
    table = self.get_table(prop.list, TableType.LIST)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <spinta.backends.postgresql.components.PostgreSQL(name='default') at 0x11ec9f4f0>
node = <spinta.components.Property(name='country', type='array_backref', model='example/dtypes/backref/mto/level4/Language')>
ttype = <TableType.LIST: '/:list'>

    def get_table(
        self,
        node: Union[Model, Property],
        ttype: TableType = TableType.MAIN,
        *,
        fail: bool = True,
    ):
        name = get_table_name(node, ttype)
        if fail:
>           return self.tables[name]
E           KeyError: 'example/dtypes/backref/mto/level4/Language/:list/country'

spinta/backends/postgresql/components.py:115: KeyError
```

So what should we expect in this case? As far as I understand, this additional intermediary table should be available for m2m test cases and not present in the others?

1. So for m2m cases, `self.tables` of the PostgreSQL class:
https://github.com/atviriduomenys/spinta/blob/24e8d030069c558717b2198546a10f185486333b/spinta/backends/postgresql/components.py#L21
SHOULD have this intermediary table in place, and `KeyError` shouldn't be thrown. 

2. And in the case of o2m, intermediary tables should not be present, and the `KeyError` should be avoided by not having a `name` like `...Language/:list/country` passed to the function `get_table` in the first place?

TLDR; `Language/:list/country` tables should only be present for m2m relationships and NOT PRESENT for o2m?

sirex:
> So what should we expect in this case? As far as I understand, this additional intermediary table should be available for m2m test cases and not present in the others?

Intermediary `:list` table should be created only for m2m relations.

> 1. So for m2m cases, `self.tables` of the PostgreSQL class:
> https://github.com/atviriduomenys/spinta/blob/24e8d030069c558717b2198546a10f185486333b/spinta/backends/postgresql/components.py#L21
> SHOULD have this intermediary table in place, and `KeyError` shouldn't be thrown. 

In m2m cases, there should be an intermediate table and error should not be

> 2. And in the case of o2m, intermediary tables should not be present, and the `KeyError` should be avoided by not having a `name` like `...Language/:list/country` passed to the function `get_table` in the first place?
> 
> 3. What should be the result of `get_table_name` function for the following node? 
> https://github.com/atviriduomenys/spinta/blob/24e8d030069c558717b2198546a10f185486333b/spinta/backends/helpers.py#L331
> ```python
> <spinta.components.Property(name='country', type='array_backref', model='example/dtypes/backref/mto/level4/Language')>
> ```
> 
> Can it just be: 
> ```python
> 'example/dtypes/backref/mto/level4/Language'
> ```
> Or is this the one applicable?
> ```python
> 'example/dtypes/backref/mto/level4/Language/:list'
> ```

PR #1200, opened 2025-03-06
[#1019] Undefined property type checks 0.1

Struchid:


spinta/types/inherit/check.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@commands.check.register(Context, Inherit)
+def check(context: Context, dtype: Inherit) -> None:
+    property = dtype.prop
+    if property.is_reserved():
+        return None
+
+    property_model = property.model
+    property_name = property.name
```
varzgalys, line 14: necessary?

varzgalys (approved):

spinta/types/inherit/check.py
Struchid: re-used more than once.

varzgalys: But it doesn't save any computation time or line length. Maybe rename to just `model` and `name`?

varzgalys: optional

PR #1232, opened 2025-03-24
1229: adding get_json_content_from_yaml function, improving prepare method

MindaugasStanionis:


spinta/cli/get.py
```diff
@@ -0,0 +1,50 @@
...
+        commands.getall(context, model, store.manifest.backend)
+    )
+    echo(response_dict)
+
+
+def getone(
+    ctx: TyperContext,
+    manifests: Optional[List[str]] = Argument(help=("Manifest files to load")),
+    dsn: str = Argument(help=("Data Source Name path")),
+    dataset: str = Argument(help=("Dataset name")),
```
varzgalys, line 38: I think it should called - `model` or `model_path`
```suggestion
    model: str = Argument(help=("Model path")),
```

varzgalys (approved):
Looks good

Struchid (approved):
I suggest preparing the minimal reproducible example in the task/PR description, since the question why it's not there will most definitely be asked during the review (Most likely can just be taken from the tests, although I am not too familiar with the logic you're introducing here, so @varzgalys might know more). Also, don't forget to prepare the test file for demonstration during the review meeting on Wednesday.

martynasracct (approved):

spinta/backends/memory/commands/load.py
```diff
@@ -7,4 +7,6 @@
 
 @commands.load.register(Context, Memory, dict)
 def load(context: Context, backend: Memory, config: Dict[str, Any]):
+    if 'dsn' in config:
+        backend.dsn = config['dsn']
```
varzgalys, line 11: still should be here (after changing to backend type)?

MindaugasStanionis: Yes, it checks if the user added the backend (dsn), if user only added backend_type 'memory' but no backend, without this it would break

PR #1237, opened 2025-03-26
ft/1196-better-error-messages

varzgalys:
## Pull Request Summary  
This PR enhances error messages by adding relevant context and removes unused code to improve maintainability.  

## Error Message Improvements  
- Add property to **SRIDNotSetForGeometry** error message.  
- Add model name to **UnableToMapIntermediateTable** error message.  
- Add model name to **BackendNotGiven** error message.  
- Add property name to **RequiredProperty** error message.  
- Add property and model to **ManagedProperty** error message.  
- Add row count and location to **MultipleRowsFound** error message.  
- Add original error to **JSONError** message.  
- Add manifest **DatasetSchemaRequiresIds** 

## Removal of Unused Code  
- Remove **UnauthorizedKeymapSync** and its **test** (not used anywhere).  
- Remove **UnauthorizedPropertyPush** (not used anywhere).  
- Remove **MultipleCallsOrModelsInDependencies** (not used anywhere).  
- Remove **MultipleCommandCallsInDependencies** (not used anywhere).  

## Test Cleanup  
- Remove duplicated **UnableToMapIntermediateTable** test.

### DIFF
--- a/spinta/commands/write.py
+++ b/spinta/commands/write.py
@@ -550,11 +550,16 @@ async def read_existing_data(
         # When updating by id only, there must be exactly one existing record.
         if data.action == Action.UPSERT or _has_id_in_where(data.given):
             rows = list(itertools.islice(rows, 2))
-            if len(rows) == 1:
+            number_of_rows = len(rows)
+            if number_of_rows == 1:
                 data.saved = rows[0]
                 data.saved['_type'] = data.model.model_type()
-            elif len(rows) > 1:
-                data.error = exceptions.MultipleRowsFound(data.model, _id=rows[0]['_id'])
+            elif number_of_rows > 1:
+                data.error = exceptions.MultipleRowsFound(
+                    data.model,
+                    _id=rows[0]['_id'],
+                    number_of_rows=number_of_rows
+                )
                 report_error(data.error, data.given, stop_on_error=stop_on_error)
             elif data.action != Action.UPSERT:
                 data.error = exceptions.ItemDoesNotExist(data.model, id=data.given['_where']['args'][1])
--- a/spinta/exceptions.py
+++ b/spinta/exceptions.py
@@ -290,7 +290,7 @@ class MultipleRowsFound(BaseError):
 
 
 class ManagedProperty(UserError):
-    template = "Value of this property is managed automatically and cannot be set manually."
+    template = "Value of property ({property}) under {this} is managed automatically and cannot be set manually."
 
 
 class InvalidManagedPropertyName(UserError):
--- a/spinta/exceptions.py
+++ b/spinta/exceptions.py
@@ -875,7 +875,7 @@ class FileSizeTooLarge(UserError):
 
 
 class SRIDNotSetForGeometry(BaseError):
-    template = "Geometry SRID is required, but was given None."
+    template = "{property} Geometry SRID is required, but was given None."
 
 
 class KeyNotFound(UserError):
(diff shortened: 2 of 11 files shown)


spinta/commands/write.py
```diff
@@ -550,11 +550,16 @@ async def read_existing_data(
...
+            number_of_rows = len(rows)
+            if number_of_rows == 1:
                 data.saved = rows[0]
                 data.saved['_type'] = data.model.model_type()
-            elif len(rows) > 1:
-                data.error = exceptions.MultipleRowsFound(data.model, _id=rows[0]['_id'])
+            elif number_of_rows > 1:
+                data.error = exceptions.MultipleRowsFound(
+                    data.model,
+                    _id=rows[0]['_id'],
```
Struchid, line 560: Only taking the `_id` from the first row?

spinta/exceptions.py
```diff
@@ -286,11 +286,11 @@ class AuthorizedClientsOnly(UserError):
 
 
 class MultipleRowsFound(BaseError):
-    template = "Multiple rows were found."
+    template = "Multiple ({number_of_rows}) rows were found under {this}."
 
 
 class ManagedProperty(UserError):
-    template = "Value of this property is managed automatically and cannot be set manually."
+    template = "Value of ({property}) property under ({this}) is managed automatically and cannot be set manually."
```
Struchid, line 293: I suggest moving {this} out of the brackets, otherwise this sentence does not make sense — `property under is managed` — under what? If we retract the words in the parantheses.
```suggestion
    template = "Value of property ({property}) under {this} is managed automatically and cannot be set manually."
```

spinta/exceptions.py
```diff
@@ -740,22 +732,12 @@ class DuplicateRowWhilePaginating(BaseError):
...
 
 
 class BackendNotGiven(UserError):
-    template = "Model is operating in external mode, yet it does not have assigned backend to it."
-
-
-class UnauthorizedKeymapSync(UserError):
-    code = 403
-    template = "You do not have permission to sync this model's keymap."
+    template = "Model ({this}) is operating in external mode, yet it does not have assigned backend to it."
```
Struchid, line 740:
```suggestion
    template = "Model ({this}) is operating in external mode, yet it does not have an assigned backend to it."
```

spinta/exceptions.py
```diff
@@ -878,7 +860,7 @@ class DatasetNameMissmatch(UserError):
 
 
 class DatasetSchemaRequiresIds(UserError):
-    template = "All given schema rows require UUID to be set in the id field."
+    template = "All given schema ({this}) rows require UUID to be set in the id field."
```
Struchid, line 863: I guess this will be a whole formatted json, which would just be reformatted badly, I suggest making a multi-line error here:

```suggestion
    template = ```
         All given schema rows require UUID to be set in the id field.
         Schema: {this}
   ```
```

tests/cli/test_push.py
```diff
@@ -1151,53 +1150,6 @@ def test_push_sync_keymap(
     assert listdata(resp_city, '_id', 'id', 'name', 'country')[0] == (city_id, 1, 'Vilnius', {'_id': country_id})
 
 
-@pytest.mark.skip("Private now sends warning that model has been skipped syncing rather throwing exception")
```
Struchid, line 1154: Why was this test removed? It does not seem to be duplicated.

Struchid (approved):

tests/cli/test_push.py
varzgalys: Removed because it was testing `UnauthorizedKeymapSync`, which is not used anywhere. It's dead and skipped test.

spinta/commands/write.py
varzgalys: yes, because if multiple rows found under same `_id` - all of them will have same `_id`

PR #1258, opened 2025-04-04
`namespace` kartojimasis keliuose failuose

karina-klinkeviciute:


tests/cli/test_check.py
```diff
@@ -181,3 +181,43 @@ def test_check_access(context: Context, rc, cli: SpintaCliRunner, tmp_path):
     assert result.exit_code != 0
     assert result.exc_info[0] is InvalidValue
     assert result.exception.context.get('given') == "test"
+
+
+def test_check_namespace_two_manifests(context: Context, rc, cli: SpintaCliRunner, tmp_path):
```
Struchid, line 186: Maybe the naming could be changed? It seems like these are not two manifests files, but rather ns goes between defined datasets, when it should be atop of them

tests/cli/test_check.py
```diff
@@ -181,3 +181,43 @@ def test_check_access(context: Context, rc, cli: SpintaCliRunner, tmp_path):
...
+'''))
+
+    cli.invoke(rc, [
+        'check',
+        tmp_path / 'manifest.csv',
+
+    ])
+
+
+def test_check_namespace_two_manifests(context: Context, rc, cli: SpintaCliRunner, tmp_path):
```
Struchid, line 205: Are these two tests running correctly? Isn't one of them skipped? They are named exactly the same

Struchid (approved):

spinta/nodes.py
```diff
@@ -75,15 +75,27 @@ def get_node(
...
+                #   because dataset path can have ns as part of their path
+                # if this namespace generated, and other generated - pass. Two datasets, each in the same namespace
+                # if this namespace declared, other - generated, pass. It's possible to first declare
+                #   a dataset, then a namespace which is included in dataset's path
+                # if this namespace declared and other - declared, raise error. Namespace can be declared only once.
+                # todo collecting namespaces should be moved to linking and this checking should be done there
+                #   https://github.com/atviriduomenys/spinta/issues/1271
+                if ctype == "ns" and other_node.generated:
+                    pass
+                else:
```
varzgalys, line 90:
```suggestion
                if ctype != "ns" or not other_node.generated:
```

varzgalys (approved):

PR #1277, opened 2025-04-11
1229-dask-memory-backend

MindaugasStanionis:


martynasracct (approved):
LGTM. Added small comment. Feel free to ignore it

spinta/cli/main.py
```diff
@@ -66,7 +66,7 @@
 add(app, 'upgrade', upgrade, short_help="Run upgrade scripts")
 
 add(app, 'keymap', keymap, short_help="Manage keymap database")
-
+add(app, 'getall', get.getall, short_help="Show data from yaml dsn as json")
```
varzgalys, line 69: Is this fixed? Not a Python dict str representation?

spinta/cli/server.py
```diff
@@ -20,25 +20,28 @@
...
+        None, "-t", "--backend_type", help=("Backend type")
+    ),
 ):
     """Run development server"""
     import os
     import uvicorn
     import spinta.api
 
-    os.environ['AUTHLIB_INSECURE_TRANSPORT'] = '1'
+    os.environ["AUTHLIB_INSECURE_TRANSPORT"] = "1"
```
varzgalys, line 39: shouldnt be here, it should be set only through env variables

spinta/datasets/backends/helpers.py
```diff
@@ -39,7 +43,9 @@ def handle_ref_key_assignment(context: Context, keymap: KeyMap, env: Env, value:
             target_value = value[0]
 
         val = None
-        contains = keymap.contains(keymap_name, target_value)
+        contains = None
+        if ref.refprops[0].name != "_id":
+            contains = keymap.contains(keymap_name, target_value)
```
varzgalys, line 48: no longer needed?

varzgalys (approved):
Check comments

spinta/cli/get.py
```diff
@@ -0,0 +1,39 @@
...
+from spinta.cli.helpers.manifest import convert_str_to_manifest_path
+from spinta.cli.helpers.store import prepare_manifest, attach_keymaps
+from spinta.core.context import configure_context
+
+log = logging.getLogger(__name__)
+
+
+def getall(
+    ctx: TyperContext,
+    manifests: Optional[List[str]] = Argument(None, help=("Manifest files to load")),
```
Struchid, line 20: I suggest being consistent and using the built-in typehints:

```suggestion
    manifests: list[str] | None = Argument(None, help=("Manifest files to load")),
```

Don't forget to:

```
from __future__ import annotations
```

^ Since spinta supports older versions, that do not support the built-in typehints.

Feel free to ignore this, but I am just mentioning this here, since Karina informed me that's the good practice in the project.

tests/cli/test_get.py
```diff
@@ -0,0 +1,154 @@
...
+name: Vilnius
+country:
+  _id: 4689C28B-1C44-4184-8715-16021EE87EAD
+  name: Lietuva
+---
+_type: datasets/gov/example/Country
+_id: 4689C28B-1C44-4184-8715-16021EE87EAD
+name: Lietuva
+---
+    """
```
Struchid, line 23: Please, add the missing tabs so the multi-line comment falls into the method

tests/cli/test_get.py
```diff
@@ -0,0 +1,154 @@
...
+       |   |                           |         |         |            |
+       |   |   | Country               |         | _id     |            | open
+       |   |   |    | _id              | string  |         | _id        |
+       |   |   |    | name             | string  |         | name       |
+       |   |   |    |                  |         |         |            |
+       |   |   | City                  |         |         |            | open
+       |   |   |    | _id              | string  |         | _id        |
+       |   |   |    | name             | string  |         | name       |
+       |   |   |    | country          | ref     | Country | ..         |
+        """
```
Struchid, line 73: Could we format these DSA's correctly as well? And I see a trailing `"""` which should not be here I think? Same for the other two tests as well

Struchid (approved):

spinta/datasets/backends/helpers.py
MindaugasStanionis: Needed, because we need to explicitly check if we are passing "_id", because in theory, its reserved as internal id, that is generated from keymap

spinta/cli/main.py
MindaugasStanionis: yes, now this returns json representation. Data is passed in json.dumps

PR #1291, opened 2025-04-18
1290 redirect palaikymas postgresql backend

JustinasKen:


spinta/commands/read.py
```diff
@@ -406,9 +406,14 @@ async def getone(
...
-    data = commands.getone(context, model, backend, id_=params.pk)
-    data = next(prepare_data_for_response(context, model, action, params, data, reserved=[]))
-    return render(context, request, model, params, data, action=action)
+    try:
+        data = commands.getone(context, model, backend, id_=params.pk)
+        data = next(prepare_data_for_response(context, model, action, params, data, reserved=[]))
+        return render(context, request, model, params, data, action=action)
+    except ItemDoesNotExist as e:
+        if redirect_id := commands.redirect(context, backend, model, params.pk):
+            return RedirectResponse(request.url.path.replace(str(params.pk), str(redirect_id), 1), status_code=301)
```
sirex, line 415: Consider using:

https://github.com/atviriduomenys/spinta/blob/1e0ff6cbdf7e580d125863860c43e592ddd93659/spinta/formats/html/helpers.py#L205

JustinasKen: In this case, `get_model_link` might not work the best, since the goal is to only replace `path` part and leave everything else as it was.
We could possibly still use `build_url_path`, so it would look something like this:
```python
ptree = params.changed_parsetree({
    'path': (
        model.name.split('/') +
        [redirect_id] +
        ([params.prop.place] if params.prop is not None else [])
    )
})
result = '/' + build_url_path(ptree)
return RedirectResponse(result, status_code=301)
```

PR #1308, opened 2025-05-19
279 Nested resource implementation for WSDL/SOAP

martynasracct:
**Summary:**
- Implements `type=wsdl` resource, that describes only WSDL endpoint and `Zeep` client initialization
- Changes `type=soap` resource to describe SOAP operation based on SOAP response. 
  - `source` column now holds path to SOAP operation (previously this was in `Model.source`) 
- Implements `LinkBuilder` env that is called during resource linking to link resources using `ufuncs`
- Implements `prepare=wsdl(<wsdl-resource-name>)` `ufunc` that parses path to SOAP operation and saves operation callable to soap resource backend

Model `source` column is not used for now. Later, it will be used to map response data to DSA models

**Ticket:**
https://github.com/atviriduomenys/spinta/issues/279

### DIFF
--- a/spinta/datasets/backends/wsdl/components.py
+++ b/spinta/datasets/backends/wsdl/components.py
@@ -1,0 +1,31 @@
+from __future__ import annotations
+import contextlib
+import logging
+
+from requests import RequestException
+from zeep import Client as ZeepClient
+from zeep.exceptions import Error as ZeepError
+
+from spinta.datasets.components import ExternalBackend
+
+log = logging.getLogger(__name__)
+
+
+class WsdlBackend(ExternalBackend):
+    type: str = "wsdl"
+    client: ZeepClient | None
+
+    @contextlib.contextmanager
+    def begin(self):
+        self.client = self.get_client()
+        yield
+
+    def get_client(self) -> ZeepClient | None:
+        client = None
+        wsdl_url = self.config.get("dsn")
+        try:
+            client = ZeepClient(wsdl=wsdl_url)
+        except (RequestException, ZeepError) as err:
+            log.error(f"Zeep client initialization with url %s failed. Error: %s", wsdl_url,  err)
+
+        return client
--- a/spinta/ufuncs/linkbuilder/ufuncs.py
+++ b/spinta/ufuncs/linkbuilder/ufuncs.py
@@ -1,0 +1,62 @@
+import logging
+
+from spinta.core.ufuncs import ufunc, Bind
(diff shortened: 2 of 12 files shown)


spinta/datasets/backends/wsdl/components.py
```diff
@@ -0,0 +1,31 @@
...
+        self.client = self.get_client()
+        yield
+
+    def get_client(self) -> ZeepClient | None:
+        client = None
+        wsdl_url = self.config.get("dsn")
+        try:
+            client = ZeepClient(wsdl=wsdl_url)
+        except (RequestException, ZeepError) as err:
+            log.error(f"Zeep client initialization with url %s failed. Error: %s", wsdl_url,  err)
```
Struchid, line 29: Using the following approach you can quickly deduct which of the variables will be in which place. Otherwise, passing the variables as additional arguments, will have you looking left-to-right multiple times to check which of the varialbes `%s` maps to.

Well, at least that's what I think. Other than that, nor flake8, nor mypy is enabled, so can't think of any other reason not to use the following approach, let me know if I am just inexperienced:d

```suggestion
            log.error(f"Zeep client initialization with url {wsdl_url} failed. Error: {err}")
```

spinta/ufuncs/linkbuilder/ufuncs.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@ufunc.resolver(LinkBuilder, WsdlBackend)
+def wsdl(env: LinkBuilder, parent_resource_backend: WsdlBackend) -> None:
+    soap_source = env.resource.backend.config.get("dsn")
+
+    with parent_resource_backend.begin():
+        if not (client := parent_resource_backend.client):
+            raise SoapServiceError(
+                f'Cannot parse SOAP operation "{soap_source}", because WSDL client is not initialized.'
```
Struchid, line 40: Not sure about this, but would it be a good idea to also point out to the user how that is done here? Might be more complex than I think though, so do as you please after reading this

Struchid (approved):
lgtm. The comments are all optional, so do with them as you please. 

Other than that, nice work!

karina-klinkeviciute (approved):

karina-klinkeviciute:
Priminimas: dar reikėtų atnaujinti `changes.rst`

TadasVasiliauskasCCT (approved):

karina-klinkeviciute (approved):

PR #1349, opened 2025-06-18
1261/1262 add model and property reading from JSON schema

gts446:


spinta/manifests/open_api/helpers.py
```diff
@@ -50,9 +59,93 @@ def get_resource_parameters(parameters: list[dict]) -> dict[str, dict]:
...
+    models.extend(nested_models)
+    return models
+
+def build_properties(dataset: str, resource: str, properties_schema: dict, model_name: str) -> tuple[dict, list[dict]]:
+
+    properties = {}
+    nested_models = []
+
+    for property_source_name, property_metadata in properties_schema.items():
+        property_type = property_metadata['type']
```
martynasracct, line 106: What would happen, if `type` is is not in `property_metadata`?

PR #1352, opened 2025-06-25
ft/1275 create/update backends via client crud endpoints

martynasracct:
**Summary:**
49a0a7d35a7a18038912b84585610458805e2e55 - Refactors client CRUD endpoints:
  - Adds `Pydantic` for request data validation.
  - Instead of `UnknownPropertyInRequest`, `InsufficientPermissionForUpdate` and `EmptyPassword` exceptions `ClientValidationError` is now raised.
  - Raises HTTP 400 instead of HTTP 403 when when user without `auth_clients` scope tried changing its own client fields (user without `auth_clients` scope can only change its own client secret).
  - Reduced size of try-except blocks.

3a5eaa08f421b1664b10ad6d9fe3850e038a600a - Save/Update `backends` via client POST/PATCH

Also, removes `clean_up_auth_client`and `ensure_clients_dont_exist` methods from tests, since tests already use `tmp_path` fixture which already cleans up temp files

**Ticket:**


spinta/api/validators.py
```diff
@@ -0,0 +1,24 @@
+from typing import Optional, Any
+
+from pydantic import BaseModel, Field, ConfigDict
+
+
+class ClientAddData(BaseModel):
+    client_name: Optional[str] = None
```
Struchid, line 7: In other places the pipes have been used to define types, maybe let's follow that here as well?

```suggestion
    client_name: str | None = None
```

spinta/api/__init__.py
```diff
@@ -132,155 +134,162 @@ def _auth_client_context(request: Request) -> Context:
...
+    return Response(status_code=204)
+
 
-async def auth_clients_patch_specific(request: Request):
+async def auth_clients_patch_specific(request: Request) -> JSONResponse:
+    context = _auth_client_context(request)
+    token = context.get('auth.token')
+    client_id = request.path_params["client"]
+
+    validator_cls = ClientPatchData
```
Struchid, line 251: **Optional:** I'd use the full name without the abbreviation. Mainly, because cls could have more than one different meaning

spinta/auth.py
```diff
@@ -724,6 +737,7 @@ def update_client_file(
         new_name = name if name else client.name
         new_secret_hash = passwords.crypt(secret) if secret else client.secret_hash
         new_scopes = scopes if scopes is not None else client.scopes
+        new_backends = backends if backends is not None else client.backends
```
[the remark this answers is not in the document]
martynasracct: Yes, I want to only avoid `None`. `{}` are allowed so I cannot use `or`.

spinta/api/validators.py
Struchid: Oh, that's a bummer, I thought improting annotations from future would suffice, ignore then 😑

spinta/api/__init__.py
```diff
@@ -132,155 +134,162 @@ def _auth_client_context(request: Request) -> Context:
...
-        }
-        return JSONResponse(return_value)
 
-    except InsufficientScopeError:
-        raise InsufficientPermission(scope=Scopes.AUTH_CLIENTS)
+    new_data = update_client_file(
+        context,
+        path,
+        client_id,
+        name=client_data["client_name"] if client_data.get("client_name") else None,
```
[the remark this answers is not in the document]
martynasracct: Good catch

karina-klinkeviciute:
kur yra užduotis šitam PR? Man aiškiau būna peržiūrėti, kai matau užduotį :)

TadasVasiliauskasCCT (approved):
LGTM

PR #1355, opened 2025-06-30
Fix/1275 resolve resource param on load

martynasracct:
**Summary:**

**Note:** only check commit 49a0a7d35a7a18038912b84585610458805e2e55. Other commits are introduced in https://github.com/atviriduomenys/spinta/pull/1352. If I create this PR with base branch other than master - CI/CD does not run :(

Refactor resource.param `input(...)` prepare function resolving. It is now resolved during resource `load(...)` instead of resolving it on every Soap request.
`LoadBuilder` is used for resolving. Resolved value is saved in `Param.soap_body`

**Ticket:**


spinta/exceptions.py
```diff
@@ -638,8 +629,8 @@ class ClientAlreadyExists(UserError):
     template = "Client '{client_id}' already exists."
 
 
-class EmptyPassword(UserError):
-    template = "Client password cannot be empty."
+class ClientValidationError(UserError):
+    template = "Errors validating client request."
```
Struchid, line 633: **Optional:** I suggest a "fuller" message:
```suggestion
    template = "Errors occurred while validating client request."
```

spinta/dimensions/param/helpers.py
```diff
@@ -7,17 +7,35 @@
...
+def load_param_formulas(context: Context, param: Param, prepare_asts: list[dict]) -> None:
+    formulas = []
+    builder = LoadBuilder(context)
+    for source, ast in zip(param.sources, prepare_asts):
+        builder.update(this=source, param=param)
+
+        # If builder cannot resolve formula - save it in formula and leave resolving for later
+        expr = asttoexpr(ast)
+        if isinstance(expr, Expr):
+            if formula := builder.resolve(expr):
```
Struchid, line 23: This could be done, since the second part (after `and`) won't be evaluated if the first part fails:
```suggestion
        if isinstance(expr, Expr) and (formula := builder.resolve(expr)):
```

spinta/datasets/backends/dataframe/backends/soap/ufuncs/ufuncs.py
```diff
@@ -62,20 +63,27 @@ def soap_request_body(env: SoapQueryBuilder, prop: Property) -> None:
     """
     if prop.external.name:
         # Ignore properties with `source`
-        return None
+        return
 
     if not isinstance(prop.external.prepare, Expr):
         # Ignore properties without `prepare` expression
-        return None
+        return
```
Struchid, line 70: **Ultra minor optional:** I would leave these explicit returns due to the Zen:

image

wisecrow (approved):

spinta/dimensions/param/helpers.py
```diff
@@ -7,17 +7,35 @@
...
 from spinta.utils.schema import NA
 
 
+def load_param_formulas(context: Context, param: Param, prepare_asts: list[dict]) -> None:
+    formulas = []
+    builder = LoadBuilder(context)
+    for source, ast in zip(param.sources, prepare_asts):
+        builder.update(this=source, param=param)
+
+        # If builder cannot resolve formula - save it in formula and leave resolving for later
```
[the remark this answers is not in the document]
martynasracct: This is a common practice in Spinta. Formulas (or expressions) goes through multiple builders that tries to resolve them in different stages of the app. I changed comment a bit

PR #1365, opened 2025-07-02
ft/1275 implement creds() prepare function

martynasracct:
**Summary:**
Adds `creds("key")` prepare function that used with resource parameters that reads extra data service authentication data from client file. If `creds(...).input()` used - it will read `key` from client file and will use it in SOAP request body.

For this to work, client file must have backends saved:
```yml
client_id: ...
...
...
backends:        # Hardcoded element where we save these values
  resource:      # Resource name 
    key: value   # Actual key-value pairs  
```
This can be done via API calls: `POST /auth/clients` and `PATCH /auth/clients/{client}`

Extra authentication data must be saved for each resource separately.

**Ticket:**

### DIFF
--- a/spinta/datasets/backends/dataframe/backends/soap/ufuncs/ufuncs.py
+++ b/spinta/datasets/backends/dataframe/backends/soap/ufuncs/ufuncs.py
@@ -2,12 +2,14 @@ from __future__ import annotations
 
 from typing import Any
 
-from spinta.auth import authorized
+from spinta.auth import authorized, query_client
 from spinta.components import Property
 from spinta.core.enums import Action
 from spinta.core.ufuncs import ufunc, Expr, Bind, ShortExpr
 from spinta.datasets.backends.dataframe.backends.soap.ufuncs.components import SoapQueryBuilder
 from spinta.datasets.components import Param
+from spinta.exceptions import InvalidClientBackendCredentials, InvalidClientBackend, UnknownMethod
+from spinta.utils.config import get_clients_path
 from spinta.utils.data import take
 from spinta.utils.schema import NA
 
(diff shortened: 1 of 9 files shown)


spinta/datasets/backends/dataframe/backends/soap/ufuncs/ufuncs.py
```diff
@@ -75,15 +95,38 @@ def soap_request_body(env: SoapQueryBuilder, prop: Property) -> None:
...
+        param.soap_body[param_body_key] = value
+
+    param_value = list(param.soap_body.values())[0] if param.soap_body else None
 
-    env.soap_request_body.update(param_soap_body)
     env.property_values.update({param.name: param_value})
+
+
+@ufunc.resolver(SoapQueryBuilder, str)
+def creds(env: SoapQueryBuilder, credential_key: str) -> Any:
```
wisecrow, line 111:
```suggestion
def creds(env: SoapQueryBuilder, credential_key: str) -> str:
```

spinta/datasets/backends/dataframe/backends/soap/ufuncs/ufuncs.py
```diff
@@ -42,16 +44,34 @@ def and_(env: SoapQueryBuilder, args: list) -> Any:
...
+        for param_body_key, param_body_value in param.soap_body.items():
+            if isinstance(param_body_value, Expr):
+                param.soap_body[param_body_key] = env.resolve(param_body_value)
+
+
+def _populate_param_soap_body_with_url_values(env: SoapQueryBuilder) -> None:
+    for prop in take(env.model.properties).values():
+        if not authorized(env.context, prop, Action.GETALL):
+            continue
+        env.call("soap_request_body", prop)
```
Struchid, line 60: **Optional:** Again, not a biggie, but in the other two functions in this file you are checking for `not` and using `continue`, it might be nice to stay with that approach for all the conditions then to decrease the mental load, but keep in mind, that it might just be me:d
```suggestion
def _finalize_param_soap_body_resolve(env: SoapQueryBuilder) -> None:
    for param in env.params.values():
        if not hasattr(param, "soap_body"):  # Skip params without resolved .input(...) function
            continue
        for param_body_key, param_body_value in param.soap_body.items():
            if not isinstance(param_body_value, Expr):
                continue
            param.soap_body[param_body_key] = env.resolve(param_body_value)


def _populate_param_soap_body_with_url_values(env: SoapQueryBuilder) -> None:
    for prop in take(env.model.properties).values():
        if not authorized(env.context, prop, Action.GETALL):
            continue
        env.call("soap_request_body", prop)
```

spinta/datasets/backends/dataframe/backends/soap/ufuncs/ufuncs.py
```diff
@@ -75,15 +95,38 @@ def soap_request_body(env: SoapQueryBuilder, prop: Property) -> None:
...
     env.property_values.update({param.name: param_value})
+
+
+@ufunc.resolver(SoapQueryBuilder, str)
+def creds(env: SoapQueryBuilder, credential_key: str) -> Any:
+    client_config_file = query_client(
+        get_clients_path(env.context.get('config')),
+        env.context.get('auth.token').get_aud()
+    )
+    backend_name = env.model.external.resource.name
```
Struchid, line 116: Will `env.model` always have the `external` attribute set?

spinta/exceptions.py
```diff
@@ -967,6 +967,14 @@ class InvalidScopes(UserError):
     template = "Request contains invalid, unknown or malformed scopes: {scopes}."
 
 
+class InvalidClientBackend(UserError):
+    template = '''Backend "{backend_name}" does not exist in configured client's backends.'''
+
+
+class InvalidClientBackendCredentials(UserError):
+    template = '''Credential "{key}" does not exist in client's backend "{backend_name}".'''
```
Struchid, line 975: Not sure that's why I am asking, but, will these messages be enough for the end-user that will receive them?

Struchid (approved):

spinta/exceptions.py
martynasracct: 🤷‍♂️

Issue #986, opened 2024-11-28
Tikrinant spintos versiją metama error


LaurynasVenck:
Suinstaliuota ant lokalaus environment (Ubuntu 24.04) PYENV ir atsisiusta su PIP šiandien (11-28 spinta).
Paleidus ENV ir bandant gauti spintos versiją metama klaida:

spinta@laurynas:~$ source venv/bin/activate
(venv) spinta@laurynas:~$ spinta --version
Traceback (most recent call last):
  File "/opt/spinta/venv/bin/spinta", line 5, in <module>
    from spinta.cli.main import app
  File "/opt/spinta/venv/lib/python3.13/site-packages/spinta/cli/main.py", line 14, in <module>
    from spinta.cli import auth
  File "/opt/spinta/venv/lib/python3.13/site-packages/spinta/cli/auth.py", line 16, in <module>
    from spinta.auth import KeyFileExists, get_clients_path, ensure_client_folders_exist
  File "/opt/spinta/venv/lib/python3.13/site-packages/spinta/auth.py", line 47, in <module>
    from spinta.utils.config import get_clients_path, get_keymap_path, get_id_path, get_helpers_path
  File "/opt/spinta/venv/lib/python3.13/site-packages/spinta/utils/config.py", line 5, in <module>
    from spinta.core.config import RawConfig, DEFAULT_CONFIG_PATH
  File "/opt/spinta/venv/lib/python3.13/site-packages/spinta/core/config.py", line 37, in <module>
    resource_filename('spinta', 'config.yml').
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/spinta/venv/lib/python3.13/site-packages/spinta/utils/path.py", line 32, in resource_filename
    with importlib.resources.files(package) / target as ref:
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
TypeError: 'PosixPath' object does not support the context manager protocol

karina-klinkeviciute:
Pabandžiau taip (Ubuntu 22.04):

```
python3 -m venv venv
source venv/bin/activate
pip install spinta
spinta --version
```
Klaidos negavau, gavau atsakymą:
`0.1.79`

Vėliau pabandysiu su Ubuntu 24, bet gal galėtum parašyti tikslias komandas, kurias leidai?

LaurynasVenck:
curl https://pyenv.run/ | bash
[...]

PR #1368, opened 2025-07-08
986 PosixPath droped support for Context Manager protoclo in Python 3.13

04d4:
Solves #986


karina-klinkeviciute (approved):

spinta/utils/path.py
```diff
@@ -1,8 +1,12 @@
...
-from contextlib import ExitStack
-from pathlib import Path
+
+try:
+    # for Python >=3.10
+    from importlib import resources
+    from importlib.resources import abc
+except ImportError:
+    # Python <=3.9
+    from importlib import resources, abc
```
Struchid, line 9: **Optional:** I'd use `sys` & `sys.version_info` for checking the version and importing the libraries according to that, but I don't see much of a problem with the try-expect path as well.

- https://docs.python.org/3/library/sys.html#sys.version_info

Struchid (approved):

PR #1384, opened 2025-07-15
1260 open api nuskaitymas model ir model property

gts446:
**Add OpenAPI (and Swagger 2.0) JSON Schema to DSA Conversion for Models and Properties**

Implemented conversion from OpenAPI/Swagger JSON schemas to DSA structure, supporting both Model and ModelProperty dimensions.
Only flattened JSON schemas are supported.

Implemented functionality:

- Parse and register root-level models
- Handle nested models
- Link ref and backref relationships
- Extract model properties
- Detect ModelProperty datatypes
- Read ModelProperty enums

### DIFF
--- a/spinta/manifests/open_api/helpers.py
+++ b/spinta/manifests/open_api/helpers.py
@@ -140,7 +140,6 @@ class Model:
                 "name": self.name,
                 "title": self.title,
                 "description": self.description,
-                "access": "open",
                 "external": {
                     "dataset": self.dataset,
                     "resource": self.resource,
@@ -176,6 +175,7 @@ class Property:
         self.datatype: str = datatype or self.get_datatype()
         self.ref: Model = ref_model
         self.enum: dict = self.get_enums(self.json_schema.get("enum", []))
+        self.required: bool = not bool(json_schema.get("nullable", False))
 
 
     def get_datatype(self) -> str:
@@ -219,6 +219,7 @@ class Property:
             "title": self.title,
             "description": self.description,
             "external": {"name": self.source},
+            "required": self.required,
         }
 
         if self.datatype in ["ref", "backref"]:
--- a/spinta/manifests/open_api/helpers.py
+++ b/spinta/manifests/open_api/helpers.py
@@ -97,14 +97,21 @@ class Model:
 
             self.properties.append(prop)
 
+    def add_child(self, basename:str, json_schema: dict) -> Model:
+        """
+        TODO: Children model "source" should be empty and name should have "/:part" suffix. e.g. f"{self.name}/:part"
+        Link to task https://github.com/atviriduomenys/spinta/issues/997
+        """
+        model = Model(self.dataset, self.resource, basename, json_schema)
+        self.children.append(model)
+        return model
+
     def _add_ref_model(self, parent_prop: Property) -> None:
-        child_model = Model(self.dataset, self.resource, parent_prop.basename, parent_prop.json_schema)
-        parent_prop.ref = child_model
-        self.children.append(child_model)
+        parent_prop.ref = self.add_child(parent_prop.basename, parent_prop.json_schema)
 
     def _add_backref_model(self, parent_prop: Property) -> None:
         json_schema = parent_prop.json_schema.get("items", {})
-        child_model = Model(self.dataset, self.resource, parent_prop.basename, json_schema)
+        child_model = self.add_child(parent_prop.basename, json_schema)
 
         backref_prop = Property(parent_prop.basename, {}, datatype="backref", source=".", ref_model=child_model)
         backref_prop.name += "[]"
@@ -113,8 +120,6 @@ class Model:
         ref_prop = Property(self.basename, {}, datatype="ref", source="", ref_model=self)
         ref_prop.ref = self
         child_model.properties.append(ref_prop)
-
-        self.children.append(child_model)
 
     def get_node_schema_dicts(self) -> list[dict]:
         schena_dict = [
--- a/spinta/manifests/open_api/helpers.py
+++ b/spinta/manifests/open_api/helpers.py
@@ -147,12 +147,14 @@ class Property:
     ):
         self.basename: str = basename
         self.json_schema: dict = json_schema
-        self.name = to_property_name(basename)
-        self.title = json_schema.get("title", "")
-        self.description = json_schema.get("description", "")
-        self.source = basename if source is None else source
-        self.datatype = datatype or self.get_datatype()
-        self.ref = ref_model
+        self.name: str = to_property_name(basename)
+        self.title: str = json_schema.get("title", "")
+        self.description: str = json_schema.get("description", "")
+        self.source: str = basename if source is None else source
+        self.datatype: str = datatype or self.get_datatype()
+        self.ref: Model = ref_model
+        self.enum: dict = self.get_enums(self.json_schema.get("enum", []))
+
 
     def get_datatype(self) -> str:
         """
@@ -180,6 +182,14 @@ class Property:
                 return date_time_types[string_format]
 
         return "string"
+    
+    def get_enums(self, items:list) -> dict:
+        enum = {}
+        for item in items:
+            if isinstance(item,(list, dict)):  # Only handling primitive type enums
+                return {}
+            enum[item] = {"source": item}
+        return enum
 
     def get_node_schema_dict(self) -> dict:
         schema = {
@@ -191,6 +201,9 @@ class Property:
 
         if self.datatype in ["ref", "backref"]:
             schema["model"] = self.ref.name
+
+        if self.enum:
+            schema['enums'] = {"": self.enum}
 
         return schema
 
@@ -238,7 +251,6 @@ def get_dataset_schemas(data: dict, dataset_prefix: str) -> Generator[tuple[None
     for api_endpoint, api_metadata in data.get("paths", {}).items():
         for http_method, http_method_metadata in api_metadata.items():
             tags = http_method_metadata.get("tags", [])
-
             dataset_name = to_dataset_name("_".join(tags)) or DEFAULT_DATASET_NAME  # Default dataset if no tags given.
             if dataset_name not in datasets:
                 datasets[dataset_name] = {
--- a/spinta/manifests/open_api/helpers.py
+++ b/spinta/manifests/open_api/helpers.py
@@ -93,16 +93,24 @@ class Model:
         self.add_properties()
 
     def add_properties(self) -> None:
-        for property_name, property_metadata in self.json_schema.get("properties", {}).items():
-            prop = Property(property_name, property_metadata)
+        for property_name, json_schema in self.json_schema.get("properties", {}).items():
+            prop = self.add_property(property_name, json_schema)
 
-            if property_metadata.get("type") == "object":
+            if json_schema.get("type") == "object":
                 self._add_ref_model(prop)
 
-            elif property_metadata.get("items", {}).get("type") == "object":
-                self._add_backref_model(prop)
+            elif json_schema.get("type") == "array":
+                json_schema = json_schema.get("items", {})
+                if json_schema.get("type") == "object":
+                    self._add_backref_model(prop)
+                else:
+                    prop = self.add_property(property_name, json_schema, source="")
+                    prop.name += "[]"
 
-            self.properties.append(prop)
+    def add_property(self, name:str, json_schema:dict, **kwargs) -> Property:
+        prop = Property(name, json_schema, **kwargs)
+        self.properties.append(prop)
+        return prop
 
     def add_child(self, basename:str, json_schema: dict) -> Model:
         """
@@ -120,18 +128,16 @@ class Model:
         json_schema = parent_prop.json_schema.get("items", {})
         child_model = self.add_child(parent_prop.basename, json_schema)
 
-        backref_prop = Property(parent_prop.basename, {}, datatype="backref", source=".", ref_model=child_model)
+        backref_prop = self.add_property(parent_prop.basename, {}, datatype="backref", source=".", ref_model=child_model)
         backref_prop.name += "[]"
-        self.properties.append(backref_prop)
         """
         TODO: Temporary workaround — adds a ref to the backref's model. 
         This should be removed once proper backref handling is implemented.
 
         See task: https://github.com/atviriduomenys/spinta/issues/1314
         """
-        ref_prop = Property(self.basename, {}, datatype="ref", source="", ref_model=self)
-        ref_prop.ref = self
-        child_model.properties.append(ref_prop)
+        child_model.add_property(self.basename, {}, datatype="ref", source="", ref_model=self)
+
 
     def get_node_schema_dicts(self) -> list[dict]:
         schena_dict = [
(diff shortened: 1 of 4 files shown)


spinta/manifests/open_api/helpers.py
Struchid, line 1: No description was given on the PR, please provide a summary of what you've tried to achieve with this PR.

Struchid, line 1: Does this PR really solve 4 tickets? Might be helpful to have some separation (separate PR's) in the future

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+def resolve_refs(schema: dict, root: dict) -> dict:
+    if isinstance(schema, dict):
+        if "$ref" in schema:
+            ref_path = schema["$ref"].lstrip("#/").split("/")
+            resolved = root
+            for part in ref_path:
+                resolved = resolved[part]
+            return resolve_refs(resolved, root)
+
+        return {key: resolve_refs(value, root) for key, value in schema.items()}
```
Struchid, line 37: I am not a fan of recursively processing and adding values to a mutable type - dictionary, it could lead to unexpected side effects and some hair might get hurt while trying to resolve those unexpected outcomes.

@lescijuscct can you please add some insights on what do you think about that?

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
-            'type': 'param',
-            'description': value.get('description', ''),
+        name = value["name"]
+        location = value["in"] if value["in"] in SUPPORTED_PARAMETER_LOCATIONS else ""
+        resource_parameters[f"parameter_{index}"] = {
+            "name": to_code_name(name),
+            "source": [name],
+            "prepare": [Expr(location)],
+            "type": "param",
+            "description": value.get("description", ""),
```
Struchid, line 64: Ok this time, but let's avoid formatting old changes with ruff/black, the changes get mixed up and it's harder to review the PR this way.

Best way is to:
- Do the ruff/black refactor before you touch a file, merge that into main/master (since it should not break anything related, and if it does, tests should inform about it)
- Start making your changes on the refactored code

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+        source: str | None = None,
+        parent: Model | None = None
+    ) -> None:
+        self.dataset: str = dataset
+        self.resource: str = resource
+        self.basename: str = basename
+        self.json_schema: dict = json_schema
+        self.source: str = source if source is not None else basename
+        self.title: str = self.json_schema.get("title")
+        self.description: str = self.json_schema.get("description")
```
Struchid, line 88: Can `title`/`description` be `None`? Than the typing would be innacurrate

Might want this:
```suggestion
        self.title: str = self.json_schema.get("title", "")
        self.description: str = self.json_schema.get("description", "")
```

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+        self.children: list[Model] = []
+        self.properties: list[Property] = []
+        self.parent = parent
+        self.add_properties()
+
+    def add_properties(self) -> None:
+        for property_name, json_schema in self.json_schema.get("properties", {}).items():
+            prop = self.add_property(property_name, json_schema)
+
+            if json_schema.get("type") == "object":
```
Struchid, line 99: Extract `json_schema.get("type")` into a new variable, it's re-used at least thrice.

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+
+            elif json_schema.get("type") == "array":
+                json_schema = json_schema.get("items", {})
+                if json_schema.get("type") == "object":
+                    self._add_backref_model(prop)
+                else:
+                    prop = self.add_property(property_name, json_schema, source="")
+                    prop.name += "[]"
+
+    def add_property(self, name:str, json_schema:dict, **kwargs) -> Property:
```
Struchid, line 110: Spaces after colons in typing (for the whole file, not just this one place)
```suggestion
    def add_property(self, name: str, json_schema: dict, **kwargs: Any) -> Property:
```

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+        TODO: Temporary workaround — adds a ref to the backref's model. 
+        This should be removed once proper backref handling is implemented.
+
+        See task: https://github.com/atviriduomenys/spinta/issues/1314
+        """
+        child_model.add_property(self.basename, {}, datatype="ref", source="", ref_model=self)
+
+
+    def get_node_schema_dicts(self) -> list[dict]:
+        schena_dict = [
```
Struchid, line 143:
```suggestion
        schema_dict = [
```

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+                    "prepare": Expr("expand") if self.parent else ""
+                    },
+                "properties": {prop.name: prop.get_node_schema_dict() for prop in self.properties},
+            }
+        ]
+        for child in self.children:
+            schena_dict.extend(child.get_node_schema_dicts())
+        return schena_dict
+
+    def __repr__(self) -> str:
```
Struchid, line 162: Let's keep all dunder methods at the top of the file.

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+        self.source: str = basename if source is None else source
+        self.datatype: str = datatype or self.get_datatype()
+        self.ref: Model = ref_model
+        self.enum: dict = self.get_enums(self.json_schema.get("enum", []))
+        self.required: bool = not bool(json_schema.get("nullable", False))
+
+
+    def get_datatype(self) -> str:
+        """
+        Returns Property datatype as a string. If type cannot be detected, defaults to "string"
```
Struchid, line 189:
```suggestion
        Returns "Property" data type as a string. If type cannot be detected, defaults to "string"
```

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+        """
+        datatype = self.json_schema.get("type", "string")
```
Struchid, line 191:
```suggestion
        data_type = self.json_schema.get("type", "string")
```

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+            "array": "array",
+            "object": "ref",
+        }
+
+        date_time_types = {"date-time": "datetime", "date": "date", "time": "time"}
+
+        if datatype in basic_types:
+            return basic_types[datatype]
+
+        if datatype == "string":
```
Struchid, line 206: Constants for re-used hard-coded strings

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+        """
+        datatype = self.json_schema.get("type", "string")
+
+        basic_types = {
+            "boolean": "boolean",
+            "integer": "integer",
+            "number": "number",
+            "array": "array",
+            "object": "ref",
+        }
```
Struchid, line 199: Could be an enum

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+
+
+def get_schema_from_response(response: dict, root: dict) -> dict:
+    json_schema = {}
+    if "openapi" in root:
+        for content_type, content in response.get("content", {}).items():
+            if content_type == "application/json" or content_type.endswith("+json"):
+                json_schema = resolve_refs(content.get("schema", {}), root)
+                break
+    elif "swagger" in root:
```
Struchid, line 247: Is this how you check for the specific 2.0 version of swagger or do you need something additional to check the version?

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+        json_schema = resolve_refs(response.get("schema", {}), root)
+    else:
+        raise ValueError("Unknown specification type. Only OpenAPI 3.* and Swagger 2.0 allowed.")
```
Struchid, line 250: This might be not as user-friendly as we'd want to, check `spinta/exceptions.py ` and the implementation of errors there to understand how a more user-friendly error could be made

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,325 @@
...
+
+    return json_schema
+
+
+def get_model_schemas(dataset_name: str, resource_name: str, response_200: dict, root: dict) -> list[dict]:
```
Struchid, line 255: The others should not even come here (unless 201 is also an option), but still, I'd keep it more generalised.

```suggestion
def get_model_schemas(dataset_name: str, resource_name: str, response: dict, root: dict) -> list[dict]:
```

Struchid (commented):
This PR can not be approved without tests that prove the changes that have been done:

- `tests/manifests/open_api/test_helpers.py`
- `tests/manifests/open_api/test_openapi.py`

spinta/manifests/open_api/helpers.py
lescijuscct: I would suggest split to multiple PR or make meaningfull commits.

spinta/manifests/open_api/helpers.py
lescijuscct: I think if the depth is unknown or known but larger than 3, recursion can be used.
But there are a few problems:
1) There is no guard against an infinite cycle
2) There is no guard for multiple same keys which would overwrite.
Also can the ref be a list or is it always a dict?

spinta/manifests/open_api/helpers.py
gts446: Yes. There is "swagger": "2.0" and then it became "openapi": "3.*"

spinta/manifests/open_api/helpers.py
karina-klinkeviciute: I would even suggest to do this as a warning, not an error and try to parse it anyway. We might successfully parse it, after all :)

spinta/manifests/open_api/helpers.py
```diff
@@ -2,99 +2,353 @@
...
-    return re.sub(r'{([^{}]+)}', lambda match: f'{{{to_code_name(match.group(1))}}}', endpoint)
+    return re.sub(r"{([^{}]+)}", lambda match: f"{{{to_code_name(match.group(1))}}}", endpoint)
 
 
 def read_file_data_and_transform_to_json(path: Path) -> dict:
     with Path(path).open() as file:
         return json.load(file)
 
 
+def resolve_refs(schema: dict, root: dict, seen: set[str] = None) -> dict:
```
karina-klinkeviciute, line 30: It shouldn't replace refs with their counterpart schemas. As we talked, this might create duplicate models where they aren't needed. Maybe check how it's done in the XSD schema parsing. It's a bit complex, but it solves all the issues without creating duplicates. 

I'd suggest to move `ref` handling to a different task (create a new ticket) and a different PR.

karina-klinkeviciute: I would suggest not to handle ref here at all, but do it properly from the beginning

spinta/manifests/open_api/helpers.py
gts446: its used twice. 3rd time is different value.

tests/manifests/open_api/test_openapi.py
```diff
@@ -125,3 +125,339 @@ def test_open_api_manifest_title_with_slashes(rc: RawConfig, tmp_path: Path):
...
+    """
+ 
+    path = tmp_path / 'manifest.json'
+    path_openapi = f'openapi+file://{path}'
+    with open(path, 'w') as json_file:
+        json_file.write(data)
+
+    manifest = load_manifest(rc, path_openapi)
+
+    assert manifest == table
```
martynasracct, line 463: New line at the end of the file missing

martynasracct (approved):
One small comment

spinta/manifests/open_api/helpers.py
```diff
@@ -27,74 +32,295 @@ def read_file_data_and_transform_to_json(path: Path) -> dict:
...
+                return "binary"
+            if (string_format := self.json_schema.get("format")) in date_time_types:
+                return date_time_types[string_format]
+
+        return DEFAULT_PROPERTY_DATATYPE
+    
+    def get_enums(self, items: list) -> dict:
+        enum = {}
+        for item in items:
+            if isinstance(item,(list, dict)):  # Only handling primitive type enums
```
Struchid, line 206: Formatting:

```suggestion
            if isinstance(item, (list, dict)):  # Only handling primitive type enums
```

spinta/manifests/open_api/helpers.py
```diff
@@ -27,74 +32,295 @@ def read_file_data_and_transform_to_json(path: Path) -> dict:
...
+
+def get_schema_from_unknown_structure(json_dict: dict) -> dict:
+    if isinstance(json_dict, dict):
+        for key, value in json_dict.items():
+            if key == "schema":
+                return value
+            schema = get_schema_from_unknown_structure(value)
+            if schema:
+                return schema
+    return {}
```
Struchid, line 237: Invert the if statement to avoid unnecessary nesting

spinta/manifests/open_api/helpers.py
```diff
@@ -27,74 +32,295 @@ def read_file_data_and_transform_to_json(path: Path) -> dict:
...
+
+
+def get_schema_from_response(response: dict, root: dict) -> dict:
+    json_schema = {}
+    if "openapi" in root:
```
Struchid, line 242: Strings to constants

spinta/manifests/open_api/helpers.py
```diff
@@ -27,74 +32,295 @@ def read_file_data_and_transform_to_json(path: Path) -> dict:
...
+
+    if SCHEMA_REF_KEY in json_schema:
+        raise NotImplementedFeature(feature="Reading OpenAPI with '$ref' structure")
+    return json_schema
+
+def get_model_schemas(dataset_name: str, resource_name: str, response: dict, root: dict) -> list[dict]:
+    if not (json_schema := get_schema_from_response(response, root)):
+        return []
+
+    if json_schema.get("type") == "array":
```
Struchid, line 260: `json_schema.get("type")` re-used, could be extracted to a variable to not be accessed twice

Struchid (approved):

tests/manifests/open_api/test_helpers.py
```diff
@@ -372,3 +373,333 @@ def test_get_dataset_schemas_no_tags_default_dataset():
...
+                }
+            }
+        }
+    ),
+    (RESPONSE, OPENAPI_SCHEMA),
+])
+def test_get_schema_from_response(response, schema):
+    assert get_schema_from_response(response, schema) == SCHEMA
+
+def test_get_schema_from_unknown_format():
```
martynasracct, line 457: This test does nothing

tests/manifests/open_api/test_helpers.py
```diff
@@ -372,3 +373,333 @@ def test_get_dataset_schemas_no_tags_default_dataset():
...
+                    "required": True,
+
+                }
+            },
+        }
+    ]
+
+
+    assert result == expected
+
```
martynasracct, line 554: There should be 2 line separation, I think

martynasracct (approved):
Added few comments. Otherwise, LGTM

PR #1403, opened 2025-07-22
[1378] Introduce sync command

Struchid:
### Overview

`Spinta sync` introduced. This command is responsible for exporting the local DSA from Agent to Catalog.

### Features

- Checking if DSA is already in Catalog
   - If yes — the following workflow is to be implemented and user received an error.
   - if no — the flow continues.
- Dataset is created in Catalog
- Distribution related to the Dataset is created in Catalog
- DSA is uploaded to Catalog following the same logic as the current import in Catalog through the `Import` button:
   - image


spinta/cli/sync.py
Struchid, line 1: Cover w/ tests

spinta/cli/sync.py
```diff
@@ -1,28 +1,144 @@
...
-    config: Config = context.get('config')
-    credentials :RemoteClientCredentials= get_client_credentials(config.credentials_file, DEFAULT_CREDENTIALS_SECTION)
+CONTENT_TYPE_TEXT_CSV = "text/csv"
+IDENTIFIER = "_id"
+
+
+def _get_dataset_name(context: Context, manifest: InlineManifest) -> str:
+    datasets = commands.get_datasets(context, manifest)
+    if len(datasets) > 1:
+        raise NotImplementedFeature(feature="Synchronizing more than 1 dataset at a time")
```
karina-klinkeviciute, line 32: gerai būtų užregistruoti issue ir čia pridėti TODO komentarą su nuoroda į issue

Struchid: Pridėjau: 
- https://github.com/atviriduomenys/spinta/compare/8c92dce0bf1e642bcd9806b694419982484f4383..07811496f76e9624430c3b6f63233b98f90b384b

spinta/cli/sync.py
```diff
@@ -1,28 +1,155 @@
...
-    echo(response.text)
+    if response_get_dataset.status_code not in {HTTPStatus.OK, HTTPStatus.NOT_FOUND}:
+        raise UnexpectedAPIResponse(
+            operation="Get dataset",
+            expected_status_code={HTTPStatus.OK, HTTPStatus.NOT_FOUND},
+            response_status_code=response_get_dataset.status_code,
+            response_data=_format_error_response_data(response_get_dataset.json()),
+        )
+
+    if response_get_dataset.status_code == HTTPStatus.OK:
```
martynasracct, line 90: With late implementations it be a good idea to split `sync` into functions like `def send_dsa_update(...)` and `def create_dsa(...)`. 

Not really needed for now, I think

martynasracct (approved):
Only one optional comment for the future

spinta/cli/sync.py
Struchid: Yeah, I will split it up once it grows, since it would become quite undreadable by the sheer amount of code that will come, but for now — MVP:]

spinta/cli/sync.py
Struchid: Thanks for the comment, added tests :-)

wisecrow (approved):

PR #1428, opened 2025-08-20
[1415] Create data service [#1]

Struchid:
Data service should be created instead of Dataset in the Catalog side during the synchronization.

Related to:
- https://github.com/atviriduomenys/katalogas/pull/1829


spinta/cli/helpers/sync/controllers/distribution.py
```diff
@@ -0,0 +1,50 @@
...
+from spinta.cli.helpers.sync.helpers import validate_api_response
+from spinta.manifests.components import ManifestPath
+
+
+def create_distribution(
+    base_path: str,
+    headers: dict[str, str],
+    dataset_name: str,
+    file_bytes: bytes,
+    dataset_id: str,
```
varzgalys, line 17: I think its an `int`, no?

PR #1456, opened 2025-09-04
[1415] Improvements after the Sprint Review: Part 2

Struchid:
Follow up to:
- https://github.com/atviriduomenys/spinta/pull/1452

---

Remaining points to fix:
4. Codename for Dataset in the Catalog should be exactly as in the DSA. ✅ (No commit since seems to be already doing just that)
5. Dataset name should be generated from title column in the DSA or fallback to the last part of the full name in the dataset column (Separator being the `/`)
6. Visibility=private rows should be completely hidden (Mostly for the resource).
7. DSA should contain the full dataset name datasets/com/cct/DatasetName.

- Dataset name should follow the Title or fallback to the last part of the full dataset name

### DIFF
--- a/spinta/cli/helpers/sync/helpers.py
+++ b/spinta/cli/helpers/sync/helpers.py
@@ -128,7 +128,7 @@ def clean_private_attributes(resource: Resource) -> dict[str, Model]:
     resource_models = resource.models
     for model_name, model in resource_models.items():
         for property_name, property in model.properties.items():
-            if property.basename.startswith("_"):  # Skip
+            if property.basename.startswith("_"):  # Skip private properties.
                 continue
             if property.visibility not in {Visibility.private, None}:
                 continue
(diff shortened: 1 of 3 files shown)


spinta/cli/helpers/sync/helpers.py
```diff
@@ -121,27 +124,25 @@ def get_data_service_name_prefix(credentials: RemoteClientCredentials) -> str:
...
+    resource.models = {name: model for name, model in resource.models.items() if model.visibility != Visibility.private}
 
-        if model.visibility not in {Visibility.private, None}:
-            continue
-        model.external.name = ""  # Source column cleanup.
+    # Property cleaning
+    for model in resource.models.values():
+        model.properties = {
+            name: property for name, property in model.properties.items() if property.visibility != Visibility.private
+        }
```
varzgalys, line 140: According to best practices comments should not say **what** code is doing, but "why". If you think that code is not self explanatory enough - refactor this code part to separate function.

PR #1464, opened 2025-09-11
post release changes

karina-klinkeviciute:
pakeitimai po release:
- CHANGES.rst atnaujinimas
- release proceso pre-release aprašas
- pataisymas, kai buvo neteisingas import
-

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,8 +1,17 @@
 Changes
 #######
 
+0.2dev6 (unreleased)
+====================
 
-0.2dev4 (unreleased)
+- Removed dependency `mypy`
+
+0.2dev5 (2025-09-03)
+====================
+
+- Added dependency `mypy`
+
+0.2dev4 (2025-08-28)
 ====================
 
 New Features:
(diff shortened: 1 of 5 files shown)


CHANGES.rst
```diff
@@ -1,8 +1,17 @@
...
 
+0.2dev6 (unreleased)
+====================
+
+- Removed dependency `mypy`
+
+0.2dev5 (2025-09-03)
+====================
+
+- Added dependency `mypy`
```
Struchid, line 12: Could you expand on this? Is this a `mypy` dependency (something needed to run `mypy`) or the dependency itself is `mypy` ? 

And if it is the second case, why do we want to remove it? I would suggest leaving it be and preparing to introduce it in the future. Unless we want to introduce TY instead of mypy :P
- https://github.com/astral-sh/ty

Struchid (approved):

CHANGES.rst
karina-klinkeviciute: the import from `mypy` was added by mistake, and it traveled to PyPI, together with `mypy` added to pyproject.toml, but later was removed, when we understood that it was added by mistake, so I did a new release just to remove that dependency

PR #1465, opened 2025-09-11
675 update authlib to newest version

JustinasKen:


spinta/datasets/backends/dataframe/backends/soap/components.py
```diff
@@ -16,3 +16,6 @@ class Soap(DaskBackend):
     @contextlib.contextmanager
     def begin(self):
         yield
+
+    def __dask_tokenize__(self):
```
martynasracct, line 20: We should add comment here, explaining why this is needed

martynasracct (approved):

karina-klinkeviciute (approved):

PR #1476, opened 2025-09-15
OAuth scope format addition

Gustascct:
Altered the function `authorized` to create scopes in the UDTS format adjacent to the original scopes. This way both formats work.

Altered the function `get_scope_name` to allow the forming of the UDTS format scopes.

Added testing by expanding on the existing test scenarios that were checking if the original scopes worked. Added new scopes and kept all other parameters the same


spinta/auth.py
```diff
@@ -472,18 +473,24 @@ def get_scope_name(
     elif isinstance(node, Model):
         name = node.model_type()
     elif isinstance(node, Property):
-        name = node.model.model_type() + "_" + node.place
+        name = node.model.model_type() + "_" + node.place if not udts else node.model.model_type() + "/@" + node.place
```
martynasracct, line 476: We should try to avoid using negatives if possible:
```suggestion
        name = node.model.model_type() + "/@" + node.place if udts else node.model.model_type() + "_" + node.place
```

martynasracct (approved):
Few places where `not udts`is used, but feel free to ignore. Otherwise - LGTM

spinta/auth.py
Gustascct: Changed, I made it this way to keep the sequence of choices the same. So it was `name = node.model.model_type() + "_" + node.place` and then I added the new statement at the end. But this is better

spinta/manifests/internal_sql/commands/auth.py
```diff
@@ -74,11 +73,14 @@ def internal_authorized(
...
-
     if not isinstance(action, (list, tuple)):
         action = [action]
-    scopes = [internal_scope_formatter(context, scope, act) for act in action for scope in scopes]
-
+    scopes = [
+        internal_scope_formatter(context, scope, act, udts)
+        for act in action
+        for scope in scopes
+        for udts in [False, True]
```
Struchid, line 82:
```suggestion
        for is_udts in [False, True]
```

Struchid (approved):

spinta/manifests/internal_sql/commands/auth.py
karina-klinkeviciute: > Is the recommendation only for this place, or to change udts -> is_udts everywhere?

I would say - for all places, where it is `boolean`, i.e. where we could answer to the question "is udts" with one of two options - yes (True) or no (False).

Struchid: Yeah, like the comment above, we should get used to using `is_` prefix for booleans, helps massively w/ readability

spinta/auth.py
```diff
@@ -235,13 +236,17 @@ def valid_scope(self, scope, *, operator="AND"):
             return True
 
     def check_scope(self, scope, *, operator="AND"):
+        token_scopes = set(scope_to_list(self._token.get('scope', '')))
+        if any(scope for scope in token_scopes if scope.startswith("spinta_")):
```
Struchid, line 240: **Optional:** The string `spinta_` could be a constant at the top of the file named something like `DEPRECATED_SCOPE_PREFIX` to avoid having a random string laying around.

Struchid (approved):

spinta/auth.py
Gustascct: Thanks, good point

PR #1485, opened 2025-09-22
Ft/oauth scope tests

Gustascct:
.sh files changed to show commands with scopes in the UDTS format.

Tests rewritten to test the UDTS format scope and parametrized.

### DIFF
--- a/tests/test_push.py
+++ b/tests/test_push.py
@@ -108,7 +108,10 @@ def sha1(s):
 
 @pytest.mark.skip("datasets")
 @pytest.mark.parametrize("scopes", [["spinta_set_meta_fields"], ["uapi:/:set_meta_fields"]])
-def test_push_different_models(app, scopes: list,):
+def test_push_different_models(
+    app,
+    scopes: list,
+):
     app.authorize(scopes)
     app.authmodel("country/:dataset/csv/:resource/countries", ["insert"])
     app.authmodel("backends/postgres/report/:dataset/test", ["insert"])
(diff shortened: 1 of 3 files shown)


spinta/auth.py
```diff
@@ -236,7 +237,7 @@ def valid_scope(self, scope, *, operator="AND"):
             return True
 
     def check_scope(self, scope, *, operator="AND"):
```
Struchid, line 239: Let's add typehints to this as well, since you're already adjusting the function

tests/commands/source/test_sql.py
```diff
@@ -18,7 +18,8 @@ def sql(rc):
 
 
 @pytest.mark.skip("datasets")
-def test_sql(rc, cli, sql, app):
+@pytest.mark.parametrize("scope", [["spinta_getall", "spinta_search"], ["uapi:/:getall", "uapi:/:search"]])
```
Struchid, line 21: Maybe `scopes` would be a better name? It is a collection of scopes and not a single scope

Struchid (approved):

tests/commands/source/test_sql.py
Gustascct: In some places yes, it is a collection but in some places only one scope is assigned. I could change it to scopes, but it would be quite time consuming now

spinta/auth.py
Gustascct: After rebase the function is changed and typehints are there now

PR #1502, opened 2025-09-26
1468 SOAP decode base64 string

martynasracct:
**Summary:**
* Rewrites existing `base64()` prepare function (`base64 -> bytes`) to use python code for decoding `base64` instead of dask dataframes. This makes function slightly slower (difference in milliseconds for ~1000 records), but shows errors for invalid base64 strings
* Introduce `base64()` prepare function that decodes to string (`base64 -> string`) 

**Ticket:**
https://github.com/atviriduomenys/spinta/issues/1486

### DIFF
--- a/spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
+++ b/spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
@@ -10,12 +10,11 @@ from spinta.datasets.backends.dataframe.ufuncs.query.components import (
     DaskSelected as Selected,
 )
 from spinta.exceptions import PropertyNotFound, NotImplementedFeature, SourceCannotBeList
-from spinta.types.datatype import DataType, PrimaryKey, Ref, Binary
+from spinta.types.datatype import DataType, PrimaryKey, Ref
 from spinta.types.text.components import Text
 from spinta.ufuncs.components import ForeignProperty
 from spinta.utils.data import take
 from spinta.utils.schema import NA
-import base64 as b64
 
 
 @ufunc.resolver(DaskDataFrameQueryBuilder, Expr, name="and")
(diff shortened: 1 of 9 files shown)


tests/ufuncs/resultsbuilder/test_ufuncs.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+        model = commands.get_model(context, manifest, "example/Data")
+        env = ResultBuilder(context).init(
+            this="test",
+            prop=model.properties["value"],
+            data={"value": "dGVzdA=="},
+            params={},
+        )
+
+        with pytest.raises(InvalidBase64String):
```
Struchid, line 51: Could we also assert the error messages where `pytest.raises` is used in tests? That way we would ensure, that correct "values" are passed to the exception "template". Relevant for the two other test files as well if you decide to do this.

Struchid (approved):
🚀🕺

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
```diff
@@ -414,25 +413,9 @@ def select(
...
-    item = f"base64({dtype.prop.external.name})"
-    env.dataframe[item] = (
-        env.dataframe[dtype.prop.external.name]
-        .str.encode("ascii")
-        .map(b64.decodebytes, meta=(dtype.prop.external.name, bytes))
-    )
-    return Selected(item=item, prop=dtype.prop)
+@ufunc.resolver(DaskDataFrameQueryBuilder, Expr)
+def base64(env: DaskDataFrameQueryBuilder, expr: Expr) -> Expr:
+    return expr  # Expression will be resolved later
```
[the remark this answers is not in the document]
martynasracct: Changed comment to `# Expression will be resolved in ResultBuilder`. It will be clearer.

tests/ufuncs/resultsbuilder/test_ufuncs.py
martynasracct: Done

PR #1504, opened 2025-09-26
1424 wsdl backend error during spinta check

gts446:
Moved `soap_operation` validation/assigment from link builder to separate Soap backend method which is called only when retrieving data.

### DIFF
--- a/spinta/datasets/backends/dataframe/backends/soap/components.py
+++ b/spinta/datasets/backends/dataframe/backends/soap/components.py
@@ -22,10 +22,7 @@ class Soap(DaskBackend):
 
     @property
     def soap_operation(self) -> OperationProxy:
-        if not self._soap_operation:
-            self._soap_operation = self._get_soap_operation()
-
-        return self._soap_operation
+        return self._soap_operation if self._soap_operation else self._get_soap_operation()
 
     def _get_soap_operation(self) -> OperationProxy:
         with self.wsdl_backend.begin():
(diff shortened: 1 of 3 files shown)


spinta/datasets/backends/dataframe/backends/soap/components.py
```diff
@@ -4,14 +4,44 @@
...
+    service: str
+    port: str
+    operation: str
+
+    @property
+    def soap_operation(self) -> OperationProxy:
+        if not self._soap_operation:
+            self._soap_operation = self._get_soap_operation()
+
+        return self._soap_operation
```
Struchid, line 28: Shorthand could be used:
```suggestion
        return self._soap_operation if self._soap_operation else self._get_soap_operation()
```

Struchid (approved):

tests/cli/test_check.py
```diff
@@ -441,3 +441,24 @@ def test_check_nested_ref_and_prepare(context: Context, rc, cli: SpintaCliRunner
...
         ],
     )
+
+
+def test_check_wsdl_unreachable(context: Context, rc, cli: SpintaCliRunner, tmp_path):
+    create_tabular_manifest(
+        context,
+        tmp_path / "manifest.csv",
+        striptable("""
+d | r | b | m | property        | type    | ref                              | prepare            | source
```
martynasracct, line 451: We should format this DSA better. ref column is also not used, so not needed

PR #1511, opened 2025-10-01
1338 soap with params fixes

gts446:
- Removed _base column when getting data with SOAP requests with parameters.
- Refactored soap_query_builder to not include params without values in request.
- Raising an exception when trying to send SOAP request with missing required parameters.

### DIFF
--- a/spinta/datasets/backends/dataframe/backends/soap/ufuncs/ufuncs.py
+++ b/spinta/datasets/backends/dataframe/backends/soap/ufuncs/ufuncs.py
@@ -107,6 +107,7 @@ def soap_request_body(env: SoapQueryBuilder, prop: Property, param: Param) -> No
 
     if final_value is NA:
         env.soap_request_body.pop(param_body_key, None)
+        final_value = None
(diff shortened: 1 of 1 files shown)


spinta/formats/html/commands.py
```diff
@@ -63,10 +63,8 @@
 
 
 def _get_model_reserved_props(action: Action, include_page: bool) -> List[str]:
-    if action == Action.GETALL:
+    if action in [Action.GETALL, Action.SEARCH]:
```
martynasracct, line 66: unit test is missing for this

martynasracct (approved):
LGTM! I didn't expect fix for required property will take only 2 lines!

Struchid (approved):

martynasracct (approved):
CHANGES.rst formatted incorrectly

spinta/formats/html/commands.py
gts446: decided test not needed, because this is for html only.

PR #1514, opened 2025-10-01
1506 cache control

JustinasKen:


spinta/config.py
```diff
@@ -259,6 +259,8 @@
     "max_file_size": 100,
     # Used to determine max amount of errors can be thrown while writing, before canceling writing stream
     "max_error_count_on_insert": 100,
+    # Response Cache-Control max-age value in seconds.
+    "cache_control_max_age": 60,
```
karina-klinkeviciute, line 263: as I understand, the default value can be set in a separate config file? Here it's quite low, for our current needs, I'd say?

karina-klinkeviciute (approved):

spinta/config.py
JustinasKen: Well this parameter controls amount of time needs to pass for nginx to recheck if value has changed, so idealy I would set it to 0 (but then nginx doesn't properly cache items).

Setting it high would mean, that we need to wait at least that amount of time for new changes to appear, so `insert`, `update`, `changelog` would not be up to date and there could be cases where sync problems would arise (especially with high `max_age` parameter). 

Lower `max_age` only means that the data gets "expired" faster prompting server to either return the new data or return `304` (meaning no changes happened and we can reuse the old cached data again).

karina-klinkeviciute: Thanks for the answer. So this is not when the data is resent, but when we check for changes, right?

JustinasKen: Well the lifecycle of this concept is something like this:

Request Data -> Nginx -> Response
or
Request Data -> Nginx -> Server -> Nginx -> Response

So at first step, nginx checks if it has cached data if yes then it checks if its expired, if no it returns it, if yes then it asks the remote server.

Remote server checks the headers, which contain instruction of latest cached data, it then validates it to check if it matches or not, if it does, server returns `304` response (`Not Modified`), which makes `nginx` reset the timer of cached data and return it.

If the given data does not match with what is latest, remote server will return `200` response with the new data, which gets recached inside nginx again.

The expiration parameter gets set with `proxy_cache_valid 200 301 302 1m;` parameter, but since we pass `max-age` inside `Cache-Control` header, it actually overwrites it.

karina-klinkeviciute: thanks for the explanation

PR #1520, opened 2025-10-02
copy resources without models during spinta copy

gts446:


martynasracct (approved):
We should also add ticket description, briefly explaining what has been done

PR #1585, opened 2025-10-31
1569: support multiple jwks, rotation. New GET jwks API (auth-server)

varzgalys:
### 1. New spinta (as auth server) GET API to retrieve JWKs (public keys to verify jwt tokens)

`GET https://get-test.data.gov.lt/.well-known/jwks.json` gives response like:

```
{
  "keys": [
    {
      "kid": "rotation-1",
      "kty": "RSA",
      "alg": "RS512",
      "use": "sig",
      "n": "oAXjeXtZxiEUI7EcG6uITGCuUHmMQxMdTuSkQMaijmX0R1xSN-sQOgrunTqzldGWYhn4CQXmE34TgoZs2l6pZKNEyzap5IstPAUTFfHamyLka-xBwVRpCJaM_ZY9dEhzn9NUB-mx1ud9_clhmlef0SRQ1E5N_oU9wA_Hgd6hdnRzzTDJzmueF_03fEEf27fd69qzPZerOO7E9ytHJm0RpTF-50MGDL9pJaomAry_m0cw66DRd8rwqE-MiSg1xo02YWYIbaNA13K7jO33lW3iqgLdmtiBvX7qoNhEXC5H_umLvd5hgETGVemFcdFgL0Xnj85uk3puiVMsYXqmzHNxdw",
      "e": "AQAB"
    },
    {
      "kid": "rotation-2",
      "kty": "RSA",
      "alg": "RS512",
      "use": "sig",
      "n": "1NYpTUTJlKS3t0OzhsPIzUS91IzV7hZEloQ_EsLXFReNDb4nm16Hnp1ohWsBnWHivxu0oCM2V3Vjw-8xnkopzgRZyDJujvaK9mwklapPAjY9OdmxRhDTpdD9wJf0DiAj-AxTo6RBiTuUM5k1A95BjI0ZXO9QQjEivNnoBNJ121-tmjDgx63pTsmkhwFQxkXjBN9TyKmeXOOPVadazpeLqUUEfCuYjZ1Mb9u8oBeBlGsGYrQLmqqWF3vZBcEJcxTc0w9qVQxbCqlyFhA5qKG_mm_K09WrtzzU3_u_EqMm9GD9ubcMKG6XCYInn6jA2tJmy_niY5DL0DUYipcLByYEGw",
      "e": "AQAB"
    },
    {
      "kty": "RSA",
      "n": "ngg7HGoRkBDkhLFZpFIF5qOSnWPt7FoThHpP5-HOeVZzrM2NlVhmn93Rl32o83jJ3GCP_l4_ExPJsRCOwJb0trMUH9hLwwkpe63ENJs3y8yH75GrGkIyVtS4sKYtJpNOWgyFQoVheRN28KV3TQxSySspooFYkNa0Z_jPlP5lA5PXS_uqYNleBNPi7zw4StOze6AqIq3F7MDQQe5lQUUnxI1ORrW2PgVPe0yKjQIxslxTBuW7spHzaU1RFlfzla8eP5k0nFEMtk8LVtHonpK7W_UDa4DgYIOzmQPJH5fwXDSUZspQaUpvkAd8SCKhcJ4sRwn9FFQu1_hHwRt-Lx5UyQ",
      "e": "AQAB"
    }
  ],
}
```

### 2. Spinta (as agent) now is able to work with multiple JWKs and pick dynamically correct for different keys (use kid header claim). JWK also can be retrieved dynamically (instead of env vars) from url - `token_validation_keys_download_url` new config value for that.
Scenario when multiple jwks exists - some auth server, for instance - gravitee, periodically rotate their public keys, which are used to encrypt JWT access tokens (signature). But after the rotation - old access tokens are still used (as their expiration is still valid) and need to be verified, but conditionally - new access token with new public key, old with the old one. All this mapping is passed on a `kid` (stand for key id) claim field both inside of a public key and access token. Sometimes (for instance our current setup) `kid` field does not exists.

Vadovas PR - https://github.com/atviriduomenys/vadovas/pull/53

### DIFF
--- a/spinta/auth.py
+++ b/spinta/auth.py
@@ -13,14 +13,15 @@ import uuid
 from collections import defaultdict
 from functools import cached_property
 from threading import Lock
-from typing import Set, Any
+from typing import Set, Any, TypedDict, Literal
 from typing import Type
 from typing import Union, List, Tuple
 
+import requests
 import ruamel.yaml
-from authlib.jose import JsonWebKey
+from authlib.jose import JsonWebKey, RSAKey, JWTClaims
 from authlib.jose import jwt
-from authlib.jose.errors import JoseError
+from authlib.jose.errors import JoseError, DecodeError, InvalidTokenError, BadSignatureError
 from authlib.oauth2 import OAuth2Error
 from authlib.oauth2 import OAuth2Request
 from authlib.oauth2 import rfc6749
@@ -34,6 +35,7 @@ from cachetools.keys import hashkey
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.asymmetric import rsa
 from multipledispatch import dispatch
+from requests import RequestException
 from starlette.datastructures import FormData, QueryParams, Headers
 from starlette.exceptions import HTTPException
 from starlette.responses import JSONResponse
@@ -77,6 +79,20 @@ DEPRECATED_SCOPE_PREFIX = "spinta_"
 
 # Scope types taken from authlib.oauth2.rfc6749.util.scope_to_list
 SCOPE_TYPE = Union[tuple, list, set, str, None]
+Kid = str  # key id
+
+
+class JWK(TypedDict):
+    kid: str
+    kty: str
+    alg: str
+    use: str
+    n: str
+    e: str
+
+
+class JWKS(TypedDict):
+    keys: List[JWK]
 
 
 class KeyType(enum.Enum):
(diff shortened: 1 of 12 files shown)


martynasracct (approved):

spinta/auth.py
```diff
@@ -166,10 +181,57 @@ def __init__(
...
+                for key in token_validation_key["keys"]:
+                    if kid := key.get("kid"):
+                        local_public_keys[kid] = JsonWebKey.import_key(key)
+            elif kid := token_validation_key.get("kid"):
+                local_public_keys = {kid: JsonWebKey.import_key(token_validation_key)}
+        return self.download_public_keys(context) | local_public_keys
+
+    @staticmethod
+    @cached(downloaded_public_keys_cache)
+    def download_public_keys(context) -> dict[Kid, RSAKey]:
```
Struchid: Typehints are missing

spinta/auth.py
```diff
@@ -166,10 +181,57 @@ def __init__(
...
+    def load_all_public_keys(self, context) -> dict[Kid, RSAKey]:
+        config = context.get("config")
+        token_validation_key = config.token_validation_key
+
+        local_public_keys: dict[Kid, RSAKey] = {}
+
+        if isinstance(token_validation_key, dict):
+            if "keys" in token_validation_key:
+                for key in token_validation_key["keys"]:
+                    if kid := key.get("kid"):
```
Struchid: Could we use full names instead of `kid`?

spinta/auth.py
```diff
@@ -166,10 +181,57 @@ def __init__(
...
+        keys: list[JWK] = jwks["keys"]
+
+        kid_key_map: dict[Kid, RSAKey] = {}
+        for jwk in keys:
+            if kid := jwk.get("kid"):
+                kid_key_map[kid] = JsonWebKey.import_key(jwk)
+        return kid_key_map
+
+    def get_key(self, kid: Kid | None = None) -> RSAKey:
+        if kid:
```
Struchid: Conditional could be inverted to avoid the whole method indentation

Struchid (approved):

PR #1586, opened 2025-11-03
Ft/1567 param.header prepare function

martynasracct:
**Summary:**
Adds `param.header` prepare function for SOAP backend. It reads headers from DSA. Headers then are passed to Zeep client and are attached to SOAP request.

**Ticket:**
https://github.com/atviriduomenys/spinta/issues/1576

### DIFF
--- a/spinta/auth.py
+++ b/spinta/auth.py
@@ -826,7 +826,7 @@ def delete_client_file(path: pathlib.Path, client_id: str):
                 Remove only empty folders
             """
     else:
-        raise (InvalidClientError(description="Invalid client id or secret"))
+        raise InvalidClientError(description="Invalid client id or secret")
 
 
 def update_client_file(
@@ -876,7 +876,7 @@ def update_client_file(
                 yml.dump(keymap, keymap_path)
         return new_data
     else:
-        raise (InvalidClientError(description="Invalid client id or secret"))
+        raise InvalidClientError(description="Invalid client id or secret")
 
 
 def get_client_id_from_name(path: pathlib.Path, client_name: str):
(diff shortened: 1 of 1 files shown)


spinta/ufuncs/loadbuilder/ufuncs.py
```diff
@@ -70,3 +70,22 @@ def input_(env: LoadBuilder, expr: Expr) -> None:
...
     env.param.soap_body = {env.this: prep_value}
+
+
+@ufunc.resolver(LoadBuilder, Expr)
+def header(env: LoadBuilder, expr: Expr) -> None:
+    args, kwargs = expr.resolve(env)
+    if len(args) != 1:
+        raise InvalidArgumentInExpression(arguments=args, expr="header")
+
+    return env.call("header", env.this, args[0])
```
varzgalys: Confusing `return` (if return typehint is correct)
```suggestion
    env.call("header", env.this, args[0])
```

varzgalys (approved):

spinta/ufuncs/loadbuilder/ufuncs.py
martynasracct: This call always returns None. Adding `return` makes it more explicit.

spinta/auth.py
```diff
@@ -988,7 +988,7 @@ def query_client(path: pathlib.Path, client: str, is_name: bool = False) -> Clie
     try:
         data = yaml.load(client_file)
     except FileNotFoundError:
-        raise (InvalidClientError(description="Invalid client id or secret"))
+        raise (InvalidClientError(description="Client file not found. Invalid client id or secret"))
```
Struchid: Is this needed?

```suggestion
        raise InvalidClientError(description="Client file not found. Invalid client id or secret")
```

Struchid (approved):

PR #1591, opened 2025-11-06
1582 Small refactoring on test_api.py

martynasracct:
**Summary:**
No bussiness logic changes. Only test refactoring.

- Introduce create_client() function that creates client file instead of calling `auth/client` for test setup
- Move `auth/client` and `auth/client/<client_id>` URLs to functions

**Ticket:**
https://github.com/atviriduomenys/spinta/issues/1582

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import pathlib
 import uuid
 from typing import Any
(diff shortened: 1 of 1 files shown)


tests/test_api.py
```diff
@@ -30,6 +31,35 @@
 from spinta.utils.types import is_str_uuid
 
 
+def clients_url() -> str:
+    return "/auth/clients"
+
+
+def client_url(client_id: str) -> str:
+    return f"/auth/clients/{client_id}"
```
Struchid: **Nitpick:** Could re-use the logic from clients_url (as a fixture) if later down the line you need to replace the url there, you will not need to do it here as well

PR #1605, opened 2025-11-13
1574: RedisKeymap - do not create entries when no value (move action)

varzgalys:
RedisKeymap - do not create entries when no value (move action)
Fix redis start command to persist always, change docs also.


Gustascct (approved):

spinta/datasets/keymaps/redis.py
Struchid: Tests are missing for the change

wisecrow (approved):

PR #1623, opened 2025-11-25
WSDL/SOAP requests use CDATA for XML elements

martynasracct:
**Summary:**
Adds `cdata()` prepare function that can be used on params. This function adds `<![CDATA[...]]>` to parameter's value so that value can be sent as is. Without automatic XML symbol escaping, that Zeep client automatically does.

With this, it's possible to send unescaped XML via Zeep client

Commit a1ff46be6a3dd5a98b30f7fb95f6a74aad2de2be is just refactoring. It can be checked separately from other two commits

**Ticket:**
https://github.com/atviriduomenys/spinta/issues/1625

### DIFF
--- a/spinta/datasets/backends/dataframe/backends/soap/ufuncs/ufuncs.py
+++ b/spinta/datasets/backends/dataframe/backends/soap/ufuncs/ufuncs.py
@@ -134,7 +134,7 @@ def soap_request_body(env: SoapQueryBuilder, prop: Property, param: Param) -> No
     final_value = _get_final_soap_request_body_value(env, prop.place, param_source)
 
     if final_value is NA:
-        # If value not in URL and DSA have no default - remove it from SOAP request completely
+        # If value not in URL and DSA has no default - remove it from SOAP request completely
         env.soap_request_body.pop(param_source, None)
         final_value = None
     elif final_value:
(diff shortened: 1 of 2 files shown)


wisecrow (approved):

spinta/datasets/backends/dataframe/backends/soap/ufuncs/ufuncs.py
```diff
@@ -102,23 +119,34 @@ def soap_request_body(env: SoapQueryBuilder, prop: Property) -> None:
...
-    url_param_value = env.query_params.url_params.get(prop.place, NA)
-    param_body_key = next(iter(param.soap_body))
-    default_value = env.soap_request_body.get(param_body_key, NA)
-    final_value = url_param_value if url_param_value is not NA else default_value
+    param_source = next(iter(param.soap_body))
+    final_value = _get_final_soap_request_body_value(env, prop.place, param_source)
 
     if final_value is NA:
-        env.soap_request_body.pop(param_body_key, None)
+        # If value not in URL and DSA have no default - remove it from SOAP request completely
```
Struchid:
```suggestion
        # If value not in URL and DSA has no default - remove it from SOAP request completely
```

spinta/ufuncs/loadbuilder/ufuncs.py
```diff
@@ -64,6 +64,12 @@ def or_(env, expr):
     return args
 
 
+@ufunc.resolver(LoadBuilder, Expr)
+def cdata(env: LoadBuilder, expr: Expr):
```
Struchid:
```suggestion
def cdata(env: LoadBuilder, expr: Expr) -> None:
```

PR #1648, opened 2025-12-17
ft/1598 check contract scopes

martynasracct:
**Summary:**
Part2 of https://github.com/atviriduomenys/spinta/issues/1598. First part: https://github.com/atviriduomenys/spinta/pull/1634

For each request does additional check if JWT token has any extra scopes (of currently loaded manifest models) that are not in client's file `contract_scopes` varialbe. If there are - raise error. 

This check is only done if `Config.check_contract_scopes` is enabled and requested data access is not `open`.

**Description:**
This check is done for zero-trust policy. We are validating that the authentication server sent us correct scopes by comparing them with scopes saved in client file's `contract_scopes` variable. For now, `contract_scopes` are populated manually. Later it will be populated during sync with Katalogas.

How it works:
1. Collects all currently loaded manifest models and get all available namespaces from it
 *Example: model `datasets/gov/Country` will result into `{"datasets", "datasets/gov", "datasets/gov/Country"}`* 
2. Filters all JWT scopes and only selects scopes that matches set from step 1. Spinta scope prefixes and scope actions are not taken into account.
  *Example: `uapi:/datasets/gov/Country/:getall` -> `datasets/gov/Country` -> compare with step 1 result*
3. Compares filtered scopes from step 2 with `contract_scopes` saved in client file. Raises error if there are 0 filtered scopes or if any of filtered scopes are not in `contract_scopes`

**Ticket:**
https://github.com/atviriduomenys/spinta/issues/1598

### DIFF
--- a/spinta/core/enums.py
+++ b/spinta/core/enums.py
@@ -128,7 +128,7 @@ class Action(enum.Enum):
         return list(cls._value2member_map_.keys())
 
     @classmethod
-    def scope_action_values(cls):
+    def scope_action_values(cls) -> list[str]:
         old_prefix_actions = list(map(lambda action: "_" + action, cls.values()))
         new_prefix_actions = list(map(lambda action: "/:" + action, cls.values()))
         return old_prefix_actions + new_prefix_actions


spinta/auth.py
```diff
@@ -464,6 +469,76 @@ def is_revoked(self) -> bool:
...
+        scope_namespace_map = self._get_namespace_scope_map(scope_prefixes)
+
+        return {
+            scope
+            for namespace, scopes in scope_namespace_map.items()
+            if namespace in model_namespaces
+            for scope in scopes
+        }
+
+    def check_contract_scopes(self, context: Context) -> None:
```
martynasracct: I am not sure if `Token` class is the best place to put this logic. If you have better ideas - please suggest!

Struchid (approved):

Issue #1641, opened 2025-12-15
Fix URL encoding


sirex:
**Specifikacijos punktas:**
- 6.3.3. nenumatyti reikalavimai
#

While running spinta with `spinta run` and making specific requests, which require encoding of spaces, we run into this issue:

Different clients, differently encodes URL query parameters, here are a few examples:

```
original:                '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
curl --data-urlencode:   %27%3c%3fxml+version%3d%221.0%22+encoding%3d%22UTF-8%22+standalone%3d%22yes%22%3f%3e%3cargs%3e%3cfmt%3exml%3c%2ffmt%3e%3cdata%3e2025-12-08%3c%2fdata%3e%3c%2fargs%3e%27
httpie param==value:     %27%3C%3Fxml+version%3D%221.0%22+encoding%3D%22UTF-8%22+standalone%3D%22yes%22%3F%3E%3Cargs%3E%3Cfmt%3Exml%3C%2Ffmt%3E%3Cdata%3E2025-12-08%3C%2Fdata%3E%3C%2Fargs%3E%27
httpie param=value:      '%3C?xml%20version=%221.0%22%20encoding=%22UTF-8%22%20standalone=%22yes%22?%3E%3Cargs%3E%3Cfmt%3Exml%3C/fmt%3E%3Cdata%3E2025-12-08%3C/data%3E%3C/args%3E'
chromium:                %27%3C?xml%20version=%221.0%22%20encoding=%22UTF-8%22%20standalone=%22yes%22?%3E%3Cargs%3E%3Cfmt%3Exml%3C/fmt%3E%3Cdata%3E2025-12-08%3C/data%3E%3C/args%3E%27
```

Probably the main issue, that some clients encode spaces as `%20` and others as `+`.

Spinta understand `%20`, but leaves `+` unchanged, and this way corrupts data, because we get `+` instead of space.

So we need to handle `+` encoded spaces too.

I think, I found the places, where `+` is left unhandled:

https://github.com/atviriduomenys/spinta/blob/c2ef133ddd52d4955727dc78405dc799cb6258c1/spinta/urlparams.py#L34

The `unquote` should be changed to [unquote_plus](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.unquote_plus).

## Related

- https://github.com/atviriduomenys/spinta/issues/136

PR #1649, opened 2025-12-17
changed unquote to unquote_plus to handle + as space

karina-klinkeviciute:
**Summary:**
In some cases (for example, using cli clients to make HTTP requests) spaces in URL query parameters are encoded differently: sometimes using `%20`, other times using `+`. Spinta currently cannot parse `+` in query parameters.

This PR fixes URL parser to also correctly parse `+` into space.

Side effect of this:
`mongo` and `postgres` backends has feature to sort data asc/desc by using +/- sign. For example: `?sort(+property)`. URL parser now removes replaces `+` with space and we get query `?sort(property)`. Spinta already sorts this ascending.

I had to implement same logic to nested properties: `?sort(+model.property)`.

**Ticket:**
https://github.com/atviriduomenys/spinta/issues/1641

### DIFF
--- a/spinta/backends/mongo/ufuncs/ufuncs.py
+++ b/spinta/backends/mongo/ufuncs/ufuncs.py
@@ -484,7 +484,7 @@ def sort(env, field):
 
 
 @ufunc.resolver(MongoQueryBuilder, DataType)
-def sort(env, dtype):
+def sort(env: MongoQueryBuilder, dtype: DataType) -> Any:
     return env.call("asc", dtype)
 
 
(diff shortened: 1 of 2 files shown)


wisecrow (approved):
entry to CHANGES.rst is missing. LGTM

spinta/backends/mongo/ufuncs/ufuncs.py
```diff
@@ -483,6 +483,11 @@ def sort(env, field):
     return env.call("asc", prop.dtype)
 
 
+@ufunc.resolver(MongoQueryBuilder, DataType)
+def sort(env, dtype):
```
Struchid: Typehints are missing, I think we can ignore that other methods in this file do not have them and start introducing them with new functionalities :)

Struchid (approved):

PR #1672, opened 2026-01-13
Prototype solution for XML DSA language tag #1658

### LINKED ISSUE #1658: Gaunant duomenis iš spintos, nesuprantamas @ tagas
Bandant gauti duomenis iš spintos, kai manifeste property pavadinimuose naudojami kalbos tag'ai, gaunama klaida

Klaida kilo bandant gauti duomenis iš XML endpointo. (RC JAR klasifikatoriai)

Pavyzdžiui: `title@lt`

tuturis:
Prototype for solution:
Goal is to change malfunctioning function `dask_get_all` calls into multi-dispatch command to transform model data and deal with the issue described in the parent ticket.

This will help to keep the spinta codebase from spinta design antipatterns, help structure tests, improve readability and reduce cognitive load while testing the DSA coverage for each backend, as well as add the missing functionality the original ticket is about.

Currently as-is state generalized data transformation flow state can be depicted within the next sequence with swim lanes.
1674-as-is

Proposed to-be state with where generalized multi dispatch command stream_object_model is introduced, following the best practices of spinta and its design patterns.
1674-to-be

The reason for this minimal change in flow is that dask_get_all is greatest common denominator for the behaviour mentioned in #1658 across various backends.

This Pull Request covers:
- Add tests coverage for #1658 issue for XML reading with language selection and search parameters.
Goal is to review the scope.
- Adds command.stream_object_model.
- Adds experimental proof of concept (TRL 2-3) handling for the dask/xml backend, with Spinta selectors, DSA Meta-types (Ref, PK) handling using the command.stream_object_model.

In-progress:
- Add extensive regression test suites.  Status ~1% Done
- Replace the rest of `dask_get_all`. instances with new command. 1/5 Done

High level overview of the to-be design 

The main change is introduced with the introduction of a new `command.stream_object_data` generator function that yields object data replacing the `dask_get_all` function call. 
`command.stream_object_data` uses multi dispatch signature, to load the model, transform it, load the data, resolve the metadata and stream the data.
In example - the to-be state, with the code that is defined here provides overloads for `command.stream_object_data`.
At its core - the stream_object_data orchestrates the loop of loading Spintas logical model and transform it into intermediate representation to enable data and metadata loading, response generation.
Currently it's done using adapters for the (UDTS)model to normalized Row (model), where it flattens the tree, constructs paths and constructs each row. Then data adapter constructs masks from given propeties, selects the correct columns and uses injected Dask dataframe, queries it, returning the collum and its values. Row metada adapter loads the Ref and Primary Key types.


karina-klinkeviciute:
> should this issue scope also handle the issue of multiple languages? ("/example/xml/City?select(code,name@lt,name@en)")

Yes

spinta/commands/__init__.py
```diff
@@ -591,6 +592,30 @@ def getfile():
...
+    metadata_adapter: DataAdapter,
+    *,
+    transformation_adapter: Callable[[TransformationModel, dict[str, object]], T],
+) -> Iterable[T]:
+    """Stream model data from the source."""
+    pass
+
+
+@command()
+def stream_model_data(*args, **kwargs):
```
JustinasKen: Is new command necessary, doesnt `getall` already encompass this idea ?

We should avoid adding new commands if we can, since there already is too much command bloat.

spinta/datasets/backends/dataframe/backends/xml/commands/read.py
```diff
@@ -9,20 +9,72 @@
...
+                            lang = arg['args'][1]['args'][0]
+                            candidate_prop = f"{arg['args'][0]['args'][0]}@{lang}"
+                            if candidate_prop in props.keys():
+                                prop_def = model.properties.get(prop)
+                                if (
+                                    prop_def and
+                                    isinstance(prop_def.dtype, Text) and
+                                    lang in prop_def.dtype.langs
+                                ):
+                                    manifest_paths.append(str(candidate_prop))
```
JustinasKen: `Expr` handling should not be hardcoded, `Env` should be used to properly parse them. Most of the time, this kind of handling would be considered a hack, since it's hard to maintain.

Also this kind of if looks like its hardcoding specific case, `getattr` can be infinitely nested.

Also `PropertyResolver` exists, which is used to resolve binds, getattr and other expression compontents into either ufunc properties or model properties.

https://github.com/atviriduomenys/spinta/blob/0c6427bcc4b16831fadd39ca5583f7943fd3a405/spinta/ufuncs/propertyresolver/components.py#L5-L16

spinta/datasets/backends/dataframe/backends/xml/commands/read.py
```diff
@@ -115,6 +167,39 @@ def _gather_namespaces_from_model(context: Context, model: Model) -> dict:
...
+        right = query_dict['args'][1]
+        if left['name'] == 'bind' and left['args'][0] == '_id':
+            pkey_props = [pkey.name for pkey in model.external.pkeys]
+            if len(pkey_props) == 1:
+                decoded = keymap.decode(pkey_props[0], right)
+                prop = model.properties[pkey_props[0]]
+                name = prop.external.name
+                df_mask[name] = {"op": 'eq', "value": decoded}
+            else:
+                pass
```
JustinasKen: Same deal as above, `Expr` should not be handled through `todict()` and should be left to `Env`.

Also this again is hardcoded case, since it does not take into consideration nesting and function mangling. 

By quick glance it looks like this will not handle queries like `and(country="Lithuania", and(city="Vilnius", age>18))`.

This is one of the main reasons why `QueryBuilder` was used, since functions can be nested infinitely and we do not want to define hardcoded limits. Only limits that exist with `QueryBuilder` is if specific functions are not properly implemented to handler specific datatypes.

JustinasKen (changes requested):

spinta/adapters/loaders.py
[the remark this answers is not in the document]
tuturis: Good question. Note - The merge request is still draft.
To answer the question, it uses multi dispatch with overload pattern so in theory you could use your own approach.
On the other hand having a multi dispatch single interface allows to tests to be more streamlined.

spinta/commands/__init__.py
tuturis: The alternative is less testable.

spinta/datasets/backends/dataframe/backends/xml/commands/read.py
tuturis: We can add this to the scope 👍

karina-klinkeviciute:
For how much time would you roughly estimate this refactor, provided that it covers all the issues addressed in the comments, including all the tests etc.?

spinta/datasets/backends/dataframe/backends/xml/commands/read.py
```diff
@@ -156,4 +248,12 @@ def getall(
...
-    yield from dask_get_all(context, query, df, backend, model, builder, extra_properties)
+
+    yield from commands.stream_model_data(
+        model,
+        Row(manifest_paths=manifest_paths,
+            context=context),
+        DaskXml(df=df, df_mask=df_mask),
+        RowMetaItem(key_map=keymap),
+        transformation_adapter=RowFormatter.to_object_data
+    )
```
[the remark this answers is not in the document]
JustinasKen: > > > the cast to backend is still called. The test are source of truth. Again about concers - very subjective. Maybe you could provide some objective metrics? I propose code coverage for example.
> > > As for all other backends (see diagrams above). The common greatest denominator where the problem comes from and a design antipattern is dask_get_all.
> > 
> > 
> > Well if you could, please provide line of code that calls `cast_backend_to_python`, I couldn't seem to find it, maybe I missed it, since it was included in `dask_get_all`.
> > I tried to use your branch to test it, but `format` feature no longer properly works there, I cannot see `html` response for `geometry` datatype. `html` `prepare_dtype_for_response` for `Geometry` datatype expects `BaseGeometry` from `shapely` library, it uses it to create openstreetmap link to center point of geometric data. If `cast_backent_to_python` actually is used, then with html format, when using `Geometry` type in manifest and passing string data like `POINT (10 20)` should create the link to map, but if it only returns `POINT (10 20)` as string in `html` format, then it most likely is not used.
> > Also could you explain what do you mean by `dask_get_all` antipattern. This functions is just a wrapper for reusability, if you look at other backend implementations, you will see that they are very simillar.
> > Another thing, I do agree that objective metrics are important, but code coverage is not the best metric to make sure that system is actually properly tested, it is a good guardrail, but for this kind of refactor its important to add as many different edge case tests, where system actually gets limit tested, where you combine multiple filters, nest them, in my opinion most effective tests are the ones that try to break the feature by combining features together, that on paper should work together. Now what I said is harder to qunatify, since you need to think about the feature from a users perspective.
> > From my experience working on this project for almost 3 years I have seen several times when people create tests, that for example test 1 depth nesting and it works, but then you look at the code and its just hardcoded, so once you change it 6 depth nesting everything breaks. And code coverage does not really show these kind of cases.
> 
> Good points, if you could provide AAA pattern unit tests for the cases you described we could talk. As mentioned earlier. It's a draft. image We can discuss the TDD - for the coverage (all dask_get_all,) and by using DDD we can formally ensure there are no edge cases. The dask_get_all is antipattern because its
> 
> * Not unit testable.
> * Does not follow established command patterns (multi dispatch).
> * Its greatest common denominator for the issue that needs to be refactored as stated in the diagrams.
> 
> About nesting if we talk about Ref data types, the referenced prototype code deals with them with no hurdles.

I am aware it is a draft, for that reason I did not comment on other parts that I did not find correct, I only commented on parts, that showed lack of understanding of how and why the current system works. For example use of `Env` for `Expr` handling and usage of if checks for specific cases (it shows that you do not think about `Expr` as a feature, that it was intended to be (`AST`)). These things raise concerns to me, since you are trying to rework a very integral core feature, which could have lasting impact on how this system will be maintained and want to be sure that you completely understand how previous version worked and most improtantly why and how it interacted with other parts of the system,

Like we discussed before, we do not want to maintain multiple versions of how data is gathered and processed, it would be difficult to onboard and keep track of multiple implementations. Which means other backends will need to be also reworked, but like I previously said, `dask` is currently very simple, compared to `postgresql` or `sql` backends and I am woried that it would create more problems trying to refactor them.

> * Not unit testable.

It's just a wrapper, you can easily unit test core features of that wrapper, like `QueryBuilder` itself, `ResultBuilder` or `cast_backend_to_python`.

- `Postgresql` contains `QueryBuilder` tests: https://github.com/atviriduomenys/spinta/blob/0c6427bcc4b16831fadd39ca5583f7943fd3a405/tests/backends/postgresql/test_query.py
- `Sql` also contains similar `QueryBuilder` tests:
https://github.com/atviriduomenys/spinta/blob/1eff8a4b8cf1af828a2452f1d5c6204aeb85a61e/tests/datasets/sql/test_query.py
- `Dask` does not have `QueryBuilder` spcific tests, but it's because at the time when `dask` was added, only simple select was required, so it was just not fully implemented.
- I don't think there are `ResultBuilder` tests, they are mainly tested using requests.
- `cast_backend_to_python` also contains tests for specific datatypes. But they are easily added if need be.

So personally I do not look at `dask_get_all` as specific feature that requires unit tests, since its just a wrapper for reusability sake, you can look at other backends and they will have very similar format.

> * Does not follow established command patterns (multi dispatch).

Like I said, its just a wrapper, you can create backend specific implementations for any other dask backends (it was done this way, because every dask backend would just copy paste the same code block, so it was done for reusability, but if need be, you don't have to use it).

> * Its greatest common denominator for the issue that needs to be refactored as stated in the diagrams.

Same thing as I mentioned in previous comments, that its just a reusability wrapper, you do not have to use, it just wraps a pipline of `QueryBuilder` -> `backend specific data` -> `ResultBuilder`(for properties / functions that could not be parsed during data gathering and filtering process) -> `cast_backend_to_python` -> other system parts that use it to process it further for specific formats. This pipline is used on pretty much all backends, just each of them have different variantions, based on the backend implementation and it's limitations.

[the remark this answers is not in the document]
tuturis: > > > > the cast to backend is still called. The test are source of truth. Again about concers - very subjective. Maybe you could provide some objective metrics? I propose code coverage for example.
> > > > As for all other backends (see diagrams above). The common greatest denominator where the problem comes from and a design antipattern is dask_get_all.
> > > 
> > > 
> > > Well if you could, please provide line of code that calls `cast_backend_to_python`, I couldn't seem to find it, maybe I missed it, since it was included in `dask_get_all`.
> > > I tried to use your branch to test it, but `format` feature no longer properly works there, I cannot see `html` response for `geometry` datatype. `html` `prepare_dtype_for_response` for `Geometry` datatype expects `BaseGeometry` from `shapely` library, it uses it to create openstreetmap link to center point of geometric data. If `cast_backent_to_python` actually is used, then with html format, when using `Geometry` type in manifest and passing string data like `POINT (10 20)` should create the link to map, but if it only returns `POINT (10 20)` as string in `html` format, then it most likely is not used.
> > > Also could you explain what do you mean by `dask_get_all` antipattern. This functions is just a wrapper for reusability, if you look at other backend implementations, you will see that they are very simillar.
> > > Another thing, I do agree that objective metrics are important, but code coverage is not the best metric to make sure that system is actually properly tested, it is a good guardrail, but for this kind of refactor its important to add as many different edge case tests, where system actually gets limit tested, where you combine multiple filters, nest them, in my opinion most effective tests are the ones that try to break the feature by combining features together, that on paper should work together. Now what I said is harder to qunatify, since you need to think about the feature from a users perspective.
> > > From my experience working on this project for almost 3 years I have seen several times when people create tests, that for example test 1 depth nesting and it works, but then you look at the code and its just hardcoded, so once you change it 6 depth nesting everything breaks. And code coverage does not really show these kind of cases.
> > 
> > 
> > Good points, if you could provide AAA pattern unit tests for the cases you described we could talk. As mentioned earlier. It's a draft. image We can discuss the TDD - for the coverage (all dask_get_all,) and by using DDD we can formally ensure there are no edge cases. The dask_get_all is antipattern because its
> > 
> > * Not unit testable.
> > * Does not follow established command patterns (multi dispatch).
> > * Its greatest common denominator for the issue that needs to be refactored as stated in the diagrams.
> > 
> > About nesting if we talk about Ref data types, the referenced prototype code deals with them with no hurdles.
> 
> I am aware it is a draft, for that reason I did not comment on other parts that I did not find correct, I only commented on parts, that showed lack of understanding of how and why the current system works. For example use of `Env` for `Expr` handling and usage of if checks for specific cases (it shows that you do not think about `Expr` as a feature, that it was intended to be (`AST`)). These things raise concerns to me, since you are trying to rework a very integral core feature, which could have lasting impact on how this system will be maintained and want to be sure that you completely understand how previous version worked and most improtantly why and how it interacted with other parts of the system,
> 
> Like we discussed before, we do not want to maintain multiple versions of how data is gathered and processed, it would be difficult to onboard and keep track of multiple implementations. Which means other backends will need to be also reworked, but like I previously said, `dask` is currently very simple, compared to `postgresql` or `sql` backends and I am woried that it would create more problems trying to refactor them.
> 
> > * Not unit testable.
> 
> It's just a wrapper, you can easily unit test core features of that wrapper, like `QueryBuilder` itself, `ResultBuilder` or `cast_backend_to_python`.
> 
> * `Postgresql` contains `QueryBuilder` tests: https://github.com/atviriduomenys/spinta/blob/0c6427bcc4b16831fadd39ca5583f7943fd3a405/tests/backends/postgresql/test_query.py
> * `Sql` also contains similar `QueryBuilder` tests:
>   https://github.com/atviriduomenys/spinta/blob/1eff8a4b8cf1af828a2452f1d5c6204aeb85a61e/tests/datasets/sql/test_query.py
> * `Dask` does not have `QueryBuilder` spcific tests, but it's because at the time when `dask` was added, only simple select was required, so it was just not fully implemented.
> * I don't think there are `ResultBuilder` tests, they are mainly tested using requests.
> * `cast_backend_to_python` also contains tests for specific datatypes. But they are easily added if need be.
> 
> So personally I do not look at `dask_get_all` as specific feature that requires unit tests, since its just a wrapper for reusability sake, you can look at other backends and they will have very similar format.
> 
> > * Does not follow established command patterns (multi dispatch).
> 
> Like I said, its just a wrapper, you can create backend specific implementations for any other dask backends (it was done this way, because every dask backend would just copy paste the same code block, so it was done for reusability, but if need be, you don't have to use it).
> 
> > * Its greatest common denominator for the issue that needs to be refactored as stated in the diagrams.
> 
> Same thing as I mentioned in previous comments, that its just a reusability wrapper, you do not have to use, it just wraps a pipline of `QueryBuilder` -> `backend specific data` -> `ResultBuilder`(for properties / functions that could not be parsed during data gathering and filtering process) -> `cast_backend_to_python` -> other system parts that use it to process it further for specific formats. This pipline is used on pretty much all backends, just each of them have different variantions, based on the backend implementation and it's limitations.

Noted.

SauliusPetraitis:
Laukiam @karina-klinkeviciute

PR #1693, opened 2026-01-26
2316 Fix enum type generation for OpenAPI

martynasracct:
**Summary:**
Fix OpenAPI Enum export. Previously enums were always exported with string type. Now type is selected based on property type. 

Same goes for enum choices: previously, always strings. Now it tries to take values from enum prepare, if given.

**Ticket:**
https://github.com/atviriduomenys/katalogas/issues/2316

### DIFF
--- a/spinta/manifests/open_api/openapi_generator.py
+++ b/spinta/manifests/open_api/openapi_generator.py
@@ -101,10 +101,10 @@ class DataTypeHandler:
         if self.is_enum_property(model_property):
             enum_values = self.get_enum_values(model_property)
             dtype_name = self.get_dtype_name(dtype)
-            enum_schema = copy.deepcopy(self.schema_registry.type_mapping.mappings.get(dtype_name, {"type": "string"}))
-            enum_schema.update({"enum": enum_values, "example": enum_values[0] if enum_values else "UNKNOWN"})
-
-            return enum_schema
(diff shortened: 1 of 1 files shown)


spinta/manifests/open_api/openapi_generator.py
```diff
@@ -92,7 +100,11 @@ def convert_to_openapi_schema(self, model_property) -> dict[str, Any]:
 
         if self.is_enum_property(model_property):
             enum_values = self.get_enum_values(model_property)
-            return {"type": "string", "enum": enum_values, "example": enum_values[0] if enum_values else "UNKNOWN"}
+            dtype_name = self.get_dtype_name(dtype)
+            enum_schema = copy.deepcopy(self.schema_registry.type_mapping.mappings.get(dtype_name, {"type": "string"}))
```
Struchid: I'd maybe extract this part into a separate variable:

`self.schema_registry.type_mapping.mappings.get(dtype_name, {"type": "string"})`

And also, maybe use the unpacking to build the `enum_schema`?

```
enum_schema = {
    **copy.deepcopy(X),
    **{"enum": enum_values, "example": enum_values[0] if enum_values else "UNKNOWN"}
}
```

Only my preferences tho, might as well ignore if deemed unworthy.

PR #1737, opened 2026-02-06
Bool-bug fix only for xml

Gustascct:
Fix for bool-like bug.

The bug appears because the DSA file specifies, that a certain field is a bool but XML passes false and true as strings. Then when the dataframe is being created, it sees that the values for that column should be booleans, but the values are strings, which causes the error.

Fixes the bug by not trying to transform data to a dtype, now it forces "object" on each column which does not trigger checks. The transformation is custom and happens at the end.

Added a rule for spinta check that does not allow non string enums to not have prepare values.

### DIFF
--- a/tests/datasets/dataframe/backends/xml/components/test_read.py
+++ b/tests/datasets/dataframe/backends/xml/components/test_read.py
@@ -9,6 +9,7 @@ from spinta.testing.client import create_test_client
 from spinta.testing.data import listdata
 from spinta.testing.manifest import prepare_manifest
 from spinta.testing.utils import get_error_codes, get_error_context
+from unittest.mock import ANY
 
 
 def test_xml_read(rc: RawConfig, tmp_path: Path):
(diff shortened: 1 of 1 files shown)


spinta/datasets/backends/dataframe/backends/xml/commands/read.py
```diff
@@ -19,10 +19,33 @@
 from spinta.datasets.backends.helpers import is_file_path
 from spinta.dimensions.param.components import ResolvedParams
 from spinta.exceptions import CannotReadResource, UnexpectedErrorReadingData
+from spinta.types.datatype import Boolean
 from spinta.typing import ObjectData
 from spinta.utils.schema import NA
 
 
+def _change_xml_bool_to_python(value):
```
Struchid: You mentioned that 

> Tested this bug with the json backend, it does not appear there, because the json.loads method handles such cases and automatically changes 'false' -> False.

Would it be possible to re-use the logic instead of writing it here twice? Or do the other backends not have this functionality implemented and would ultimately break them?

Gustascct: But then we would have to convert xml to json and then run the function. From what I found other backends solve this issue implicitly, because they support booleans by themselves and the parsing steps solve the issue automatically (for example json.loads function accepts 'false' as a string, but changes it to pythonic False). On the other hand, XML parsing keeps the booleans as strings until the very end and then break when dataframe is populated with values that do not match ex: column is bool (defined in the manifest), but value is string (defined in the XML data file)

tests/datasets/dataframe/backends/xml/components/test_read.py
```diff
@@ -806,3 +806,95 @@ def test_xml_read_many_refs_0(rc: RawConfig, tmp_path: Path):
...
     assert resp.status_code == 200
+
+
+def test_xml_read_bool_enum(rc: RawConfig, tmp_path: Path):
+    test_cases = [
+        ("false", "true"),
+        ("False", "True"),
+        ("0", "1"),
+    ]
+    for first_val, second_val in test_cases:
```
martynasracct: Let's use pytest parametrize instead

```suggestion
@pytest.mark.parametrize("first_val, second_val", [("false", "true"), ("False", "True"), ("0", "1")])
def test_xml_read_bool_enum(rc: RawConfig, tmp_path: Path, first_val: str, second_val: str):
```

martynasracct (approved):

tests/datasets/dataframe/backends/xml/components/test_read.py
```diff
@@ -806,3 +806,95 @@ def test_xml_read_many_refs_0(rc: RawConfig, tmp_path: Path):
...
+        app.authmodel("example/xml/City", ["getall"])
+
+        resp = app.get("/example/xml/City")
+        assert listdata(resp, sort=False) == [
+            (False, "Kaunas"),
+            (True, "Vilnius"),
+        ]
+
+
+def test_xml_read_bool(rc: RawConfig, tmp_path: Path):
```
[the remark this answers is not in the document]
Gustascct: added

spinta/datasets/backends/dataframe/commands/read.py
```diff
@@ -194,7 +194,7 @@ def get_dask_dataframe_meta(model: Model):
     dask_meta = {}
     for prop in model.properties.values():
         if prop.external and prop.external.name:
-            dask_meta[prop.external.name] = spinta_to_np_dtype(prop.dtype)
+            dask_meta[prop.external.name] = "object"
```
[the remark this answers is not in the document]
Gustascct: moved

PR #1779, opened 2026-02-14
#1776 properties with language tags return data on getall (authored b…

karina-klinkeviciute:
properties with language tags return data on getall (authored by Artūras @tuturis )

### DIFF
--- a/spinta/datasets/backends/dataframe/backends/xml/commands/read.py
+++ b/spinta/datasets/backends/dataframe/backends/xml/commands/read.py
@@ -135,7 +135,7 @@ def getall(
 
     for prop in model.flatprops.values():
         if prop.external and prop.external.name:
-            props[prop.name] = {
+            props[prop.place] = {
                 "source": prop.external.name,
                 "pkeys": get_pkeys_if_ref(prop),
             }


spinta/datasets/backends/dataframe/backends/xml/commands/read.py
```diff
@@ -135,6 +135,9 @@ def getall(
     for prop in model.properties.values():
         if prop.external and prop.external.name:
             props[prop.name] = {"source": prop.external.name, "pkeys": get_pkeys_if_ref(prop)}
+        if prop.dtype.name == "text":
```
Struchid: "text" should be in a constant, that could be reused

tests/datasets/dataframe/backends/xml/components/test_read.py
```diff
@@ -806,3 +806,60 @@ def test_xml_read_many_refs_0(rc: RawConfig, tmp_path: Path):
...
+            "name": {"lt": "Kaunas", "en": "Kaunas_en"},
+            "code": "KNS",
+        },
+        {
+            "_type": "example/xml/City",
+            "_revision": None,
+            "name": {"lt": "Vilnius", "en": "Vilnius_en"},
+            "code": "VNO",
+        },
+    ]
```
Struchid: It would look like:


```suggestion
    assert data == [
        {
            "_id": ANY,
            "_type": "example/xml/City",
            "_revision": None,
            "name": {"lt": "Kaunas", "en": "Kaunas_en"},
            "code": "KNS",
        },
        {
            "_id": ANY,
            "_type": "example/xml/City",
            "_revision": None,
            "name": {"lt": "Vilnius", "en": "Vilnius_en"},
            "code": "VNO",
        },
    ]
```

Struchid (approved):

spinta/datasets/backends/dataframe/backends/xml/commands/read.py
Struchid: Yup, I agree with Justinas, this would be a better way, scratch my suggestion then

martynasracct (approved):

Issue #204, opened 2022-03-09
Add support for text type


sirex:
In manifest there is a possibility to separate natural language text properties from any other string types.

Specification: https://atviriduomenys.readthedocs.io/dsa/duomenu-tipai.html#tekstiniai-duomenys

`text` type should be implemented only on `postgresql` backend.

On PostgreSQL `text` type properties should be saves as JSONB objects, like this:

```python
{
  "lt": "Pavyzdys",
  "en": "Example",
  "": "Example",
}
```

Here `""` key will be used, when language code for property is not specified.

- [x] Add support for `text` type.
- [x] Parse `name@lang` string and store language code as type parameter.
- [x] Implement write operation on `text` types.
- [x] Implement read operation on `text` types.
    - [ ] Select language by [Accept-Language](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language).
    - [ ] Fall back to default language specified in configuration.
- [x] Test if patching and changelog works properly.
- [x] Test if reading from external sources works properly.
- [x] Test if pushing works properly.
- [x] Add support for cases, when language code is not given, in this case it should save under empty language code and level must be less than or equal to 3. Level 2 if multiple languages are mixed.
- [x] Test if select(prop@lang) works.
- [x] Test if prop@lang=value works.
- [x] Test if sort(prop@lang) works.

### Links

- https://w3c.github.io/json-ld-syntax/#string-internationalization

sirex:
A suggestion. Maybe we should separate text from string types? Like here:

```
d | r | b | m | property | type   | ref | source  | uri
example                  |        |     |         |
                         | prefix | dct |         | http://purl.org/dc/terms/
  |   |   | City         |        |     | CITIES  |
  |   |   |   | name     | text   |     |         | dct:title
  |   |   |   | name@en  | string |     | NAME_EN |
  |   |   |   | name@lt  | string |     | NAME_LT |
```

Here we have a `name` of `text` type and we can specify `uri` for `name`, which also includes all language tags. We have to specify `uri` only for `text` type, because RDF has built-in support for languages.

`name@en` and `name@lt` are defined as `string`, but since property name has language tag, we know, that these strings are part of `name`.

Also `text` type could be implied, without explicitly declaring it, like this:

```
d | r | b | m | property | type   | ref | source
example                  |        |     |
  |   |   | City         |        |     | CITIES
  |   |   |   | name@en  | string |     | NAME_EN
[...]

PR #1784, opened 2026-02-17
1777 xml įgyvendinti kalbos žymės palaikymą su getall bei search/select užklausą

tuturis:


spinta/backends/helpers.py
```diff
@@ -214,6 +246,14 @@ def _select_prop(
...
     node: Union[Namespace, Model, Property],
 ) -> Optional[Property]:
+    text_props = {p for p in props.values() if isinstance(p.dtype, Text) and p.dtype.langs}
+    if text_props:
+        for prop in text_props:
+            langs = prop.dtype.langs.values()
+            for lang in langs:
+                if lang.place == key:
+                    return lang
+
```
JustinasKen: Similar as above, whats the reason of changing function, that is used everywhere, while `postgresql` did not need to apply these changes ?

spinta/datasets/backends/dataframe/commands/read.py
```diff
@@ -277,6 +281,17 @@ def dask_get_all(
                     val = keymap.encode(sel.prop.model.model_type(), val)
                 elif isinstance(sel.prop.dtype, Ref):
                     val = handle_ref_key_assignment(context, keymap, env, val, sel.prop.dtype)
+                elif sel.prep:
```
Struchid: I would advise against using `sel` as a variable name

spinta/backends/helpers.py
```diff
@@ -89,6 +91,36 @@ def get_select_tree(
     return flat_select_to_nested(select)
 
 
+def get_model_select_tree(
+    context: Context,
+    model: Model,
+    action: Action,
+    select: Optional[List[str]],
+) -> SelectTree:
+    if select is not None and isinstance(select, dict):
```
Struchid: Is it possible to avoid the 6-level nesting here?

spinta/datasets/backends/dataframe/backends/xml/commands/read.py
```diff
@@ -135,6 +136,9 @@ def getall(
     for prop in model.properties.values():
         if prop.external and prop.external.name:
             props[prop.name] = {"source": prop.external.name, "pkeys": get_pkeys_if_ref(prop)}
+        if isinstance(prop.dtype, Text):
+            for lang in prop.dtype.langs.values():
+                props[lang.place] = {"source": lang.external.name, "pkeys": []}
```
JustinasKen: Now this part is a little bit tricky. As far as I remember `dask` backend type did not have nested property implementation (I think at that time nested properties were just being made and `dask` at the time was not priority, only `sql` and `postgresql`). What it means that alot of choices on how things were done, did not take into consideration that properties can be nested.

In this case it might be better to use `model.flatprops`, which would include all properties (meaning it should also cover other nested types, like `Object`, `Denorm`, etc), also this way you might no longer need to do if checks for specific dtypes. Now you would just need to make sure that `props[prop.place]` does not break anything (which it probably doesn't).

spinta/datasets/backends/dataframe/commands/read.py
```diff
@@ -277,6 +281,17 @@ def dask_get_all(
...
+                    prep_dtype = getattr(sel.prep, "dtype", None)
+                    if isinstance(sel.prop.dtype, Text):
+                        val = {}
+                        for lang_key, lang in sel.prop.dtype.langs.items():
+                            val[lang_key] = row[lang.external.name]
+
+                    if not isinstance(sel.prep, Expr) and isinstance(prep_dtype, Text):
+                        key = sel.prop.place
+                        val = row[sel.item]
+
```
JustinasKen: This does not seem right, since `sel.prep` expressions should be handled above. Idealy these parts should not include any if checks for specific types (`PrimaryKey` and `Ref` are hacks (and currently would require some reworks, external `Sql` backend has fixed `Ref` part, but not `PrimaryKey` (race condition issues), these things should be handled using `cast_backend_to_python` when possible).

Struchid (commented):
This fixes the initial problem, but there is still problem loading the data:

```
  File "/Users/X/Projects/vssa/spinta/spinta/utils/json.py", line 31, in fix_data_for_json
    raise TypeError(f"{type(data)} probably won't serialize to JSON.")
TypeError: <class 'spinta.components.Property'> probably won't serialize to JSON.
```

Using this data given in the description of the task:

<details>
<summary>XML</summary>

<?xml version="1.0" encoding="UTF-8"?>
<miestai>
  <miestas>
    <pavadinimas>Kaunas</pavadinimas>
    <kodas>KNS</kodas>
  </miestas>
  <miestas>
    <pavadinimas>Vilnius</pavadinimas>
    <kodas>VNO</kodas>
  </miestas>
  <miestas>
    <pavadinimas>Klaipėda</pavadinimas>
    <kodas>KLJ</kodas>
  </miestas>
  <miestas>
    <pavadinimas>Šiauliai</pavadinimas>
    <kodas>SQQ</kodas>
  </miestas>
  <miestas>
    <pavadinimas>Panevėžys</pavadinimas>
    <kodas>PNV</kodas>
  </miestas>
</miestai>

</details>

<details>
<summary>DSA</summary>

id,dataset,resource,base,model,property,type,ref,source,access
,test_dataset,,,,,,,,
,,resource1,,,,dask/xml,,/Users/gytis/Desktop/test.xml,open
,,,,,,,,,
,,,,City,,,,/miestai/miestas,open
,,,,,name@lt,string,,pavadinimas/text(),open
,,,,,code,string,,kodas/text(),open

</details>

---

Also, we should have a minimal description (a lot of things changed, they should be explained, since spinta's logic is quite hard) and even better, a prepared DSA & XML (data) file before-hand, so the testing could be performed without additional work.

spinta/backends/__init__.py
```diff
@@ -656,8 +656,18 @@ def prepare_data_for_response(
...
+        model,
+        prop_names,
+        value,
+        select,
+        reserved_props,
+    )
+
+    response = {}
+    for prop, val, sel in selected_props:
+        response[prop.place] = commands.prepare_dtype_for_response(
```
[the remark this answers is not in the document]
tuturis: some unit test would prove you right.

spinta/commands/read.py
```diff
@@ -158,7 +158,7 @@ def prepare_data_for_response(
     prop_names = get_select_prop_names(
         context,
         model,
-        model.properties,
+        model.flatprops,
```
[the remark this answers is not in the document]
tuturis: good catch

spinta/backends/helpers.py
tuturis: i  did not got into the postgresql and sql specifics

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
```diff
@@ -210,6 +210,12 @@ def select(env: DaskDataFrameQueryBuilder, prop: Property) -> Selected:
 
 @ufunc.resolver(DaskDataFrameQueryBuilder, DataType)
 def select(env: DaskDataFrameQueryBuilder, dtype: DataType) -> Selected:
+    if dtype.prop.parent and isinstance(dtype.prop.parent.dtype, Text):
+        return Selected(
+            item=dtype.prop.external.name,
+            prop=dtype.prop,
+            prep=dtype.prop.parent,
+        )
```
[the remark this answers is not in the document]
JustinasKen: Sorry, didin't see that it was parent, then it would be better that this is moved to `String` instead of `DataType`, since `Text` children can only be `String` type

spinta/backends/__init__.py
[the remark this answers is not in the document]
JustinasKen: Thing is, I think, since it is root level `prepare_data_for_response`, it should not matter what do you use here, but for `prepare_dtype_for_response` you have to use `prop.name`, otherwise you will get dict structures like this:
```python
{
    "obj_1": {
        "obj_1.obj_2": {
            "obj_1.obj_2.obj_3": "value"
        }
    }
}
```

instead of:
```python
{
    "obj_1": {
        "obj_2": {
            "obj_3": "value"
        }
    }
}
```

So the reason why I said it would be better to keep `prop.name` is so that other people would not be misslead to use `prop.place` in `prepare_dtype_for_response` parts (keep it more consistent).

spinta/datasets/backends/dataframe/commands/read.py
tuturis: the whole usage of dask_get_all is an antipattern and code smell. It's a thick facade with god responsibilities.

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
```diff
@@ -435,6 +447,20 @@ def base64(env: DaskDataFrameQueryBuilder, expr: Expr) -> Expr:
...
+    model = env.model
+    if obj.name in model.properties:
+        prop = model.properties[obj.name]
+        if isinstance(prop.dtype, Text):
+            if attr.name in prop.dtype.langs:
+                return prop.dtype.langs[attr.name]
+            else:
+                raise PropertyNotFound(model, property=obj, lang=attr)
+
+    return GetAttr(obj=obj, attr=attr)
```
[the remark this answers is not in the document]
JustinasKen: Well the idea is to reuse `QueryBuilder` for all backends instead of creating new solutions. Since with AST everything will need to be done through `Env` classes either way.

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
```diff
@@ -288,6 +294,12 @@ def select(env: DaskDataFrameQueryBuilder, dtype: Ref, prep: Any) -> Selected:
     )
 
 
+@ufunc.resolver(DaskDataFrameQueryBuilder, Text, object)
+def select(env: DaskDataFrameQueryBuilder, dtype: Text, prep: Any) -> Selected:
+    prep = {lang: env.call("select", prop, prep) for lang, prop in dtype.langs.items()}
+    return Selected(prop=dtype.prop, prep=prep)
+
+
```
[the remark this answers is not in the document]
JustinasKen: What I meant here is that `text` has unique rules on how default language is supposed to be given, so I wanted to mention that, since it was not mentioned anywhere else.

tests/datasets/dataframe/backends/xml/components/test_read.py
```diff
@@ -806,3 +807,168 @@ def test_xml_read_many_refs_0(rc: RawConfig, tmp_path: Path):
...
+      |   |   |   | name@en  | string   |      | pavadinimas_en/text() | open
+      |   |   |   | code     | string   |      | kodas/text()          | open
+    """,
+        mode=Mode.external,
+    )
+    context.loaded = True
+    app = create_test_client(context)
+    app.authmodel("example/xml/City", ["getall", "search"])
+
+    resp = app.get("/example/xml/City?select(code,name@lt,name@en)")
```
karina-klinkeviciute: beje, siūlau naujuose testuose `select` ir kitas query rašyti pagal naują sinteksę, nes sena bus po truputį deprecate'inta. 

`select naujojoje sintaksėje atrodo taip: https://ivpk.github.io/uapi/#section/Query/Select

spinta/backends/helpers.py
```diff
@@ -89,6 +91,36 @@ def get_select_tree(
...
+                            _select.append(lang.place)
+        select = _select
+    if isinstance(select, dict):
+        select = list(select.keys())
+
+    select = _apply_always_show_id(context, action, select)
+    if select is None and action in (Action.GETALL, Action.SEARCH):
+        # If select is not given, select everything.
+        select = {"*": {}}
+    return flat_select_to_nested(select)
```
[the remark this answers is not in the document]
JustinasKen: Could you go into little bit more details on what case it happens ?

I have tried to check this on https://github.com/atviriduomenys/spinta/blob/0c6427bcc4b16831fadd39ca5583f7943fd3a405/tests/dtypes/test_text.py#L332
`get_select_tree` initialy gets this:
```python
select = {'name.en': GetAttr(obj='name', name=en), 'name.lt': GetAttr(obj='name', name=lt)}
```
the result becomes:
```python
select = {'name': {'en': {}, 'lt': {}}}
```

as I see `prop_names = get_select_prop_names(...)` part only takes the root part of the selected dictionary (meaning it only returns `["name"]`).

Now the select tree (dictionary with languages included) and `prop_names` (only root properties) get passed to `prepare_data_for_response` command, which calls `select_props` function, it then return this response:
```python
prop, value, select = <spinta.components.Property(name='name', type='text', model='types/text/content/Country')>, {'en': 'Lithuania', 'lt': None}, {'en': {}, 'lt': {}}
```

Then we go to `prepare_dtype_for_response` with `Text` type (let's say `json` response format in this case):
https://github.com/atviriduomenys/spinta/blob/0c6427bcc4b16831fadd39ca5583f7943fd3a405/spinta/formats/json/commands.py#L65-L85

and it returns this response:
```python
{'en': 'Lithuania', 'lt': None}
```

which `prepare_data_for_response` converts it to:
```python
{"name": {'en': 'Lithuania', 'lt': None}}
```


Unless you are talking about different scenario, where this logic breaks ?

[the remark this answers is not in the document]
JustinasKen: I think there might be few missunderstandings on how `text` type should be returned and how it works.

I saw in parent issue mention on sql backend:
https://github.com/atviriduomenys/spinta/issues/1658#issuecomment-3828827833

but `text` type without specified language should return default language or language given by `Accept-Language` header in http. This functionality is based on original `text` implementation task, more details here: https://github.com/atviriduomenys/spinta/issues/204#issuecomment-1778606801

`flattened` vs `nested` values are determined by output format, so for example:
`html` or `csv` formats are tabular, so they are `flattened` by default, in those cases `name@lt, name@en` makes sense. But `json` format is nested, so all nested properties should be grouped by their root parent, it becomes like a tree for easier json parsing, I don't think its a good idea to start mixing ideas on how data is represented in json format, since everything else is designed to return child properterties under parent property.

Could you give me a reference to the example where it stated `{ name@en: "en", name@lt: "lt" }` response for `json`, I couldn't find it, since as far as I remember (with the original `text` implementation idea), they should be under the parent and not flattened.

Now all of these missunderstandings and changes could be result of unfinished text documentation and that alot of technical details are scattered across many different comments, but it would be good that @sirex could give definite answers on how it should actually behave (does it follow same logic as `postgresql` did (using http headers, default languages), does it need to work differently, etc).

spinta/backends/__init__.py
```diff
@@ -1738,6 +1740,98 @@ def cast_backend_to_python(
...
+                        context,
+                        dtype.langs.get(dtype_lang_key),
+                        backend,
+                        lang_external[external_name],
+                        **kwargs,
+                    ),
+                )
+                continue
+
+    return dict(iter_lang_values())
```
JustinasKen: Could you move this function to `datasets/backends/dataframe/backends/xml/commands/cast.py`

Its better to keep backend specific implementations under backend folders.

spinta/backends/__init__.py
```diff
@@ -1738,6 +1740,98 @@ def cast_backend_to_python(
...
+                    commands.cast_backend_to_python(
+                        context,
+                        lang.dtype,
+                        backend,
+                        lang_external[external_name],
+                        **kwargs,
+                    ),
+                )
+
+    return dict(iter_lang_values())
```
JustinasKen: Could you move this function to `datasets/backends/dataframe/backends/csv/commands/cast.py`

Its better to keep backend specific implementations under backend folders.

tuturis:
The drone CI build fails due to the diacritics  in the branch name

Gustascct (commented):
I tried launching spinta with my DSA from this [issue](https://github.com/atviriduomenys/spinta/issues/1813), it launched fine but it threw a bunch of warnings indicating, that the changes you made may result in ambiguous behavior.  Adding the warnings below.

Can you fix these warnings so that we protect ourselves from possible errors in the future?


``` 
The following signatures may result in ambiguous behavior:
	[QueryBuilder, ForeignProperty, DataType, Flip], [DaskDataFrameQueryBuilder, ForeignProperty, DataType, object]
	[DaskDataFrameQueryBuilder, Ref, object], [QueryBuilder, DataType, Flip]
	[QueryBuilder, Denorm, Flip], [DaskDataFrameQueryBuilder, DataType, object]
	[DaskDataFrameQueryBuilder, Ref, object], [DaskDataFrameQueryBuilder, DataType, set]
	[QueryBuilder, DataType, ReservedProperty], [DaskDataFrameQueryBuilder, DataType, object]
	[DaskDataFrameQueryBuilder, Ref, object], [QueryBuilder, DataType, ReservedProperty]
	[QueryBuilder, DataType, Flip], [DaskDataFrameQueryBuilder, DataType, object]

```

tuturis:
> > > > > I tried launching spinta with my DSA from this [issue](https://github.com/atviriduomenys/spinta/issues/1813), it launched fine but it threw a bunch of warnings indicating, that the changes you made may result in ambiguous behavior. Adding the warnings below.
> > > > > Can you fix these warnings so that we protect ourselves from possible errors in the future?
> > > > > ```
> > > > > The following signatures may result in ambiguous behavior:
> > > > > 	[QueryBuilder, ForeignProperty, DataType, Flip], [DaskDataFrameQueryBuilder, ForeignProperty, DataType, object]
> > > > > 	[DaskDataFrameQueryBuilder, Ref, object], [QueryBuilder, DataType, Flip]
> > > > > 	[QueryBuilder, Denorm, Flip], [DaskDataFrameQueryBuilder, DataType, object]
> > > > > 	[DaskDataFrameQueryBuilder, Ref, object], [DaskDataFrameQueryBuilder, DataType, set]
> > > > > 	[QueryBuilder, DataType, ReservedProperty], [DaskDataFrameQueryBuilder, DataType, object]
> > > > > 	[DaskDataFrameQueryBuilder, Ref, object], [QueryBuilder, DataType, ReservedProperty]
> > > > > 	[QueryBuilder, DataType, Flip], [DaskDataFrameQueryBuilder, DataType, object]
> > > > > ```
> > > > 
> > > > 
> > > > Yes I have created, and added an item into the backlog at #1815. It should address your issue after cleanup. I will update the description to address your comment.
> > > 
> > > 
> > > But will it be addressed before merging this PR? Such warnings and bugs have a tendency to be forgotten after merging.
> > 
> > 
> > To answer your question straight - it depends if #1815 will get prioritized. As #1815 is labeled as tech-debt, it most likely involves refactoring efforts, and by proxy - not in the scope of the PR ticket. From the error I see that there were some ambiguities before, and only the following was introduced related to his PR (other comes from subclass'ing)
> > ```
> > > > 	[DaskDataFrameQueryBuilder, Ref, object], [DaskDataFrameQueryBuilder, DataType, set]
> > ```
> 
> If I run this DSA on the master branch, there are no ambiguity warnings, so the way I understand it is that all warnings come because of this change.



> This PR introduced inconsistencies, so let's not merge it until they are resolved to not introduce new bugs into the system:
> 
> * #1815



> @tuturis I told you that those issues, documented in #1815 can be addressed as a tech debt in a separate ticket, but I wasn't aware that they are causing actual issues. Which, as we see from the comments made by @Gustascct , they do.
> 
> In this situation, please correct those issues before merging this PR. They can be addressed either in this PR or in a separate one, but it must be solved before merging.

Taking a look if theres a simple way.

PR #1789, opened 2026-02-19
"spinta check" error message on ref filters in column "prepare"

### LINKED ISSUE #1788: The `spinta check` command reports that reference filters are not supported by the Dask backend
The command `spinta check` reports that reference filters (compare operations in the column "prepare") are not supported by the Dask backend (XML, JSON, CSV, WSDL).

04d4:
Fixes #1788

## Changes Summary

Adds validation for unsupported comparison operators in Dask backend `prepare` formulas to the `spinta check` command ([#1788](https://github.com/atviriduomenys/spinta/issues/1788)).

### New file
- [`spinta/datasets/backends/dataframe/commands/check.py`](spinta/datasets/backends/dataframe/commands/check.py) — New `check` command implementation for `DaskBackend`. Recursively inspects `Expr` trees in a model's `prepare` formula, collects any comparison/filter operators (from the `COMPARE` set), and reports them via `ErrorManager` as `DaskBackendCompareNotSupported`.

### Modified files
- [`spinta/cli/config.py`](spinta/cli/config.py) — The [`check()`](spinta/cli/config.py:31) CLI command now iterates over all manifest models and dispatches `commands.check(context, model, backend)` for models that have an external resource with a backend, enabling backend-specific validation.
- [`spinta/exceptions.py`](spinta/exceptions.py) — Added new exception class [`DaskBackendCompareNotSupported`](spinta/exceptions.py:972) with a descriptive template listing the unsupported operators found.
- [`spinta/types/model.py`](spinta/types/model.py) — Added a no-op fallback [`check(Context, Model, Backend)`](spinta/types/model.py:541) registration so backends without specific check logic silently pass. Also consolidated imports (multi-import lines, alphabetical ordering).
- [`tests/cli/test_check.py`](tests/cli/test_check.py) — Added two new tests:
  - `test_spinta_check_message_no_filters_support_dask_backend` — parametrized over `dask/xml`, `dask/json`, `dask/csv`; verifies that unsupported filter operators produce the expected warning messages and error count.
  - `test_spinta_check_no_message_filters_support_sql_backend` — verifies that SQL backends with filters do **not** trigger Dask-specific warnings.
- [`tests/cli/test_copy.py`](tests/cli/test_copy.py) — Parametrized `test_copy_with_filters_and_externals` over `dask/xml` and `sql/sqlite` backend types to ensure the copy command works correctly with both backends.


spinta/datasets/commands/link.py
```diff
@@ -50,6 +68,18 @@ def link(context: Context, entity: Entity):
...
+                and context.get("strict_check")
+                and isinstance(entity.resource.backend, DaskBackend)
+                and entity.prepare is not NA
+            ):
+                ops = _get_compare_operators(entity.prepare)
+                if ops:
+                    context.get("error_manager").handle_error(
+                        exceptions.DaskBackendCompareNotSupported(entity.model, operators=", ".join(ops))
+                    )
+
```
JustinasKen: Now this might be semantics, but as far as I remember we have this workflow:

load -> link -> check

load - used to create and populate majority of the node data
link - used when parts of the data requires other nodes to be loaded (references to other models, backends, etc).
check - used to validate if everything is correct

This looks like it should go under `check` command (I am also aware there are some other `link` commands that do validation checks, which in some cases could be just missunderstanding of work flow).

Another thing, idealy we should not do specific backend checks in backend agnostic features (like global load, link, check functions, that should not be affected by backend type, since they are domain level components).

These kind of checks used to be handled while trying to fetch the data (which might not be ideal situation for this task).

I am aware that this is an edge case and it could be kept here, but maybe this implies that there should be backend specific checks for these domain level compontents (but as far as I could see, I could not find other places where there are `isinstance` checks for specific backends (except `NoBackend` which is more of a concept, rather than actual backend)).

Struchid: I agree with Justinas and ultimately, if you did decide this for a reason, we do not know it, the task description could be used to reveal that reason.

Struchid (approved):
Let's use the task description and describe the implemented changes that solve the problem, not only list the ticket we are doing

PR #1793, opened 2026-02-23
1795 Model with level=0 does not return "_id"

### LINKED ISSUE #1795: Model without `_id` field
**Specifikacijos punktas:**
- https://github.com/atviriduomenys/katalogas/issues/2389
#

This story was split from https://github.com/atviriduomenys/spinta/issues/1030. Requirements are taken from https://github.com/atviriduomenys/spinta/issues/1030#issuecomment-3878319196.

Currently Spinta generates and returns `_id` field for all rows. But in some cases, based on `model.level` `_id` should not be returned, because objects cannot be identified uniquely.

* For model `level < 4` - `_id` should not be generated and returned in response.
* For model `level >= 4` - `_id` should be generated and returned in response.

#### Example 1:

DSA:
```
m | property | type    | level | access
Country      |         | 0     | 
  | id       | integer |       | open
```

Spinta response:
```json
{
    "_data": [
        {
            "_type": "example/Country",
            "_revision": "acbc88c6-a4f0-47ea-b9e1-9917d96794ec",
            "id": 123,
        },
    ],
}
```

#### Example 2:

DSA:
```
m | property | type    | level | access
Country      |         | 4     | 
  | id       | integer |       | open
```

Spinta response:
```json
{
    "_data": [
        {
            "_type": "example/Country",
            "_id": "ad0ca653-bd4c-4658-998a-0d247043d13d",
            "_revision": "acbc88c6-a4f0-47ea-b9e1-9917d96794ec",
            "id": 123,
        },
    ],
}
```

### `spinta push` and Saugykla

`spinta push` is affected by this change, because it heavily relies on **same** exact `_id` for each row existing in data provider's Spinta instance and in Saugykla Spinta instance. With current implementation of `spinta push`, `_id` field is mandatory.

* `spinta push` for manifests with models `level > 3` will work. Changes not needed, because Model `_id` field exists.
* `spinta push` for manifests with models `level < 4` will not work, because Model `_id` will not be generated in data provoder's Spinta. 

Since Saugykla is already used there is no way to implement this task without breaking changes for Saugykla. One of the suggestions, to minimize changes for `spinta push` is to:

* Do not allow `spinta push` if at least one of the models have `level < 4`.
* Raise model level to at least level 4 for all manifests, used by Saugykla manually (before upgrading Spinta version)

If it is not possible to raise model level to `level 4`, because data is missing unique identificator, Spinta could generate one automatically, based on other fields of the model using new `generate_pk()` prepare function:

**Example:**
Given DSA without primary key, for data that doesn't have unique identifier:
```
m | property | type    | level | access | ref | prepare | source
Country      |         | 0     |        |     |         | COUNTRY
  | name     | string  |       | open   |     |         | NAME
```

We could add new dynamically generated `id` field. And instead of `property.source` we could use `property.prepare` function `generate_pk()` that would generate unique identificator based on all or specified fields from the model.

This dynamically generated property could become primary key of the model and this would raise model level to `level 4`.
```
 m | property | type    | level | access | ref | prepare         | source
Country       |         | 4     |        | id  |                 | COUNTRY
   | id       | string  |       | open   |     | generate_pk()   |
   | name     | string  |       | open   |     |                 | NAME
```

### Related

- https://github.com/atviriduomenys/spinta/issues/400
- https://github.com/ivpk/dsa/issues/39
- https://github.com/atviriduomenys/spinta/issues/1159

### Blocks:
- https://github.com/atviriduomenys/spinta/issues/1030

martynasracct:
**Summary:**
Model with `level=0` does not return `_id` column in response.

95% of this PR are new tests, because I had to test columns returned with different Spinta actions and formats.

Main business logic change of this PR is in `spinta/backends/helpers.py`: `_id` is only added if `commands.identifiable(model)` check passes.

Other places has small refactorings or fixes for `Level.absent=0`.

**Ticket:**
https://github.com/atviriduomenys/spinta/issues/1795

### DIFF
--- a/spinta/backends/__init__.py
+++ b/spinta/backends/__init__.py
@@ -671,7 +671,7 @@ def prepare_data_for_response(
             prop_names,
             value,
             select,
-            get_model_reserved_props(action, page_in_data(value)),
+            get_model_reserved_props(model, action, page_in_data(value)),
         )
     }
 
--- a/spinta/backends/helpers.py
+++ b/spinta/backends/helpers.py
@@ -316,19 +316,24 @@ def flat_select_to_nested(select: Optional[List[str]]) -> SelectTree:
     return res
 
 
-def get_model_reserved_props(action: Action, include_page: bool) -> List[str]:
+def get_model_reserved_props(model: Model, action: Action, include_page: bool) -> list[str]:
     if action == Action.GETALL:
-        reserved = ["_type", "_id", "_revision"]
+        reserved = ["_type", "_revision"]
     elif action == Action.SEARCH:
-        reserved = ["_type", "_id", "_revision", "_base"]
+        reserved = ["_type", "_revision", "_base"]
+    # TODO: Should CHANGES and MOVE still return _id or no?
     elif action == Action.CHANGES:
         return ["_cid", "_created", "_op", "_id", "_txn", "_revision", "_same_as"]
     elif action == Action.MOVE:
         return ["_type", "_revision", "_id", "_same_as"]
     else:
-        reserved = ["_type", "_id", "_revision"]
+        reserved = ["_type", "_revision"]
+
+    if commands.identifiable(model):
+        reserved.insert(1, "_id")
     if include_page:
         reserved.append("_page")
+
     return reserved
 
 
--- a/spinta/commands/read.py
+++ b/spinta/commands/read.py
@@ -11,7 +11,7 @@ from spinta import commands
 from spinta.accesslog import AccessLog
 from spinta.accesslog import log_response
 from spinta.backends.components import Backend
-from spinta.backends.helpers import get_select_prop_names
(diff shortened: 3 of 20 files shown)


JustinasKen (commented):
I have a question about `internal` backends (like `postgresql`), currently its designed to always contain `_id`, no matter the `model.level`, how will this affect `saugykla`, at the same time how will this affect `spinta push`, it uses `commands.getall` to fetch the data from `external` backend and sends it to `saugykla` (that expects `_id`).

In `postgresql` backend, `_id` is primary key, so it cannot be nulls.

Now this might not be a issue, since `postgresql` should generate random `_id` when its not given, but I am more concerned about `keymap` and `push.state.db` since as far as I remember it was designed to always contain `_id`. Could be good to have some tests for `push` feature that uses `model.level`.

PR #1817, opened 2026-03-12
1802 - default_access_value in config.py

MindaugasStanionis:
Summary:

- Added configuration value `default_access_value` and set its default value to "private"
- Added parameter access_level to `link_access_param` func to set access to its value, and set its default value to "private"
- Set default value of Manifest component access value to `private`

Ticket: https://github.com/atviriduomenys/spinta/issues/1802

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,8 +8,11 @@ Improvements:
 
 - Added validation to `spinta check` command for comparison operators in Dask backend prepare formulas.
   The check now detects and reports unsupported comparison operators in expressions (`#1788`_).
+- Added configuration value default_access_value and set its default value to `private`.
+  Set default value of Manifest component access value to private (`#1802`_).
 
 .. _#1788: https://github.com/atviriduomenys/spinta/issues/1788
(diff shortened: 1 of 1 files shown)


spinta/types/model.py
```diff
@@ -468,7 +471,10 @@ def link(context: Context, prop: Property):
                 model.ns.parents(),
             )
         )
-    link_access_param(prop, parents, use_given=not prop.name.startswith("_"))
+    config = context.get("config")
+    link_access_param(
+        prop, parents, use_given=not prop.name.startswith("_"), default_access=config.default_access_level
```
Struchid: **Optional:** Since it's not yours and is not a critical problem but:

Maybe a method for the `not prop.name.startswith("_")` which would be a check for non-private properties. E.g. `property_is_private` or `property_is_reserved`. Current check is very implicit and might cause confusion for new readers.

MindaugasStanionis: Good suggestion, added it in f3cd9327c9a5085cf994a3aecd1ad8fd6d2df06c commit

PR #1852, opened 2026-03-30
[1848] Implement UNKNOWN type for `inspect`

Struchid:
The initial solution for handling unrecognized types during `spinta inspect` was not what we needed and propogates further errors. It is needed to implement the full `UNKNOWN` type and support it during this particular command.

This PR only revolves around `spinta inspect` and no other commands, since this is the current number one priority.

! The `unknown` type is set for a field when the type could not be recognized during `spinta inspect`.

### Commits

Due to the codebase already having the initial solution, this PR is separated into commits rather than different PR's so the implementation could come out with the newly supported changes from the get-go, instead of reverting one functionality and introducing the other one.

- Reverting changes from the initial solution #1797
    - [4e03a38](https://github.com/atviriduomenys/spinta/pull/1852/commits/4e03a38a4caf07de46261bf4519e905d47c97f2c)
- Implementing the `unknown` type.
    - [85c384f](https://github.com/atviriduomenys/spinta/pull/1852/commits/85c384fdc0a6c611b599bb3ec75d7830beeacb83)
- Test coverage + versioning notes
    - [0bb60d6](https://github.com/atviriduomenys/spinta/pull/1852/commits/0bb60d6e5371c79e9cd5877c8df246e6ab56ef60)


Struchid:
Will open when tests are prepared, be my guest if you want to review early.

martynasracct (approved):

PR #1853, opened 2026-03-30
Sending data for composite properties

Gustascct:
Support for composite properties (x.y.z...)

Changes:

1. The relevant ref resolvers now include a dictionary as prep which consists of:
 - _id, which is the resolved prepare value or source value.
 - composite properties.

2. The parsing for keymap and data population was altered to accept this dict that may be there for certain manifests

3. Partial datatype link updated, now it saves source immediately (only for external backends) and it searches for the property only one level deep (in get_ref_value), because its immediate parent was already resolved to ref.


spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
```diff
@@ -164,6 +164,9 @@ def select(env: DaskDataFrameQueryBuilder, expr: Expr):
         for prop in take(["_id", all], env.model.properties).values():
             if authorized(env.context, prop, Action.GETALL):
                 env.selected[prop.place] = env.call("select", prop)
+                if hasattr(prop.dtype, "properties") and prop.dtype.properties:
+                    for dtype_prop in prop.dtype.properties.values():
+                        env.selected[dtype_prop.place] = env.call("select", dtype_prop)
```
Gustascct: Change to recursion

Gustascct: Changed, thanks

martynasracct (approved):

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
```diff
@@ -164,6 +164,10 @@ def select(env: DaskDataFrameQueryBuilder, expr: Expr):
         for prop in take(["_id", all], env.model.properties).values():
             if authorized(env.context, prop, Action.GETALL):
                 env.selected[prop.place] = env.call("select", prop)
+                if hasattr(prop.dtype, "properties") and prop.dtype.properties:
```
Struchid: What about? + A walrus to re-use the retrieved attribute if the condition is truthy

```suggestion
                if getattr(prop.dtype, "properties", None):
```

Struchid (approved):

spinta/datasets/backends/dataframe/commands/read.py
```diff
@@ -279,7 +279,16 @@ def dask_get_all(
...
+
+            if sel.prop and sel.prop.parent and sel.prop.parent.name != sel.prop.name:
+                parent = sel.prop.parent.name
+
+                if parent not in res:
+                    res[parent] = {}
+
+                res[parent][sel.prop.name] = val
+            else:
+                res[key] = val
```
JustinasKen: I think dask is just missing:
https://github.com/atviriduomenys/spinta/blob/0c6427bcc4b16831fadd39ca5583f7943fd3a405/spinta/datasets/backends/sql/commands/read.py#L74

`flat_dict_to_nested` function, which is added in other backend reads.

Could you check if adding it, would remove the need to add these checks ?

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
```diff
@@ -164,6 +164,10 @@ def select(env: DaskDataFrameQueryBuilder, expr: Expr):
...
+                    nested_selected = env.call("select", prop.dtype.properties)
+                    for key, value in nested_selected.items():
+                        env.selected[f"{prop.place}.{key}"] = value
```
JustinasKen: Idealy if we could, we should avoid coding datatype specific implementation details inside generic functions.

`properties` field can be used by `Object`, `Ref`, `Backref` types, but what about `Array` type, which uses `items`, or `Text` type, which uses `langs`. I think it would be better to leave datatype specific implementation details to `select(Env, Object)`, `select(Env, Ref)`, etc. This would allow full control and would decouple generic algorithm from datatypes.

JustinasKen (changes requested):

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
Gustascct: This implementation was done here to always include the composite keys in the returned json, no matter if the source or prepare is present or no. I would agree, that moving the implementation to something like `select(Env, Ref)` is better, but then I need answers to these questions, maybe you can look at them? 
https://github.com/atviriduomenys/spinta/issues/1843#issuecomment-4161160553

JustinasKen: Well I tend to reference `sql` implementation, since it the most complete external backend.

We have to look at 2 code blocks:

https://github.com/atviriduomenys/spinta/blob/29d8d6cf1a309dc3e66bd9fa5b94734636f9e761/spinta/datasets/backends/sql/ufuncs/query/ufuncs.py#L381-L420

https://github.com/atviriduomenys/spinta/blob/29d8d6cf1a309dc3e66bd9fa5b94734636f9e761/spinta/datasets/backends/sql/ufuncs/query/ufuncs.py#L443-L453

Looking at this code:

> 1. Jei ref source yra nurodytas, ar composite property yra grąžinami?

```python
```

if `ref` source is given, it will call `select(Env, Ref)`, which will execute above code (it will include all dtype.properties as part of the result).

> 2. Jei ref source yra nenurodytas, ar composite property yra grąžinami? Pagal dabartinį veikimą, net _id nėra grąžinami.

Now there is few conditions, where properties can be processed even if they dont have source given:

```python
            if isinstance(prop.external.prepare, Expr):
                result = env(this=prop).resolve(prop.external.prepare)
            else:
                result = process_literal_value(prop.external.prepare)

            # XXX: Maybe interpretation of prepare formula should be done under
            #      a different Env? This way, select resolvers would know when
            #      properties are resolved under a formula context and for
            #      example would be able to decide to not check permissions on
            #      properties.
            #      tag:resolving_private_properties_in_prepare_context
            result = env.call("select", prop.dtype, result)
```
If `prepare` is given, it will process it as `select(Env, Ref, object)`, which could be something like:
https://github.com/atviriduomenys/spinta/blob/29d8d6cf1a309dc3e66bd9fa5b94734636f9e761/spinta/datasets/backends/sql/ufuncs/query/ufuncs.py#L486-L496

So it will still return `properties` in that case.

There are other cases like:
* prop.is_reserved() - will not work for `ref`, since `is_reserved()` just checks if names starts with `_`.
* not prop.dtype.requires_source - `Ref` datatype does not have that field set, so it states, that it requires source.
* prop.dtype.inherited - This is an exception, where the property does not belong to the model itself and its gained through joins (mainly important when you nest multiple refs, then nested refs does not need source to be given).

If nothing matches it just returns as None.

That means in normal cases, atleast for `sql` backend, `Ref` has to have `source` or `prepare` or be inherited from other properties.

> 3. Jei prepare yra nurodytas, ar composite property yra grąžinami? Ar svarbu, ar tai expr ar ne?

Yes, described it in second point.

Gustascct: I am running into a problem. If I change the resolvers to also include the `dtype.properties.values`, I will mess up the keymap. For example with the `getattr` function, now keymap saves the value of the Property that was resolved from `getattr`. If I include the `dtype.properties.values` it will save a dict instead. 

Same goes for source, if source is included and the ref has composite properties, keymap will save the dictionary and will try to connect the tables using composite properties also.

So I am not sure how to proceed, one idea would be to build _id without using the `dtype.properties.values` which would require rewriting the primary key resolver

JustinasKen: https://github.com/atviriduomenys/spinta/blob/6820986d49804d17dbb60f7f46fcf0b973a4c782/spinta/datasets/backends/dataframe/commands/read.py#L280-L281

This part of the code is responsible for keymaps with `ref` type in `dask` backend.

In `Sql`, this part was reworked, to be included in `cast_backend_to_python`:

https://github.com/atviriduomenys/spinta/blob/aec459867a8b98ca089de42a5aeb34387f60ec75/spinta/datasets/backends/sql/commands/cast.py#L118-L150

https://github.com/atviriduomenys/spinta/blob/aec459867a8b98ca089de42a5aeb34387f60ec75/spinta/datasets/backends/sql/commands/cast.py#L28-L95

So if you changed / modified to return the same way `Sql` does, you would need to apply these similar changes to `dask` backend aswell.

Unless I missunderstood what you meant ?

Gustascct: If I write a Ref resolver (triggered when ref has source but no prepare) which takes in the composite properties like this:
``` 
@ufunc.resolver(DaskDataFrameQueryBuilder, Ref)
def select(env: DaskDataFrameQueryBuilder, dtype: Ref) -> Selected:
    prep = {}
    for prop in dtype.properties.values():
        sel = env.call("select", prop)
        prep[prop.name] = sel

    if not prep:
        return Selected(
            item=dtype.prop.external.name,
            prop=dtype.prop,
        )

    return Selected(
        prop=dtype.prop,
        prep=prep,
    )
```
the _id Selected chain gets built with the composite properties as well. Because of this, keymap gets the dtype.properties.values and not the source value
```
{'_id': Selected(item=None, prop=_id, prep=list...)
- Selected(item=None, prop=asset_type, prep=None)
- Selected(item=None, prop=asset_type_attribute, prep={'code': Selected(item=kodas/text(), prop=asset_type_attribute.code, prep=<NA>)
, 'title_lt': Selected(item=pavadinimas/text(), prop=asset_type_attribute.title_lt, prep=<NA>)})
```


And from my testing, at least for this instance, the `handle_ref_key_assignment` did not change the keymap, it stayed the same as when it was created with the PrimaryKey
````

Comment (JustinasKen) [reply spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py]: Yes this was a issue with `Sql` aswell, for that reason it added:
https://github.com/atviriduomenys/spinta/blob/29d8d6cf1a309dc3e66bd9fa5b94734636f9e761/spinta/datasets/backends/sql/ufuncs/query/ufuncs.py#L446-L447

So it separated Selected properties that are needed for `_id` and others, that are needed for return.

Then `keymap` handling function was changed to look for `_id` key and extract data from that field instead of using everything that was returned with `ref` `prep`. Also that was the reason why it got moved to `cast_backend_to_python`, since having it as if check was a hack.

You can compare 
 @commands.cast_backend_to_python.register(Context, Ref, Sql, dict) 
 
 with 
handle_ref_key_assignment (from dask)

To see the changes, since it was the same function, just extracted and changed to work as `cast_backend_to_python` function

Comment (Gustascct) [reply spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py]: Okay so having different Selected chains for _id and ref properties is valid? For _id the Selected chain would only include the keymap values essentially and ref properties would have all

Comment (JustinasKen) [reply spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py]: Yes, atleast thats what I did. I separated concerns, since having to deal with everything included in the same dict, was just a mess and was prone to alot of issues.

Atleast that solution fixed alot of issues with `Sql` backend and `ref` type.

**Review by Struchid — APPROVED** [review]

File: spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py [hunk 5]
```diff
@@ -353,6 +354,18 @@ def select(env: DaskDataFrameQueryBuilder, dtype: Ref, prep: Any) -> Selected:
     )
 
 
+@ufunc.resolver(DaskDataFrameQueryBuilder, Ref)
+def select(env: DaskDataFrameQueryBuilder, dtype: Ref) -> Selected:
+    prep = {}
+    prep["_id"] = Selected(item=dtype.prop.external.name, prop=dtype.prop)
```
JustinasKen, line 362: Will this work, when Ref is linked with more than 1 key ?

Like: `Model[prop1, prop2]`

Or when `Model` is composite primary key ?

JustinasKen (approved):

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
Gustascct: changed

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
Gustascct: No, this does not and did not work. Will present the changes tomorrow, and point out this flaw, maybe it will get priority

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
Gustascct: Implemented

spinta/datasets/backends/dataframe/commands/read.py
Gustascct: Not needed anymore, but I did test it out with the previous changes and it did not seem to help. Did not do a lot of testing though

MindaugasStanionis (approved):

spinta/datasets/backends/dataframe/commands/read.py
Gustascct, line 1: Collects only values needed for _id

spinta/datasets/backends/helpers.py
Gustascct, line 1: Handles how values are changed to _id. The value that comes is {_id: smth, another_val: {_id: smth, another_another_val: {_id: smth, ...}}} so it recursively goes through the entire value changing the _id values to uuid

spinta/types/partial/link.py
Gustascct, line 1: Partial datatype link updated, now it saves source immediately (only for external backends) and it searches for the property only one level deep (in get_ref_value), because its immediate parent was already resolved to ref. After linking fills the dtype.refprops

spinta/datasets/backends/dataframe/commands/read.py
```diff
@@ -276,6 +276,11 @@ def dask_get_all(
             val = _get_row_value(context, row, sel, env.params)
             if sel.prop:
                 if isinstance(sel.prop.dtype, PrimaryKey):
+                    if isinstance(val, list):
+                        val = [
+                            list_value["_id"] if isinstance(list_value, dict) and "_id" in list_value else list_value
+                            for list_value in val
+                        ]
```
martynasracct, line 283: Suggestion: maybe this way is a bit more readable. But fell free to ignore this suggestion
```suggestion
                        val = [
                            list_value.get("_id", list_value) if isinstance(list_value, dict) else list_value
                            for list_value in val
                        ]
```

JustinasKen (approved):
I dont't really know the scope of this task, but should it include denorm properties like these sql tests:
https://github.com/atviriduomenys/spinta/blob/0c6427bcc4b16831fadd39ca5583f7943fd3a405/tests/datasets/test_sql.py#L3532-L3616

There are more there in that file.

For me biggest gripe in your current tests, that you provide `source` to `ref` property, but you also overwrite it's primary key value by giving it another source. Now this is fine if those values match, but what happens once you give `ref` property one source, but to it's primary key another?

`Denorm` values are special since they are annotated by not giving `type` value, meaning it inherts from `ref` model's property type. By looking at tests, `Denorm` properties ignore `source` field (I don't remember if that was intended or not). But it means that it's value should be gotten by joining that model.

Like I mentioned, I don't know if denorm properties are part of this task scope, so I will not push for it. Similarly, I don't know if this task requires to check support for `?select` url query (as far as I remember you can add `?select` to ref models property and if you have permissions they should be returned (even if they were not declared in original model, that contains the ref property).

Example:

        d | r | m | property            | type    | ref     | source       | prepare | access
    datasets/denorm                 |         |         |              |         |
      | rs                          | sql     |         |              |         |
      |   | Planet                  |         | code    | PLANET       |         | open
      |   |   | code                | string  |         | code         |         |
      |   |   | name                | string  |         | name         |         |
      |   | Country                 |         | code    | COUNTRY      |         | open
      |   |   | code                | string  |         | code         |         |
      |   |   | name                | string  |         | name         |         |
      |   |   | planet              | ref     | Planet  | planet       |         |

`datasets/denorm/Country?_select=planet.name` should work, even if it was not added in manifest

spinta/types/partial/link.py
```diff
@@ -2,23 +2,25 @@
 
 from spinta import commands
 from spinta.components import Context, Property
-from spinta.exceptions import PartialTypeNotFound, ParentNodeNotFound
+from spinta.core.enums import Mode
+from spinta.exceptions import PartialTypeNotFound, ParentNodeNotFound, PartialIncorrectProperty
 from spinta.types.datatype import Partial, Ref
 
 
 def get_ref_value(context: Context, prop: Property):
```
martynasracct: We should add type hint here

spinta/types/partial/link.py
```diff
@@ -2,23 +2,25 @@
...
     parent = prop.parent
     if isinstance(parent, Property) and isinstance(parent.dtype, Ref):
```
martynasracct: We can do early return here. It would reduce identation

martynasracct (approved):

Gustascct:
> I dont't really know the scope of this task, but should it include denorm properties like these sql tests:
> 
> https://github.com/atviriduomenys/spinta/blob/0c6427bcc4b16831fadd39ca5583f7943fd3a405/tests/datasets/test_sql.py#L3532-L3616
> 
> There are more there in that file.
> 
> For me biggest gripe in your current tests, that you provide `source` to `ref` property, but you also overwrite it's primary key value by giving it another source. Now this is fine if those values match, but what happens once you give `ref` property one source, but to it's primary key another?
> 
> `Denorm` values are special since they are annotated by not giving `type` value, meaning it inherts from `ref` model's property type. By looking at tests, `Denorm` properties ignore `source` field (I don't remember if that was intended or not). But it means that it's value should be gotten by joining that model.
> 
> Like I mentioned, I don't know if denorm properties are part of this task scope, so I will not push for it. Similarly, I don't know if this task requires to check support for `?select` url query (as far as I remember you can add `?select` to ref models property and if you have permissions they should be returned (even if they were not declared in original model, that contains the ref property).
> 
> Example:
> 
> ```
>     d | r | m | property            | type    | ref     | source       | prepare | access
> datasets/denorm                 |         |         |              |         |
>   | rs                          | sql     |         |              |         |
>   |   | Planet                  |         | code    | PLANET       |         | open
>   |   |   | code                | string  |         | code         |         |
>   |   |   | name                | string  |         | name         |         |
>   |   | Country                 |         | code    | COUNTRY      |         | open
>   |   |   | code                | string  |         | code         |         |
>   |   |   | name                | string  |         | name         |         |
>   |   |   | planet              | ref     | Planet  | planet       |         |
> ```
> 
> `datasets/denorm/Country?_select=planet.name` should work, even if it was not added in manifest

No, I found this denorm datatype and talked with Mantas a few weeks ago about it, I was also under the impression that this is needed, but we decided to only treat these composite properties as properties and not denorm properties.

spinta/types/partial/link.py
Gustascct: Added two options for early return, check it out!

spinta/types/partial/link.py
Gustascct: Added thanks

spinta/datasets/backends/dataframe/commands/read.py
Gustascct: Yep very pretty, implemented

Struchid (approved):

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
```diff
@@ -336,12 +336,17 @@ def select(env: DaskDataFrameQueryBuilder, selected: Selected) -> Selected:
...
-            prep=prep,
-        )
+    resolved_prep = env.call("select", prep)
+    if resolved_prep is None:
+        return None
+
+    result = {}
+    result["_id"] = Selected(prop=dtype.prop, prep=resolved_prep)
+    for prop in dtype.properties.values():
+        sel = env.call("select", prop)
```
wisecrow, line 344: This result can be `None`. Some resolvers just returns `None` in this case. Shouldn't we have a `None` guard here?

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
```diff
@@ -353,6 +358,18 @@ def select(env: DaskDataFrameQueryBuilder, dtype: Ref, prep: Any) -> Selected:
     )
 
 
+@ufunc.resolver(DaskDataFrameQueryBuilder, Ref)
+def select(env: DaskDataFrameQueryBuilder, dtype: Ref) -> Selected:
+    prep = {}
+    prep["_id"] = Selected(item=dtype.prop.external.name, prop=dtype.prop)
+
+    for prop in dtype.properties.values():
+        sel = env.call("select", prop)
```
wisecrow, line 365: Same here

wisecrow (approved):

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
Gustascct: The prop is a Property so it triggers a select which has no way to return None without failing before

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
Gustascct: The prop is a Property so it triggers a select which has no way to return None without failing before

PR #1865, opened 2026-04-09
[1807] Introduce config.access param

### LINKED ISSUE #1807: `access` level configuration
**Specifikacijos punktas:**
- https://github.com/atviriduomenys/katalogas/issues/2555
#

Currently Spinta only uses following `access` values:

- `open` - uses this for open data and allows unauthorized client access.
- `private` - does not allow data access at all.

We need to add configuration value, to allow other access levels.

Which access level should be used, depends on Spinta environemnt:

- `open` - used only for open data (get.data.gov.lt, Saugykla)
- `public` - used with external API GW (api.gov.lt, Išoriniai vartai)
- `protected` - used with internal API GW (apigw.gov.lt, Vidiniai vartai, SVDPT netwrok)
- `private` - no access

There should be Spinta configuration value `access`, which should limit data access only up to specified access level.

For example if Spinta is configured with:

```yaml
access: public
```

We need to ignore all Manifest elements, with `access < public`.

If user is trying to access elements that have lower access than `config.access`, show `404 NOT FOUND`.

By default `config.access` should be set to `open`.

### Responses matrix:

**UNAUTHENTICATED:**

Image

**AUTHENTICATED:**

Image

<s>This could be done at load time, Spinta would not load any Manifest elemments if `elem.access < config.access`.

This will also enforce requirement, that higher level access elements, can't be linked with lower level elements. But opposite should work fine. This enforcement might brake cases, when model's primary key uses property which is has lower access level than model. In this case, primary key property access level must be aqual or higher of model's access level. Same applies for refs and prepare formulas.

If this restriction would be enforced at load time, then probably there is no need to check access at authorize, like it is done now:</s>

## Related

- https://github.com/atviriduomenys/spinta/issues/1802
- https://github.com/atviriduomenys/katalogas/issues/2416

gts446:
Introduced new config parameter `access`.
It should restrict access to all nodes that are lower `access` level than `config.access`. This will override client scopes. I.e. clients with scopes to access such node, won't be able to do so, if `node.access < config.access`.
However, it returns such nodes as `404 Not Found` instead of `403 Forbidden`.


spinta/types/config.py
```diff
@@ -83,6 +83,11 @@ def load(context: Context, config: Config) -> Config:
         default="private",
         cast=lambda name: get_enum_by_name(Access, name),
     )
+    config.access = rc.get(
+        "access",
+        default="open",
+        cast=lambda name: get_enum_by_name(Access, name),
+    )
```
JustinasKen, line 90: If I understood correctly, this feature will completely ignore clients permissions if node.access is lower than this set value. So for backwards compatibility shouldnt it be `private` default ?

I am thinking about `saugykla`, since after this update, clients that could access private properties (by having special permissions), wont be able to. Now I don't know if there is currently a use case like this, but I feel like these kind of changes should default to work as before the change happened (especially when you set testing env to be `private` value, meaning old behaviour of tests, which in alot of cases reflect `saugykla`, expects for client permissions to not be ignored).

If I missunderstood how it works, or there is a good reason why it should be default `open`, please comment.

spinta/testing/pytest.py
```diff
@@ -55,6 +55,7 @@ def rc():
                     "type": "sqlalchemy",
                     "dsn": "sqlite:///{data_dir}/keymap.db",
                 },
+                "access": "private",
```
JustinasKen: You can leave this here, but idealy we would set testing env configuration parameters here:
https://github.com/atviriduomenys/spinta/blob/bdcdafead4316c08cbd932f4fa17e4c29400aa2e/spinta/config.py#L312-L351

JustinasKen (approved):

spinta/types/config.py
gts446: This was requirement by Mantas, however your point is valid. I will discuss this case with Mantas today.

Struchid (approved):

martynasracct (approved):

wisecrow (approved):

spinta/types/config.py
gts446: Agreed with Mantas, that default `open` should work, because spinta is now only used for open data, and other access levels, than default, are not used in production.

spinta/auth.py
```diff
@@ -858,40 +859,46 @@ def authorized(
     scope_formatter: ScopeFormatterFunc = None,
 ):
     config: Config = context.get("config")
+
+    # Disable access to nodes that have lower access level than config.access
+    if config.access > node.access:
+        if throw:
+            raise ModelNotFound(model=node.name)
```
Struchid, line 866: What about a test case in `test_auth.py` for this?

Struchid (approved):

spinta/auth.py
gts446: https://github.com/atviriduomenys/spinta/pull/1865/changes/2660c1628c6c2d3871db661c8b0d45e6f1887249

PR #1889, opened 2026-04-21
[1872] Implicit `link` change for external references

### LINKED ISSUE #1872: `link` eigos metu atliekamas implicitinis pakeitimas su `external ref` reikšmėmis
#### Specifikacijos punktas:
- https://github.com/atviriduomenys/spinta/issues/1093

#### Situacija

Tarkim turint tokį DSA:
```csv
id,dataset,resource,base,model,property,type,ref,source,source.type,prepare,origin,count,level,status,visibility,access,uri,eli,title,description
,example,,,,,,,,,,,,,,,,,,,
,,resource,,,,sql,,,,,,,,,,,,,,
,,,,,,,,,,,,,,,,,,,,
,,,,Country,,,id,COUNTRY,,,,,,develop,private,,,,,
,,,,,code,string,,CODE,,,,,,develop,private,,,,,
,,,,,id,integer,,ID,,,,,,develop,private,,,,,
,,,,,name,string,,NAME,,,,,,develop,private,,,,,
,,,,,city,ref,example2/City,,,,,,,develop,private,,,,,
,,,,,,,,,,,,,,,,,,,,
```

Leidžiant `spinta run` bus metamos klaidos, nes `ref` negalima susieti su `example2/City` modeliu (tokio modelio faile nėra).

#### Sprendimas

Po apsitarimų atskirų susitikimų metu buvo nuspręsta, kad bus atliekamas implicitinis pakeitimas:

> Mano supratimu, šią užduotį galima spręsti `link` komandos vykdymo metu, atliekant implicit transformaciją, nepriklausomai ar leidžiama `spinta comment`, `run`, `check` ar kita CLI komanda.

> Visais atvejais, `link` komandoje, jei aptinkama, kad `ref` modelis nėra deklaruotas, tada `property.type` keičia į `object` ir įtraukia `comment` eilutę.

> Siūlau keisti į `object`, o ne `string`, nes tada bus lengviau spręsti ir jungtinių modelių atvejį.

- https://github.com/atviriduomenys/spinta/issues/1872#issuecomment-4258442980

Struchid:
Implicitly handle undeclared `ref`/`backref`/`array` targets during manifest `link` part.

Previously if a property, having one of the types defined above, pointed to a model not declared in the manifest, spinta raised `ModelReferenceNotFound` and refused to load. This made it impossible to work with partial manifests where only some models are described.

This use case is required due to the requirements proposed by @riteris13, who has difficulties when models are defined through different manifest files.

#### What changes:

During the `link` phase, when the target model is not found, the property is implicitly downgraded instead of raising:
- `dtype` is replaced with `object` (level 2);
- a comment row is appended with a spyna restore expression (`update(type:ref, ref:example2/Country)`) so the original intent is preserved and can be restored in the future.

**To support comment round-tripping**, `Comment` gained `prepare` and `level` fields, with read/write support added for both the tabular and internal-sql manifest backends.
                                                                                               
Tests cover:
- `spinta check`, `spinta copy`, and `spinta run` (`CSV`, `SQLite`, `XML` data sources) don't fail with undeclared `refs`;
- All three `dtype` paths (`ref`, `backref`, `array`) transform correctly, including the bracket `refprops` syntax (`ref:example2/Country[code]`)

### DIFF
--- a/spinta/dimensions/comments/helpers.py
+++ b/spinta/dimensions/comments/helpers.py
@@ -20,12 +20,14 @@ def load_comments(
             access_given = params.pop("access")
             access = enum_by_name(parent, "access", Access, access_given)
             access = access or Access.private
+            level = int(level) if (level := params.pop("level", None)) else None
             comments.append(
                 Comment(
                     access=access,
                     given=CommentGiven(
                         access=access_given,
                     ),
+                    level=level,
                     **params,
                 )
             )
--- a/spinta/types/helpers.py
+++ b/spinta/types/helpers.py
@@ -1,16 +1,23 @@
 from __future__ import annotations
 
-
-from typing import TYPE_CHECKING, Iterable
+import logging
+from datetime import datetime, timezone
+from typing import TYPE_CHECKING, Iterable, List
 
 from spinta import exceptions
 from spinta.components import Config
 from spinta.components import Context
 from spinta.components import Model
 from spinta.components import Property
+from spinta.core.enums import Access, load_level, Level
+from spinta.dimensions.comments.components import Comment, CommentGiven
 from spinta.exceptions import BackendNotFound
 from spinta.exceptions import InvalidName
+from spinta.types import TYPE_OBJECT
 from spinta.utils.naming import is_valid_model_name, is_valid_property_name
+
+logger = logging.getLogger(__name__)
+
 
 if TYPE_CHECKING:
     from spinta.types.datatype import DataType
@@ -38,6 +45,8 @@ RESERVED_PROPERTY_NAMES = {
 
 C_LANG = "C"
 
+SYSTEMIC_COMMENT_AUTHOR = "author"
+
 
 def check_no_extra_keys(dtype: DataType, schema: Iterable, data: Iterable):
     unknown = set(data) - set(schema)
@@ -60,6 +69,53 @@ def set_dtype_backend(dtype: DataType):
             dtype.backend = backends[dtype.backend]
     else:
         dtype.backend = dtype.prop.model.backend
+
+
+def replace_undeclared_ref_with_object(
+    context: Context,
+    prop: Property,
+    original_type: str,
+    original_model: str,
+    refprops: List[str] | None = None,
+) -> None:
+    from spinta.types.datatype import Object
+
+    ref_str = f"{original_model}[{','.join(refprops)}]" if refprops else original_model
+
+    logger.warning(
+        "Model %r referenced by %r property %r was not found in the manifest. Downgrading to object (level 2).",
+        original_model,
+        original_type,
+        prop.place,
+    )
+
+    new_dtype = Object()
+    new_dtype.name = TYPE_OBJECT
+    new_dtype.type = TYPE_OBJECT
+    new_dtype.type_args = []
+    new_dtype.prop = prop
+    new_dtype.properties = {}
+    prop.dtype = new_dtype
+    set_dtype_backend(new_dtype)
+
+    load_level(context, prop, Level.structured)
+
+    if prop.comments is None:
+        prop.comments = []
+
+    prop.comments.append(
+        Comment(
+            id=None,
+            parent="type",
+            author=SYSTEMIC_COMMENT_AUTHOR,
+            access=Access.private,
+            created=datetime.now(timezone.utc).isoformat(),
+            comment="",
+            given=CommentGiven(access=None),
+            prepare=f"update(type:{original_type}, ref:{ref_str})",
+            level=4,
+        )
+    )
 
 
 def check_model_name(context: Context, model: Model):
--- a/tests/types/test_link.py
+++ b/tests/types/test_link.py
@@ -1,7 +1,11 @@
+from pathlib import Path
+
 import pytest
 
+from spinta.core.config import RawConfig
 from spinta.exceptions import MissingRefModel
-from spinta.testing.manifest import load_manifest
+from spinta.testing.manifest import load_manifest, load_manifest_get_context
+from spinta.types.datatype import Object
 
 
 @pytest.mark.manifests("internal_sql", "csv")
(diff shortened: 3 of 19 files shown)


martynasracct (approved):

tests/types/test_link.py
```diff
@@ -43,3 +46,136 @@ def test_raises_missing_ref_model(manifest_type, tmp_path, rc):
...
+      |   |   |   | country  | ref    | example2/Country | A      | private |
+    """,
+        manifest_type=manifest_type,
+        tmp_path=tmp_path,
+    )
+    store = context.get("store")
+    city = store.manifest.get_objects()["model"]["example/City"]
+    country_prop = city.properties["country"]
+
+    assert country_prop.dtype.name == "object"
```
JustinasKen: We should avoid checking `DataType` through `name`, since it does not support inheritance and the values could be changed, breaking the tests / checks.

Whenever you can please use something like this:
```python
isinstance(country_prop.dtype, Object)
```

tests/types/test_link.py
```diff
@@ -43,3 +46,136 @@ def test_raises_missing_ref_model(manifest_type, tmp_path, rc):
...
+      |   |   |   | country  | ref    | example2/Country[code] | private
+    """,
+        manifest_type=manifest_type,
+        tmp_path=tmp_path,
+    )
+    store = context.get("store")
+    city = store.manifest.get_objects()["model"]["example/City"]
+    country_prop = city.properties["country"]
+
+    assert country_prop.dtype.name == "object"
```
JustinasKen: Same deal as above

tests/types/test_link.py
```diff
@@ -43,3 +46,136 @@ def test_raises_missing_ref_model(manifest_type, tmp_path, rc):
...
+      |   |   |   | countries   | backref | example2/Country | private
+    """,
+        manifest_type=manifest_type,
+        tmp_path=tmp_path,
+    )
+    store = context.get("store")
+    city = store.manifest.get_objects()["model"]["example/City"]
+    countries_prop = city.properties["countries"]
+
+    assert countries_prop.dtype.name == "object"
```
JustinasKen: [Same issue](https://github.com/atviriduomenys/spinta/pull/1889/changes#r3121901454)

tests/types/test_link.py
```diff
@@ -43,3 +46,136 @@ def test_raises_missing_ref_model(manifest_type, tmp_path, rc):
...
+      |   |   |   | countries | array  | example2/CityCountry[city,country]| private
+    """,
+        manifest_type=manifest_type,
+        tmp_path=tmp_path,
+    )
+    store = context.get("store")
+    city = store.manifest.get_objects()["model"]["example/City"]
+    countries_prop = city.properties["countries"]
+
+    assert countries_prop.dtype.name == "object"
```
JustinasKen: [Same issue](https://github.com/atviriduomenys/spinta/pull/1889/changes#r3121901454)

JustinasKen (approved):
I have a question about partial manifests, since I haven't really dealt with them. What are they ?

In cases where I saw multiple manifest files, they are usually conncted into one using `spinta copy`.

If by multiple manifests you mean multiple instances of different manifests working together in single application, wouldn't these errors indicate that there are issues with model resolving using multiple manifests, which would be a different bug ?

But if the model exists in different file it should still work, I have tested it myself just now, when configuring manifests, you can set `sync` property and give a list of other manifests that should be concatinated into root manifest. That way it can reference models from other files, as long as those file actually are added to application and connected to main manifest using `sync` property in configuration, it seemed to work, no error when launching the app.

```yml
    manifests:
      default:
        type: csv
        path: ...
        sync:
          - another
      another:
        type: csv
        path: ...
```


Or I missunderstood what partial manifests mean ?

spinta/types/helpers.py
```diff
@@ -62,6 +71,52 @@ def set_dtype_backend(dtype: DataType):
...
+        original_type,
+        prop.place,
+    )
+
+    new_dtype = Object()
+    new_dtype.name = TYPE_OBJECT
+    new_dtype.type = TYPE_OBJECT
+    new_dtype.type_args = []
+    new_dtype.prop = prop
+    new_dtype.properties = {}
```
wisecrow, line 97: Just curious: shouldn't dtypes resolve a backend with `set_dtype_backend(new_dtype)` ?

spinta/dimensions/comments/helpers.py
```diff
@@ -20,12 +20,14 @@ def load_comments(
             access_given = params.pop("access")
             access = enum_by_name(parent, "access", Access, access_given)
             access = access or Access.private
+            level = int(level) if (level := params.pop("level", None)) else None
```
wisecrow, line 23: if `level` is 0, this expression will be resolved to `None`.  Probably should guard it using `None if level is None else int(level)`

wisecrow (approved):

Gustascct (approved):

spinta/types/helpers.py
Struchid: Yup, I missed it, thanks! Amended the old commit:
- https://github.com/atviriduomenys/spinta/compare/e3abceaa823975bd4f3b6d2756639cb3c51db32c..e7d30dd533a1ef7a2ae157c9c1d5d0f51fa7e6d2

spinta/dimensions/comments/helpers.py
Struchid: You're right to think that this would be falsy on an integer `0` value. However, in this specific path `params` comes from CSV parsing, where values are always strings. `"0"` is a non-empty string and therefore - truthy.

And your suggested change would also introduce a bug that could happen: it would call int("") on an empty cell in the CSV file and raise a `ValueError`.

Leaving it as is, but leaving this thread open if you want to discuss further :)

Struchid:
> I have a question about partial manifests, since I haven't really dealt with them. What are they ?
> 
> In cases where I saw multiple manifest files, they are usually conncted into one using `spinta copy`.
> 
> If by multiple manifests you mean multiple instances of different manifests working together in single application, wouldn't these errors indicate that there are issues with model resolving using multiple manifests, which would be a different bug ?
> 
> Or I missunderstood what partial manifests mean ?

So first of all, I apologize for introducing a term I myself thought of without additional explanation :D

What I meant by partial manifests is manifest (DSA) files that do not contain **all the information** they need inside themselves. 

Answering one of your concerns - yes, if the second file containing the referenced dataset & model would be used together, there would be no problem and it would get linked correctly as intended and work perfectly.

---

But this solves another problem, let's start with an example.

Let's say `manifest_1.csv` file contains a model, that has a property of type `ref`, let's say it is `city`:

```csv
id,dataset,resource,base,model,property,type,ref,source,source.type,prepare,origin,count,level,status,visibility,access,uri,eli,title,description
,example,,,,,,,,,,,,,,,,,,,
,,resource,,,,sql,,,,,,,,,,,,,,
,,,,,,,,,,,,,,,,,,,,
,,,,Country,,,id,COUNTRY,,,,,,develop,private,,,,,
,,,,,code,string,,CODE,,,,,,develop,private,,,,,
,,,,,id,integer,,ID,,,,,,develop,private,,,,,
,,,,,name,string,,NAME,,,,,,develop,private,,,,,
,,,,,city,ref,example2/City,,,,,,,develop,private,,,,,
,,,,,,,,,,,,,,,,,,,,
```

The `example2` dataset is not defined in this file, therefore we can't find model `City` of dataset `example2`, and this is the only file being launched (by one of the `spinta` commands). 

Before this change, we would have got an exception, signaling, that the referenced Model could not be found (`ModelReferenceNotFound`) and the process would be finished.

Though, this does not work for business people. They raised the concern, because a lot of manifest files from institutions (e.g. the files defined in the table of this tickets description: https://github.com/atviriduomenys/dvms/issues/428) are partial. Partial in a sense that they are split in a way, that to launch a single file, you need to go down a reference tree and download >10 files because the references go deep down. 

We have communicated, that the best course of action would be to leave it as is to avoid any unexpected/implicit changes, but the business people were adamant.

That's why we with @sirex have decided, that this will be solved by an implicit change in the `link` step. It was decided that it would be the safest course of action for the use-case we need here that would not propogate and destroy any existing logic.

Would be happy to answer any other questions regarding this, or explain something more if this was not good enough @JustinasKen :-)

tests/types/test_link.py
Struchid: I hope you're not tired from pointing this out to everyone, thanks for the reminder! :D

Adjusted in:
- https://github.com/atviriduomenys/spinta/compare/e7d30dd533a1ef7a2ae157c9c1d5d0f51fa7e6d2..dc093053f33d8edb4df20fcba7ada6489cf7507b

tests/types/test_link.py
Struchid: https://github.com/atviriduomenys/spinta/pull/1889#discussion_r3122514443

tests/types/test_link.py
Struchid: https://github.com/atviriduomenys/spinta/pull/1889#discussion_r3122514443

tests/types/test_link.py
Struchid: https://github.com/atviriduomenys/spinta/pull/1889#discussion_r3122514443

JustinasKen:
> I have a question about partial manifests, since I haven't really dealt with them. What are they ?
> 
> In cases where I saw multiple manifest files, they are usually conncted into one using spinta copy.
> 
> If by multiple manifests you mean multiple instances of different manifests working together in single application, wouldn't these errors indicate that there are issues with model resolving using multiple manifests, which would be a different bug ?
> 
> Or I missunderstood what partial manifests mean ?
> 
> So first of all, I apologize for introducing a term I myself thought of without additional explanation :D
> 
> What I meant by partial manifests is manifest (DSA) files that do not contain all the information they need inside themselves.
> 
> Answering one of your concerns - yes, if the second file containing the referenced dataset & model would be used together, there would be no problem and it would get linked correctly as intended and work perfectly.
> 
> But this solves another problem, let's start with an example.
> 
> Let's say manifest_1.csv file contains a model, that has a property of type ref, let's say it is city:
> 
> id,dataset,resource,base,model,property,type,ref,source,source.type,prepare,origin,count,level,status,visibility,access,uri,eli,title,description
> ,example,,,,,,,,,,,,,,,,,,,
> ,,resource,,,,sql,,,,,,,,,,,,,,
> ,,,,,,,,,,,,,,,,,,,,
> ,,,,Country,,,id,COUNTRY,,,,,,develop,private,,,,,
> ,,,,,code,string,,CODE,,,,,,develop,private,,,,,
> ,,,,,id,integer,,ID,,,,,,develop,private,,,,,
> ,,,,,name,string,,NAME,,,,,,develop,private,,,,,
> ,,,,,city,ref,example2/City,,,,,,,develop,private,,,,,
> ,,,,,,,,,,,,,,,,,,,,
> The example2 dataset is not defined in this file, therefore we can't find model City of dataset example2, and this is the only file being launched (by one of the spinta commands).
> 
> Before this change, we would have got an exception, signaling, that the referenced Model could not be found (ModelReferenceNotFound) and the process would be finished.
> 
> Though, this does not work for business people. They raised the concern, because a lot of manifest files from institutions (e.g. the files defined in the table of this tickets description: https://github.com/atviriduomenys/dvms/issues/428) are partial. Partial in a sense that they are split in a way, that to launch a single file, you need to go down a reference tree and download >10 files because the references go deep down.
> 
> We have communicated, that the best course of action would be to leave it as is to avoid any unexpected/implicit changes, but the business people were adamant.
> 
> That's why we with @sirex have decided, that this will be solved by an implicit change in the link step. It was decided that it would be the safest course of action for the use-case we need here that would not propogate and destroy any existing logic.
> 
> Would be happy to answer any other questions regarding this, or explain something more if this was not good enough @JustinasKen :-)

Thanks for the explanation, sadly cannot open `dvms` project issues, I get 404. I guess my thoughts in all of this would, be if they don't have model definitions and cannot actually connect them, maybe they could just remove that `ref` property in their own manifest file, but maybe I just don't know the full situation what commands they run / how they use it.

Struchid:
> > I have a question about partial manifests, since I haven't really dealt with them. What are they ?
> > In cases where I saw multiple manifest files, they are usually conncted into one using spinta copy.
> > If by multiple manifests you mean multiple instances of different manifests working together in single application, wouldn't these errors indicate that there are issues with model resolving using multiple manifests, which would be a different bug ?
> > Or I missunderstood what partial manifests mean ?
> > So first of all, I apologize for introducing a term I myself thought of without additional explanation :D
> > What I meant by partial manifests is manifest (DSA) files that do not contain all the information they need inside themselves.
> > Answering one of your concerns - yes, if the second file containing the referenced dataset & model would be used together, there would be no problem and it would get linked correctly as intended and work perfectly.
> > But this solves another problem, let's start with an example.
> > Let's say manifest_1.csv file contains a model, that has a property of type ref, let's say it is city:
> > id,dataset,resource,base,model,property,type,ref,source,source.type,prepare,origin,count,level,status,visibility,access,uri,eli,title,description
> > ,example,,,,,,,,,,,,,,,,,,,
> > ,,resource,,,,sql,,,,,,,,,,,,,,
> > ,,,,,,,,,,,,,,,,,,,,
> > ,,,,Country,,,id,COUNTRY,,,,,,develop,private,,,,,
> > ,,,,,code,string,,CODE,,,,,,develop,private,,,,,
> > ,,,,,id,integer,,ID,,,,,,develop,private,,,,,
> > ,,,,,name,string,,NAME,,,,,,develop,private,,,,,
> > ,,,,,city,ref,example2/City,,,,,,,develop,private,,,,,
> > ,,,,,,,,,,,,,,,,,,,,
> > The example2 dataset is not defined in this file, therefore we can't find model City of dataset example2, and this is the only file being launched (by one of the spinta commands).
> > Before this change, we would have got an exception, signaling, that the referenced Model could not be found (ModelReferenceNotFound) and the process would be finished.
> > Though, this does not work for business people. They raised the concern, because a lot of manifest files from institutions (e.g. the files defined in the table of this tickets description: [atviriduomenys/dvms#428](https://github.com/atviriduomenys/dvms/issues/428)) are partial. Partial in a sense that they are split in a way, that to launch a single file, you need to go down a reference tree and download >10 files because the references go deep down.
> > We have communicated, that the best course of action would be to leave it as is to avoid any unexpected/implicit changes, but the business people were adamant.
> > That's why we with @sirex have decided, that this will be solved by an implicit change in the link step. It was decided that it would be the safest course of action for the use-case we need here that would not propogate and destroy any existing logic.
> > Would be happy to answer any other questions regarding this, or explain something more if this was not good enough @JustinasKen :-)
> 
> Thanks for the explanation, sadly cannot open `dvms` project issues, I get 404. I guess my thoughts in all of this would, be if they don't have model definitions and cannot actually connect them, maybe they could just remove that `ref` property in their own manifest file, but maybe I just don't know the full situation what commands they run / how they use it.

The link you can't open consists of a numerous amount of DSA + data examples (XML files) for RC. And they are separated into a numerous amount of files; 

AFAIK, they did not want this approach as well because it does not solve the initial problem - they want to avoid as much additional work as possible running _spinta_ commands. If this was not the case, they could just use the multiple manifest files and follow the refs through multiple of the files as well (the initial solution we suggested). 

I do not strongly approve of this approach, due to it having even a slight chance of causing problems down the line, but it is what has been decided and what we went for.

PR #1893, opened 2026-04-23
Cooperating backends for DASK

### LINKED ISSUE #1905: Cooperating backends only _id part.

This ticket is only for the reserved property `_id`.
Full information: 

Gustascct:
Introduces specific handling for _id properties defined in the manifest (cooperating backends).

The logic for _id handling is more or less universal, but each backend will have specifics on how it uses it, so this PR is only for Dask backend (XML, JSON).

_revision field handling in a separate PR.

Most of the diff is tests, left helper comments


spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
```diff
@@ -238,6 +238,13 @@ def select(env: DaskDataFrameQueryBuilder, prop: Property) -> Selected:
         elif prop.external and prop.external.name:
             # If prepare is not given, then take value from `source`.
             result = env.call("select", prop.dtype)
+        elif prop.name == "_id" and prop is getattr(prop.model.external, "id_prop", None):
+            pkeys = prop.model.external.pkeys or list(take(prop.model.properties).values())
+            if len(pkeys) == 1:
+                prep = env.call("select", pkeys[0])
+            else:
+                prep = [env.call("select", pk) for pk in pkeys]
+            result = Selected(prop=prop, prep=prep)
```
Gustascct: _id is not handled as a PrimaryKey, because it is assigned its own type. This part makes sure that it gets the same selected chain a PrimaryKey would get.

spinta/manifests/tabular/helpers.py
```diff
@@ -2538,10 +2538,16 @@ def _model_to_tabular(
...
+    if (
+        model.external
+        and not model.external.unknown_primary_key
+        and all(p.access >= access for p in model.external.pkeys)
+    ):
+        # Add `ref` only if all properties are available in the
+        # resulting manifest. `ref` lists the model's primary-key
+        # property names and is preserved even when source metadata
+        # is stripped (e.g. `copy --no-source`).
+        data["ref"] = ", ".join([p.given.name or p.name for p in model.external.pkeys])
```
Gustascct, line 2556: spinta copy --no-source strips sources and model.ref. If a manifest has _id, the _id field has to have model.ref or a source, so this saves model.ref if it can exist

spinta/types/datatype.py
```diff
@@ -383,6 +387,12 @@ def check(context: Context, dtype: DataType):
     pass
 
 
+@commands.check.register(Context, Base32)
+def check(context: Context, dtype: Base32):
+    if dtype.prop.name != "_id":
+        raise exceptions.Base32TypeOnlyAllowedOnId(dtype.prop, property=dtype.prop.place)
```
Gustascct, line 393: Only _id can have Base32 as a type

spinta/datasets/backends/dataframe/commands/read.py
```diff
@@ -284,6 +285,8 @@ def dask_get_all(
                     val = keymap.encode(sel.prop.model.model_type(), val)
                 elif isinstance(sel.prop.dtype, Ref):
                     val = handle_ref_key_assignment(context, keymap, env, val, sel.prop.dtype)
+                elif sel.prop is id_prop and isinstance(val, list):
+                    val = ",".join("" if list_value is None else str(list_value) for list_value in val)
```
Gustascct: Used for composite model.ref only (value1, value2)

spinta/types/model.py
```diff
@@ -463,6 +469,37 @@ def link(context: Context, prop: Property):
...
+                    model_primary_key_dtype = model_ref_set[0][0].dtype
+                if model_ref_set and (type(prop.dtype), type(model_primary_key_dtype)) in INCORRECT_DTYPES:
+                    raise ReservedPropertyTypeShouldMatchPrimaryKey(
+                        property=prop,
+                        model=model.name,
+                        reserved_type=prop.dtype.name,
+                        primary_type=model_primary_key_dtype.name,
+                    )
+
+        setattr(model.external, attr, prop)
```
Gustascct: Sets _id property, runs checks for manifest correctness and for types, ex: a string should not be able to become an int

tests/datasets/dataframe/backends/xml/components/test_reserved_props.py
```diff
@@ -0,0 +1,684 @@
...
+            mode=Mode.external,
+        )
+        context.loaded = True
+        app = create_test_client(context)
+        app.authmodel("example/Region", ["getall", "getone"])
+
+        resp = app.get("/example/Region/:format/html")
+        assert resp.status_code == 200
+        header = resp.context["header"]
+        rows = [{k: cell.as_dict() for k, cell in zip(header, row)} for row in resp.context["data"]]
```
Struchid: Does this extract only the property rows? If so, you could do:


```suggestion
        row_uuid_id, row_private_id = [{k: cell.as_dict() for k, cell in zip(header, row)} for row in resp.context["data"]]
```

And use those in asserts.

spinta/utils/url.py
```diff
@@ -100,6 +101,9 @@ def build_url_path(query: List[UrlParseNode]):
     for param in query:
         name = param["name"]
         args = param["args"]
+        id_prop = param.get("id_prop")
+        if id_prop and is_accessible_by_equals_sign(id_prop, args[-1]):
```
Struchid: **Optional:** Could use walrus here:


```suggestion
        if (id_prop := param.get("id_prop")) and is_accessible_by_equals_sign(id_prop, args[-1]):
```

spinta/formats/html/helpers.py
```diff
@@ -186,6 +187,9 @@ def get_model_link_params(
             "args": (model.name.split("/") + ([pk] if pk is not None else []) + ([prop] if prop is not None else [])),
         }
     ]
+    if isinstance(model, Model) and model.external is not None:
+        if id_prop := model.external.id_prop:
```
Struchid: Conditions could be combined

spinta/datasets/components.py
```diff
@@ -213,6 +213,8 @@ class Entity(External):
     name: str  # model.source
     prepare: Expr  # model.prepare
     params: List[Param]
+    id_prop: Optional[Property] = None
+    revision_prop: Optional[Property] = None
```
Struchid: Let's use the built in types for typehinting instead of the `typing` lib ones:


```suggestion
    id_prop: Property | None = None
    revision_prop: Property | None = None
```

spinta/backends/helpers.py
```diff
@@ -596,3 +596,20 @@ def extract_table_data_from_logical_name(table_name: str) -> tuple[str | None, T
...
+    if isinstance(id_prop.dtype, Base32):
+        return True
+
+    if isinstance(id_prop.dtype, String):
+        return check_if_model_primary_key_is_composite(id_prop.model)
+
+    return False
+
+
+def check_if_model_primary_key_is_composite(model: Model):
```
Struchid: Return type is missing

spinta/types/model.py
```diff
@@ -463,6 +469,37 @@ def link(context: Context, prop: Property):
     prop.enum = _link_prop_enum(prop)
 
 
+def _detect_cooperating_reserved_properties(model: Model) -> None:
+    for name, attr in (("_id", "id_prop"), ("_revision", "revision_prop")):
```
Struchid: Could we define this tuple we are looping through above with a name on what is it?

spinta/types/model.py
```diff
@@ -44,6 +47,8 @@
 if TYPE_CHECKING:
     from spinta.datasets.components import Attribute
 
+INCORRECT_DTYPES = [(Integer, UUID), (Integer, String), (UUID, String), (UUID, Integer)]
```
Struchid: This would seem more truthy:

```suggestion
INCORRECT_DTYPE_COUPLES = [(Integer, UUID), (Integer, String), (UUID, String), (UUID, Integer)]
```

spinta/types/model.py
```diff
@@ -463,6 +469,37 @@ def link(context: Context, prop: Property):
...
+        if prop is None or not prop.explicitly_given:
+            return
+        if name == "_id":
+            model_ref_set = prop.model.unique
+            reserved_id_source_set = prop.external.name
+
+            if not model_ref_set and not reserved_id_source_set:
+                raise ReservedPropertySourceOrModelRefShouldBeSet(property=prop)
+            if reserved_id_source_set and model_ref_set:
+                raise ReservedPropertySourceOrModelRefShouldBeSet(property=prop)
```
Struchid: Why not combine these as well?


```suggestion
            if (not model_ref_set and not reserved_id_source_set) or (reserved_id_source_set and model_ref_set):
                raise ReservedPropertySourceOrModelRefShouldBeSet(property=prop)
```

Struchid (approved):

spinta/backends/helpers.py
```diff
@@ -596,3 +596,20 @@ def extract_table_data_from_logical_name(table_name: str) -> tuple[str | None, T
...
+    unique = getattr(model, "unique", None) if model is not None else None
+    if unique and len(unique[0]) > 1:
+        return False
+    return True
```
Domchix: Isn't this bool inverted? Shouldn't more than one unique mean that there is a composite pk?

martynasracct (approved):

spinta/backends/helpers.py
Gustascct: biblical catch, thanks!

spinta/types/model.py
Gustascct: Combined

spinta/types/model.py
Gustascct: Renamed

spinta/types/model.py
Gustascct: Defined

spinta/backends/helpers.py
Gustascct: Added

spinta/datasets/components.py
Gustascct: Changed

spinta/formats/html/helpers.py
Gustascct: Combined

spinta/utils/url.py
Gustascct: Using walrus now

tests/datasets/dataframe/backends/xml/components/test_reserved_props.py
Gustascct: Is the point of this change to not have rows[0] and rows[1]? The naming you provided is weird so I might be understanding incorrectly

Struchid: Yup, the naming can be anything you choose, but `rows[0]` and `rows[1]` does not state what is the thing you got, when I think in this case we can define the variables that would state what they hold.

Let's discuss if you disagree or do not think the same way.

Gustascct: I kind of disagree, I understand rows[0] and rows[1] as first row and second row, but I agree that having a variable name would leave nothing to interpretation. Will change to first_row_data, second_row_data

Struchid: If you disagree, let's not change anything then. If it's clear enough for you, it will be clear enough for readers as well, these are only my preferences, but I might not be invested that deep into this implementation.

Gustascct: Renamed, if it was not clear to you it may not be clear to others, thanks!

spinta/backends/__init__.py
```diff
@@ -590,6 +604,18 @@ def is_object_id(context: Context, value: str):
...
+                return False
+            candidate = value[1:]
+        try:
+            id_prop.dtype.load(candidate)
+        except exceptions.InvalidValue:
+            return False
+        return True
     try:
         return uuid.UUID(value).version == 4
     except ValueError:
```
JustinasKen, line 614: If we have multiple supported datatypes for primary key, we could expand this function further and probably use dispatchers for all supported types.

Also I have noticed everywhere  `id_prop = getattr(model.external, "id_prop", None)`

I assume this is how you check if this property was given / overwritten. But couldn't we just use `_id` property everywhere instead of doing `getattr()`, since as far as I saw you only care that id is not `PrimaryKey` datatype, which would be default `_id` behaviour ?

So why not just use dispatchers for it (unless users can set type as pk and it should not follow old behaviour) ?

```python
@is_object_id.register(Context, Backend, Model, str)
def is_object_id(context: Context, backend: Backend, model: Model, value: str):
    return is_object_id(context, backend, model.properties["_id"].dtype, value)


@is_object_id.register(Context, Backend, PrimaryKey, str)
def is_object_id(context: Context, backend: Backend, dtype: PrimaryKey, value: str):
    try:
        return uuid.UUID(value).version == 4
    except ValueError:
        return False


@is_object_id.register(Context, Backend, DataType, str)
def is_object_id(context: Context, backend: Backend, dtype: DataType, value: str):
    candidate = value
    if is_accessible_by_equals_sign(dtype.prop, value):
        if not value.startswith("="):
            return False
        candidate = value[1:]
    try:
        dtype.load(candidate)
    except exceptions.InvalidValue:
        return False
    return True
```

spinta/types/model.py
```diff
@@ -463,6 +470,35 @@ def link(context: Context, prop: Property):
...
+                    model_primary_key_dtype = model_ref_set[0][0].dtype
+                if model_ref_set and (type(prop.dtype), type(model_primary_key_dtype)) in INCORRECT_DTYPE_COUPLES:
+                    raise ReservedPropertyTypeShouldMatchPrimaryKey(
+                        property=prop,
+                        model=model.name,
+                        reserved_type=prop.dtype.name,
+                        primary_type=model_primary_key_dtype.name,
+                    )
+
+        setattr(model.external, attr, prop)
```
JustinasKen: Is it necessary to add `id_prop` and `revision_prop` to `model.external` ?

My biggest concerns are, if you add them for quick access to `_id` and `_revision` properties, then its fine, you dont need to hardcode strings everywhere, but as I see in code you also you them to see if the `_id` was overwritten or not, which in my opinion makes it more messy, since now you instead of just passing `id_prop` as `_id`, you have to add checks everywhere if its not `None` and if it is, then default to previous behaviour, instead of using `dtype` field to know the property behaviour.

Now this is fine, but I feel like it adds alot of complexity everwhere, since now, every single feature that tries to interact with `_id` will need to add `getattr` for `id_prop` and then do if else `None` checks.

JustinasKen (approved):

spinta/backends/__init__.py
Gustascct: Introduced the dispatchers you suggested.

Removed id_prop all together, it was for easy access but indeed, it requires quite a few checks before I can use it, so instead added a helper function which checks if the default  _id is a PrimaryKey and if it is explicitly given. Completely relying on only the _id property I think would be harder to implement, a lot of checks would still be needed.

spinta/types/model.py
Gustascct: Removed id_prop all together, it was for easy access but indeed, it requires quite a few checks before I can use it, so instead added a helper function which checks if the default  _id is a PrimaryKey and if it is explicitly given. Completely relying on only the _id property I think would be harder to implement, a lot of checks would still be needed.

PR #1899, opened 2026-04-28
Cooperating backends for SQL

### LINKED ISSUE #1905: Cooperating backends only _id part.

Gustascct:
Introduces specific handling for _id properties defined in the manifest (cooperating backends).

This PR is only for SQL backend

_revision field handling in a separate PR.

Most of the diff is tests, tried to stay as close to the implementation of DASK as possible


spinta/datasets/backends/sql/ufuncs/query/ufuncs.py
```diff
@@ -401,6 +401,13 @@ def select(env: SqlQueryBuilder, prop: Property) -> Selected:
         elif prop.external and prop.external.name:
             # If prepare is not given, then take value from `source`.
             result = env.call("select", prop.dtype)
+        elif prop.name == "_id" and prop is getattr(prop.model.external, "id_prop", None):
```
Struchid: `prop is getattr(prop.model.external, "id_prop", None)` Are you really checking for the same specific object or that it would be an instance of some class here? 

If it's the second one, maybe the use of `isinstance()` would be a better approach?

Struchid (approved):

spinta/datasets/backends/sql/ufuncs/query/ufuncs.py
Gustascct: I am checking for this specific object

spinta/datasets/backends/sql/commands/read.py
```diff
@@ -66,6 +68,8 @@ def getall(
                 if sel.prop:
                     if isinstance(sel.prop.dtype, PrimaryKey):
                         val = generate_pk_for_row(context, sel.prop.model, row, keymap, val)
+                    elif sel.prop is id_prop and isinstance(val, list) and not isinstance(id_prop.dtype, Base32):
+                        val = encode_composite_string_id(val, model.external.pkeys)
```
JustinasKen: Could you comment why this check is needed, idealy this part should not even have `if isinstance(sel.prop.dtype, PrimaryKey):`, these things should be moved to `cast_backend_to_python`, unless you really need to handle it here.

JustinasKen (approved):

spinta/datasets/backends/sql/commands/read.py
Gustascct: This handles what happens with composite _id values, like id, code. The function adds both properties together separated by a comma and checks if the values are correct (no special characters, commas, etc.). It skips this check for base32, because base32 is there to allow special symbols.

I could probably move it to cast_backend_to_python but it would result in more if statements because it would be passed as a string

PR #1917, opened 2026-05-04
_revision implementation

### LINKED ISSUE #1906: Cooperating backends only _revision part.

This ticket is only for the reserved property _revision.
Full information:

Gustascct:
Implementation for _revision field support in manifest files.

This implementation is for Dask (which is XML, JSON and by extension, WSDL and CSV) and for SQL. 

Allows adding _revision to the manifest file, made _revision visible in HTML also, not only JSON.

NOTES:
- Tests for WSDL coming, they will be the same as for other backends.
- Dask does not support multiple properties in prepare yet (val1, val2), so the tests for composite _revision are skipped for now. If you want to review full tests, I advise you to look only at SQL, no tests skipped there..

### DIFF
--- a/spinta/datasets/backends/sql/commands/read.py
+++ b/spinta/datasets/backends/sql/commands/read.py
@@ -75,6 +75,12 @@ def getall(
                         is_custom_id_prop(sel.prop) and isinstance(val, list) and not isinstance(sel.prop.dtype, Base32)
                     ):
                         val = encode_composite_string_id(val, model.external.pkeys)
+                    elif (
+                        sel.prop.name == "_revision"
+                        and isinstance(val, (list, tuple))
+                        and not isinstance(sel.prop.dtype, Base32)
+                    ):
+                        val = encode_composite_string_id(val, model.external.pkeys)
                 res[key] = val
             if is_page_enabled:
                 res["_page"] = get_page_values(env, row)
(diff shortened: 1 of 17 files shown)


martynasracct (approved):

spinta/datasets/backends/sql/commands/read.py
```diff
@@ -143,6 +146,14 @@ def getone(
...
             data[field] = value
 
+    revision = model.properties.get("_revision")
+    if revision and revision.explicitly_given and revision.external:
+        if revision.external.name:
+            data["_revision"] = row[revision.external.name]
+        elif isinstance(revision.external.prepare, Expr) and revision.external.prepare.name == "testlist":
+            values = [row[model.properties[arg.args[0]].external.name] for arg in revision.external.prepare.args]
+            data["_revision"] = values if isinstance(revision.dtype, Base32) else ",".join(str(v) for v in values)
+
```
JustinasKen, line 157: Idealy expr should be handled through `Env` class, handling outside it is considered a hack. By the looks of it, it seems that getone is not using `QueryBuilder` and `ResultBuilder`, which could be partial reason why there is no good expr handling setup for this function.

This can be kept, but eventually it will need either separte resultbuilder or reuse `SqlResultBuilder` and add `testlist` feature there

JustinasKen (approved):

spinta/datasets/backends/sql/commands/read.py
Gustascct: Good point, left a comment addressing this.

PR #1929, opened 2026-05-08
1915 citus spinta migrate

### LINKED ISSUE #1915: citus spinta migrate palaikymas
1691 užduoties pull request pridėjo `citus` palaikymą ir distribution migracijas. Distribution migracijos vyksta per `spinta admin` komandą, reikia atnaujinti `spinta migrate` funkcionalumą atsižvelgiant į `citus` limitacijas.

JustinasKen:
- Added citus table distribution / undistribution logic to `migrate` command.
- Fixed `pkey` not being affected when soft deleting table.
- Removed `foreign_key` attribute from migration order logic (outdated and was no longer needed after previous patches).

Need to add CHANGES.rst


martynasracct (approved):

spinta/backends/postgresql/helpers/migrate/migrate.py
```diff
@@ -1331,7 +1338,12 @@ def generate_model_tables_mapping(
     return mapped_tables
 
 
-def create_table_migration(table: sa.Table, handler: MigrationHandler, table_identifier: TableIdentifier = None):
+def create_table_migration(
+    migration_ctx: PostgresqlMigrationContext,
+    table: sa.Table,
+    table_identifier: TableIdentifier = None,
```
Struchid:
```suggestion
    table_identifier: TableIdentifier | None = None,
```

spinta/backends/postgresql/helpers/migrate/migrate.py
```diff
@@ -1527,3 +1540,38 @@ def update_primary_key(
...
+    if distribution_type is None:
+        return
+
+    handler = migration_ctx.handler
+    match distribution_type:
+        case DistributionType.COPY:
+            handler.add_action(DistributeReference(table_identifier=table_identifier))
+            migration_ctx.distribute_plan.discard(table_identifier)
+        case _:
+            pass
```
Struchid, line 1577: Maybe some log here to know that the DistributionType was not found and therefore nothing was executed?

Struchid (approved):

spinta/backends/postgresql/helpers/migrate/migrate.py
JustinasKen: This might be a case where you need to log (maybe I should add comment to these functions).

There are 3 stages where distributions / undistributions can happen:

1. Before migrations
2. During migrations (after table has been changed)
3. After migrations

Idealy all of the distributions would happen before or after migrations, which is alot easier to maintain, but some (in this case `COPY`) needs to be assigned during migrations, since they are sensitive to foreign key constraint changes.

Currently it this scenario:
1. Before migrations
    * Undistribute all schema based (since their conditions demand that they are isolated, so undistributing before doing any changes wont cause any fkey constraint issues with other schemas)
2. During migrations (after table has migrated)
    * Undistribute from `table` type (distributed by columns), in case there are new references, which will add fkey constraints, since current database structure does not support fkey constraint between different table distributions
    * Distribute to `reference` type (which copies to all shards). This type is mainly used for fkey constraints, so future model migrations will going to need to have references to it (so you must distribute during migrations and cannot do it before, since there might be constraints attached, that need to be cleaned up beforehand).
 3. After migrations
     * Undistribute everything that has not yet been undistributed
     * Distribute everything that has not yet been distributed

PR #1930, opened 2026-05-08
1922: Adding ScopeLoader to update Env and pass it to ufuncs

### LINKED ISSUE #1922: Scope dimensijoje, prepare funkcija turi būti validuojama
**Specifikacijos punktas:**
- https://github.com/atviriduomenys/katalogas/issues/1395
#

## Problema

`prepare` stulpelio formulė Scope dimensijoje šiuo metu tikrinama tik per `spyna.parse` — patikrinama tik sintaksė. Jei formulėje nurodomas neegzistuojantis property, stulpelis nevaliduojamas teisingai.

## Sprendimas

Po `spyna.parse` papildomai patikrinti, ar visi formulėje naudojami property egzistuoja modelio scope'e. Jei ne — grąžinti aiškią validavimo klaidą.

## Priėmimo kriterijai

- [x] Formulė su neegzistuojančiu property nepraeina validavimo
- [x] Klaidos pranešime nurodomas trūkstamas stulpelis ir formulės vieta
- [x] Pridėti testai galiojančiai formulei ir formulei su neegzistuojančiu stulpeliu

MindaugasStanionis:
Summary:
- Added new ScopeLoader class
- Added `link_scopes` func to resolve new ufunc resolvers
- Added ufuncs.py in scopes, with validate_prepare resolvers for bind objects, and for plain objects

Ticket: https://github.com/atviriduomenys/spinta/issues/1922

### DIFF
--- a/spinta/dimensions/scope/components.py
+++ b/spinta/dimensions/scope/components.py
@@ -1,11 +1,8 @@
-from spinta.components import ExtraMetaData
+from spinta.components import ExtraMetaData, Model
 from spinta.core.enums import Access, Visibility, Status
 from spinta.core.ufuncs import Expr, Env
 from spinta.manifests.components import Manifest
-
-
-class ScopeLoader(Env):
-    manifest: Manifest
+from spinta.ufuncs.propertyresolver.components import PropertyResolver
 
 
 class ScopeGiven:
@@ -49,3 +46,17 @@ class Scope(ExtraMetaData):
 
     def __init__(self):
         self.given = ScopeGiven()
+
+
+class ScopeLoader(Env):
+    manifest: Manifest
+    model: Model
+    property_resolver: PropertyResolver | None
+
+    def resolve_property(self, *args, **kwargs):
+        resolver = self._scope.get("property_resolver")
+        if resolver is None:
+            resolver = PropertyResolver(self.context)
+            resolver = resolver.init(model=self.model, ufunc_types=False)
+            self.update(property_resolver=resolver)
+        return resolver.resolve_property(*args, **kwargs)
--- a/spinta/dimensions/scope/components.py
+++ b/spinta/dimensions/scope/components.py
@@ -1,6 +1,11 @@
 from spinta.components import ExtraMetaData
 from spinta.core.enums import Access, Visibility, Status
-from spinta.core.ufuncs import Expr
+from spinta.core.ufuncs import Expr, Env
+from spinta.manifests.components import Manifest
+
+
+class ScopeLoader(Env):
+    manifest: Manifest
 
 
 class ScopeGiven:
--- a/spinta/dimensions/scope/components.py
+++ b/spinta/dimensions/scope/components.py
@@ -1,3 +1,5 @@
+from typing import Any
+
 from spinta.components import ExtraMetaData, Model
 from spinta.core.enums import Access, Visibility, Status
 from spinta.core.ufuncs import Expr, Env
@@ -53,7 +55,7 @@ class ScopeLoader(Env):
     model: Model
     property_resolver: PropertyResolver | None
 
-    def resolve_property(self, *args, **kwargs):
+    def resolve_property(self, *args, **kwargs) -> Any:
         resolver = self._scope.get("property_resolver")
         if resolver is None:
             resolver = PropertyResolver(self.context)
--- a/spinta/dimensions/scope/helpers.py
+++ b/spinta/dimensions/scope/helpers.py
@@ -1,5 +1,6 @@
 from typing import cast
 
+from spinta import commands
 from spinta.components import Context, Model
 from spinta.core.access import load_access_param
 from spinta.core.enums import load_level, load_status, load_visibility
@@ -7,14 +8,6 @@ from spinta.core.ufuncs import Expr
 from spinta.dimensions.scope.components import Scope, ScopeLoader
 from spinta.manifests.components import Manifest
 from spinta.nodes import load_node
-
-
-def link_scopes(context: Context, manifest: Manifest, scopes: dict[str, Scope]) -> None:
-    if not scopes:
-        return
-    loader = ScopeLoader(context)
-    loader.update(manifest=manifest)
-    loader.resolve(Expr("resolve_scope", list(scopes.values())))
 
 
 def _load_scope(
@@ -47,3 +40,17 @@ def load_scopes(
     if not scopes:
         return {}
     return {name: _load_scope(context, parents, name, data) for name, data in scopes.items()}
+
+
+def link_scopes(context: Context, manifest: Manifest, scopes: dict[str, Scope]) -> None:
+    if not scopes:
+        return
+    loader = ScopeLoader(context)
+    loader.update(manifest=manifest)
+    loader.resolve(Expr("resolve_scope", list(scopes.values())))
+
+
+def finalize_scope_link(context: Context, manifest: Manifest) -> None:
+    for model in commands.get_models(context, manifest).values():
+        if model.scopes:
+            link_scopes(context, manifest, model.scopes)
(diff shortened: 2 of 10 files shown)


spinta/dimensions/scope/ufuncs.py
```diff
@@ -0,0 +1,47 @@
...
+            env.call("validate_prepare", model, expr.args[0])
+        return
+    if expr.name == "bind":
+        if expr.args and isinstance(expr.args[0], str):
+            env.call("validate_prepare", model, Bind(expr.args[0]))
+        return
+    for arg in expr.args:
+        env.call("validate_prepare", model, arg)
+    for value in expr.kwargs.values():
+        env.call("validate_prepare", model, value)
```
JustinasKen: We should avoid checking `expr.name` as much as possible, especially for `bind` and `getattr`. Now I understand you probably need to do this, since you do not want to remove `Expr` resolver (you would need to add all posbile `prepare` functions that way here).

The way you are handling `getattr` in my opinion might create issues with nested properties, since you are only checking root parent. If you only care about checking if the property exists I would suggest looking into `PropertyResolver` class, it specializes into converting `Bind` and `GetAttr` objects into properties. But you would probably need to call `expr.resolve(env)` on `getattr` and `bind` and then pass them into property resolver.

For that you might need to create `_getattr` resolvers that would create `GetAttr` object.
I see that there are multiple places where `getattr` resolver is duplicated:
https://github.com/atviriduomenys/spinta/blob/0c6427bcc4b16831fadd39ca5583f7943fd3a405/spinta/ufuncs/requestparamsbuilder/ufuncs.py#L10-L23

Maybe it would be best to move it to:
https://github.com/atviriduomenys/spinta/blob/635cc43efccd8b57e3e15b6bf948d85bc8316096/spinta/ufuncs/binds.py#L9-L16

So `getattr` and `bind` would be generic `Env` resolvers. That way you would only need to check if `expr.name` is `bind` or `getattr`, then call `result = expr.resolve(env)`, given result should be then passed to `env.property_resolver.resolve_property(result)`, or something similar.

This pattern is used for `QueryBuilder` (you would just need to set `ufunc_types` to `False`, since you only care about spinta types):
https://github.com/atviriduomenys/spinta/blob/0c6427bcc4b16831fadd39ca5583f7943fd3a405/spinta/ufuncs/querybuilder/components.py#L122-L128
https://github.com/atviriduomenys/spinta/blob/05065d29cc5427ed262a23dbb190a3cb146b557c/spinta/ufuncs/querybuilder/ufuncs.py#L76-L79

tests/types/test_link.py
```diff
@@ -180,3 +180,82 @@ def test_undeclared_array_model_transforms_to_object(manifest_type, tmp_path, rc
...
+    d | r | b | m | property | type   | ref | prepare              | access
+    example                  |        |     |                      |
+      |   |   | City         |        | id  |                      |
+      |   |   |              | scope  | bad | country.name = 'lt'  | private
+      |   |   |   | id       | string |     |                      | private
+      |   |   |   | code     | string |     |                      | private
+            """,
+            manifest_type=manifest_type,
+            tmp_path=tmp_path,
+        )
```
JustinasKen, line 327: You should add valid nested property tests, for `Denorm`, `Object`, `Text`, etc types.

JustinasKen (approved):

spinta/dimensions/scope/ufuncs.py
MindaugasStanionis: Thanks for recomendation, did not known that, added in 1bf997122e16e9373a88cd295026771ce2a5b5f1

tests/types/test_link.py
MindaugasStanionis: Added more tests in 1bf997122e16e9373a88cd295026771ce2a5b5f1

spinta/dimensions/scope/ufuncs.py
```diff
@@ -0,0 +1,57 @@
+from spinta.core.ufuncs import ufunc, Expr, Bind, GetAttr, asttoexpr
+from spinta.dimensions.scope.components import ScopeLoader, Scope
+
+
+@ufunc.resolver(ScopeLoader, str, name="bind")
+def bind_(env: ScopeLoader, name: str) -> Bind:
+    return Bind(name)
```
JustinasKen: This might be unecessary, since `Env` already has this resolver and `ScopeLoader` should inherit it,

spinta/dimensions/scope/helpers.py
```diff
@@ -37,3 +40,17 @@ def load_scopes(
...
+        return
+    loader = ScopeLoader(context)
+    loader.update(manifest=manifest)
+    loader.resolve(Expr("resolve_scope", list(scopes.values())))
+
+
+def finalize_scope_link(context: Context, manifest: Manifest) -> None:
+    for model in commands.get_models(context, manifest).values():
+        if model.scopes:
+            link_scopes(context, manifest, model.scopes)
```
JustinasKen: It seems that there is no longer need to pass `manifest`, also it seems like scopes from `model.scopes` `scope.model` will always be the same `model`. Because of that you could change `loader.update(manifest=manifest)` to `loader.update(model=model)`.

That way you could also remove `resolve_scope` `model` and `property_resolver` resets, since the scopes in `ScopeLoader` are always from the same model. Unless there are plans in the future where that might not be the case ?

spinta/dimensions/scope/ufuncs.py
```diff
@@ -0,0 +1,57 @@
...
+    for scope in scopes:
+        env.call("resolve_scope", scope)
+
+
+@ufunc.resolver(ScopeLoader, Scope)
+def resolve_scope(env: ScopeLoader, scope: Scope):
+    prepare = scope.prepare
+    if not prepare:
+        return
+    env.update(model=scope.model, property_resolver=None)
```
JustinasKen: Might be unecessary in current scenario, since all of the scopes are from the same model (atleast in current link scenario).

JustinasKen (approved):

spinta/dimensions/scope/ufuncs.py
MindaugasStanionis: Thanks, added in f521ac9eb1597d8144356f0e7a184bb4a660baf9

spinta/dimensions/scope/helpers.py
MindaugasStanionis: Yes, for now, we can pass only the model, because scopes will always be the same model

spinta/dimensions/scope/ufuncs.py
MindaugasStanionis: Yes, dont need to update the loader

spinta/dimensions/scope/components.py
```diff
@@ -44,3 +46,17 @@ class Scope(ExtraMetaData):
...
     def __init__(self):
         self.given = ScopeGiven()
+
+
+class ScopeLoader(Env):
+    manifest: Manifest
+    model: Model
+    property_resolver: PropertyResolver | None
+
+    def resolve_property(self, *args, **kwargs):
```
Struchid: Would it be possible to add the return type here?

spinta/dimensions/scope/ufuncs.py
Struchid, line 1: Return type is missing for quite a number of methods in this file

Struchid (approved):

Gustascct (approved):

spinta/dimensions/scope/components.py
MindaugasStanionis: Good catch, total oversight on my part. Added in b90ed532aedf631f9b237d5f308378068d7e4fa4

spinta/dimensions/scope/ufuncs.py
MindaugasStanionis: Added in b90ed532aedf631f9b237d5f308378068d7e4fa4

PR #1949, opened 2026-05-26
1937: Adding custom_scope as UrlParam that we get from url param. Add…

### LINKED ISSUE #1937: Custom scope `prepare` pritaikymas užklausoms
**Specifikacijos punktas:**
- https://github.com/atviriduomenys/katalogas/issues/1395
#

### Tikslas

Užklausos metu vartotojas gali nurodyti scope formatu `Model/@{scope_name}`. Server'is taiko to scope'o `prepare` išraišką kaip duomenų apribojimą — vartotojas mato tik tai, ką leidžia scope.

**Esminis principas:** scope yra **viršutinė matomumo riba**. Papildomi user'io filtrai (`?` query parametrai) gali tik **dar labiau susiaurinti** rezultatą, bet niekada negali grąžinti daugiau, nei leidžia pats scope.

### Pavyzdys

Manifeste apibrėžtas scope `ltu`:

```
scope | ltu | | | country.code='lt'
```

| Užklausa | Rezultatas |
|----------|-----------|
| `GET /City/@ltu` | Visos LT eilutės |
| `GET /City/@ltu?country.code='lv'` | **Tuščia** (user filter prieštarauja scope'ui) |
| `GET /City/@ltu?name='Vilnius'` | Vilnius (LT IR vardas Vilnius) |
| `GET /City/@ltu?country.code='lt'&name='Vilnius'` | Vilnius |

### Priimtinumo kriterijai

- [x] Scope formatu `Model/@scope_name` parsuojamas iš užklausos
- [x] Scope `prepare` su `select(...)` apriboja grąžinamus laukus
- [x] Scope `prepare` su palyginimu apriboja grąžinamas eilutes
- [x] User filter (`?...`) sujungiamas su scope filtru per AND — niekada per OR
- [x] User filter, kuris reikalauja daugiau, nei leidžia scope, grąžina tuščią rezultatą
- [x] Neegzistuojantis scope - klaida

### Testavimo reikalavimai

Kritiškai svarbu patikrinti, kad scope **nėra apeinama** per user'io filtrus.

- [x] Test: `GET /City/@ltu` → tik LT eilutės
- [x] Test: `GET /City/@ltu?country.code='lv'` → tuščia (scope laimi prieš user filtrą)
- [x] Test: `GET /City/@ltu?name='Vilnius'` → Vilnius (papildomas susiaurinimas leidžiamas)
- [x] Test: `GET /City/@ids` → tik `id` laukas
- [x] Test: `GET /City/@ids?select(name)` → klaida, kad Property name neegzistuoja

MindaugasStanionis:
Custom scope enforcement via `@scopeName` URL syntax
  Summary:
    - Implementing named scope enforcement for data access restriction. Scopes are defined in the manifest and applied per request to restrict what rows and fields a consumer can see.

  Manifest definition:
  | scope | ltu | | country_code='lt' |
  | scope | pub | | select(name)&country_code='lt' |
  
  URL activation:
  GET /example/City/@ltu
  GET /example/City/@pub?name="Vilnius"

  Changes
  
  - spinta/utils/url.py — @scopeName URL segment parsed as custom-scope param; serialized back in build_url_path
  - spinta/components.py — custom_scope: Optional[str] added to UrlParams
  - spinta/urlparams.py — _apply_custom_scope called after path resolution; five new private functions:
    - _extract_scope_parts — splits scope AST into (select_args, filter_exprs)
    - _collect_field_refs — recursively extracts field names from an AST expression
    - _check_field_access — blocks user filter/sort on fields hidden by scope select
    - _apply_scope_select — intersects user select with scope-allowed fields
    - _apply_custom_scope — orchestrates the above
  - spinta/dimensions/scope/helpers.py — get_active_custom_scope(model, params) resolves the active scope by name
  - tests/test_urlparams.py — 25 unit tests covering row filter, select restriction, and field access enforcement
  - tests/cli/test_run.py — 14 integration tests across SQLite, CSV, and XML backends

  Security guarantees
  
  - Scope row filters are appended after user conditions are validated — user cannot override or drop them
  - A contradicting user filter (country_code='de' against scope country_code='lt') produces zero rows, not a bypass
  - When scope defines select(...), user filter/sort on hidden fields raises PropertyNotFound — field existence is not revealed
  - Sorting by a hidden field is also blocked (prevents value inference by ordering)

### DIFF
--- a/spinta/components.py
+++ b/spinta/components.py
@@ -929,6 +929,8 @@ class UrlParams:
     select_props: Optional[Dict[str, Union[Expr, Bind]]] = None
     select_funcs: Optional[Dict[str, FuncProperty]] = None
 
+    custom_scope: Optional[str] = None
+
     sort: List[dict] = None
     limit: Optional[int] = None
     offset: Optional[int] = None
--- a/spinta/components.py
+++ b/spinta/components.py
@@ -929,7 +929,7 @@ class UrlParams:
     select_props: Optional[Dict[str, Union[Expr, Bind]]] = None
     select_funcs: Optional[Dict[str, FuncProperty]] = None
 
-    custom_scope: Optional[str] = None
+    custom_scope: str | None = None
 
     sort: List[dict] = None
     limit: Optional[int] = None
--- a/spinta/urlparams.py
+++ b/spinta/urlparams.py
@@ -381,11 +381,10 @@ def _apply_scope_select(params: UrlParams, allowed_args: list[dict]) -> None:
         return
 
     allowed = {spyna.unparse(allowed) for allowed in allowed_args}
-    intersection = [select for select in params.select if spyna.unparse(select) in allowed]
-    if not intersection:
-        first_requested = spyna.unparse(params.select[0])
-        raise exceptions.PropertyNotFound(property=first_requested)
-    params.select = intersection
+    for select in params.select:
+        name = spyna.unparse(select)
+        if name not in allowed:
+            raise exceptions.PropertyNotFound(property=name)
 
 
 def _apply_custom_scope(params: UrlParams) -> None:
--- a/spinta/urlparams.py
+++ b/spinta/urlparams.py
@@ -1,8 +1,7 @@
 import re
 import urllib.parse
 from collections import OrderedDict
-from typing import List
-from typing import Union
+from typing import Any, List, Union
 
 from starlette.requests import Request
 
@@ -18,6 +17,7 @@ from spinta.components import Model
 from spinta.components import Namespace
 from spinta.components import UrlParams, Version
 from spinta.core.ufuncs import Bind, Expr, asttoexpr
+from spinta.dimensions.scope.helpers import get_active_custom_scope
 from spinta.exceptions import ModelNotFound, InvalidPageParameterCount, InvalidPageKey
 from spinta.manifests.components import Manifest
 from spinta.ufuncs.querybuilder.components import QueryBuilder
@@ -84,6 +84,8 @@ def parse_url_query(query):
 def prepare_urlparams(context: Context, params: UrlParams, request: Request):
     _prepare_urlparams_from_path(params)
     _resolve_path(context, params)
+    _apply_custom_scope(params)
+
     params.format = get_response_type(context, request, params)
     params.accept_langs = get_preferred_accept_lang(request, params)
     params.content_langs = get_preferred_content_lang(request, params)
@@ -256,6 +258,18 @@ def _prepare_urlparams_from_path(params: UrlParams):
                 params.lang = args
             else:
                 params.lang += args
+        elif name == "custom-scope":
+            if params.custom_scope is not None:
+                raise exceptions.InvalidValue(
+                    operator="custom_scope",
+                    message="Custom scope can be set only once per request.",
+                )
+            if len(args) != 1:
+                raise exceptions.InvalidValue(
+                    operator="custom_scope",
+                    message="Custom scope expects exactly one name.",
+                )
+            params.custom_scope = args[0]
         else:
             if params.query is None:
                 params.query = []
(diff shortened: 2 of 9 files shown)


spinta/components.py
```diff
@@ -929,6 +927,8 @@ class UrlParams:
     select_props: Optional[Dict[str, Union[Expr, Bind]]] = None
     select_funcs: Optional[Dict[str, FuncProperty]] = None
 
+    custom_scope: Optional[str] = None
```
Struchid: We can already start introducing the built-in typehints `str | None`, even though the whole file seems to be using the pre Python3.10 type-hinting;

spinta/urlparams.py
```diff
@@ -321,6 +335,74 @@ def _resolve_path(context: Context, params: UrlParams) -> None:
...
+
+    if params.sort:
+        for sort_item in params.sort:
+            for field in _collect_field_refs(sort_item) - allowed:
+                raise exceptions.PropertyNotFound(property=field)
+
+    if params.query:
+        for condition in params.query:
+            for field in _collect_field_refs(condition) - allowed:
+                raise exceptions.PropertyNotFound(property=field)
```
Struchid: Would it be worth it to collect all the properties that are not allowed first and only throw if that collection has any items?

spinta/urlparams.py
```diff
@@ -321,6 +335,74 @@ def _resolve_path(context: Context, params: UrlParams) -> None:
...
+
+
+def _apply_scope_select(params: UrlParams, allowed_args: list[dict]) -> None:
+    if params.select is None:
+        params.select = list(allowed_args)
+        return
+
+    allowed = {spyna.unparse(allowed) for allowed in allowed_args}
+    intersection = [select for select in params.select if spyna.unparse(select) in allowed]
```
Struchid: There is a function for that if the naming is right:
- https://www.w3schools.com/python/ref_set_intersection.asp

Struchid (approved):

Gustascct (approved):
🔥

martynasracct (approved):

spinta/urlparams.py
MindaugasStanionis: Thanks for suggestion! But set.intersection() would return strings, and params.select needs AST objects. The list comprehension converts each AST object to a string just for the comparison, but keeps the original AST object in the result.

spinta/urlparams.py
MindaugasStanionis: That would be a good approach — collecting all invalid properties and passing them to an ErrorHandler.  However, currently ErrorHandler is only implemented for the CLI (CLIErrorHandler). The server doesn't have this functionality yet, so this would require additional work beyond the scope of this PR. Or did you have another approach in mind?

spinta/components.py
MindaugasStanionis: added it in 14e7f63270a88c72a036d466441c55ac5b1ac94a, thanks!

spinta/urlparams.py
Struchid: I was rather thinking to have some sort of collection instantiated at the top of the cycle (e.g.`fields_not_found = [ ]` and then throwing an error for all of them using something like `exceptions.PropertiesNotFound(properties=fields_not_found)`.

Definitely do not try to implement the server-level ErrorHandler :-)

MindaugasStanionis: Thanks! Added in d6e6fefcd37bbe4297471e000e9de4b34079d4a5

tests/cli/test_run.py
```diff
@@ -124,3 +124,726 @@ def test_undeclared_ref_run_does_not_fail_xml(rc: RawConfig, tmp_path: Path):
...
+            {"NAME": "Vilnius", "COUNTRY_CODE": "lt"},
+        ],
+    )
+
+    app = create_client(rc, tmp_path, sqlite)
+    app.authmodel("example/City", ["getall", "search"])
+
+    # User asks for a field entirely outside the scope — looks like the property doesn't exist.
+    resp = app.get("/example/City/@pub?select(country_code)")
+    assert resp.status_code == 404
```
JustinasKen, line 843: Would be good to also check which error is raised:
```python
assert error(resp, status=404) == "..."
```

tests/cli/test_run.py
```diff
@@ -124,3 +124,726 @@ def test_undeclared_ref_run_does_not_fail_xml(rc: RawConfig, tmp_path: Path):
...
+    app = create_client(rc, tmp_path, sqlite)
+    app.authmodel("example/City", ["getall", "search"])
+
+    # Sorting by an allowed field is fine
+    resp = app.get("/example/City/@pub?sort(name)")
+    assert resp.status_code == 200
+
+    # Sorting by a hidden field must be rejected
+    resp = app.get("/example/City/@pub?sort(country_code)")
+    assert resp.status_code == 404
```
JustinasKen, line 874: Would be good to also check which error is raised:
```python
assert error(resp, status=404) == "..."
```

tests/cli/test_run.py
```diff
@@ -124,3 +124,726 @@ def test_undeclared_ref_run_does_not_fail_xml(rc: RawConfig, tmp_path: Path):
...
+    app = create_client(rc, tmp_path, sqlite)
+    app.authmodel("example/City", ["getall", "search"])
+
+    # Filtering by an allowed field is fine
+    resp = app.get("/example/City/@pub?name='Vilnius'")
+    assert resp.status_code == 200
+
+    # Filtering by a hidden field must be rejected
+    resp = app.get("/example/City/@pub?country_code='lt'")
+    assert resp.status_code == 404
```
JustinasKen, line 911: Would be good to also check which error is raised:
```python
assert error(resp, status=404) == "..."
```

tests/cli/test_run.py
```diff
@@ -124,3 +124,726 @@ def test_undeclared_ref_run_does_not_fail_xml(rc: RawConfig, tmp_path: Path):
...
+
+    app = create_client(rc, tmp_path, sqlite)
+    app.authmodel("example/City", ["getall", "search"])
+
+    # User requests country.code alongside the scope-allowed fields — blocked
+    resp = app.get("/example/City/@public?select(name,country.name,country.code)")
+    assert resp.status_code == 200
+    data = resp.json()["_data"]
+    assert sorted(row["name"] for row in data) == ["Kaunas", "Vilnius"]
+    assert all("code" not in row.get("country", {}) for row in data)
```
JustinasKen: I saw in other tests you raise an error, when you user requests properties, that are not part of the scope, is there a reason why `ref` values did not (`country.code`) ?

Probably related to:
https://github.com/atviriduomenys/spinta/pull/1949#discussion_r3341298532

spinta/urlparams.py
```diff
@@ -321,6 +335,76 @@ def _resolve_path(context: Context, params: UrlParams) -> None:
...
+    if not intersection:
+        first_requested = spyna.unparse(params.select[0])
+        raise exceptions.PropertyNotFound(property=first_requested)
```
JustinasKen: Don't know exact requirements for the task, but shouldn't you check if there is any `select` property from user given select that falls outside allowed properties, instead of if all selected properties do not match ?

Currently if we try to select atleast one property that does not exist (or we don't have access to), we get `PropertyNotFound` error

JustinasKen (approved):

spinta/urlparams.py
MindaugasStanionis: Good catch. Changed _apply_scope_select to raise `PropertyNotFound` on any user-selected field that falls outside the scope's allowed fields, instead of only raising when none of them match. Added in 8685de2d89aca5712686eb9ff451137dae15127d

tests/cli/test_run.py
MindaugasStanionis: Yes, fixed the test in Added in https://github.com/atviriduomenys/spinta/commit/8685de2d89aca5712686eb9ff451137dae15127d

PR #1958, opened 2026-06-01
1901 remove implicit ref filters sql

### LINKED ISSUE #1901: Too many tables; MySQL can only use 61 tables in a join
DSA: [https://github.com/ivpk/metadata/blob/main/datasets/gov/lrvk/ppvis/strateginiu_projektu_duomenys/dsa.csv](https://github.com/ivpk/metadata/blob/main/datasets/gov/lrvk/ppvis/strateginiu_projektu_duomenys/dsa.csv)

Bandant gauti Communication modelio sąrašą gauname tokią klaidą: 1116, 'Too many tables; MySQL can only use 61 tables in a join'

Problema tikriausiai su `get_ref_filters`, nes bandant išparsinti visus filtrus, kurie nubanguoja per `ref` į viršų, bandoma sujoininti daug lentelių. 

Konkrečiu atveju, prie kiekvieno modelio sudėta `deleted_at` filtrai, taigi imant `Communication` modelį, tikrinami ne tik jo, bet ir modelių, į kuriuos jame yra ref'ai. Pavyzdžiui, `Project` modelis turi filtrą `deleted_at=null`, tai `Communication` modelis filtruojamas ir taip, kad rodytų tik tuos Communication įrašus, kurių Project `deleted_at=null`. Taip, bandant surinkti šiuos filtrus, esant dideliam ref'ų kiekiui, gaunamas JOIN su daug lentelių.

Galbūt reikėtų nuimti šituos "nubanguojančius", paveldimus filtrus, ir padaryti, kad tokiu atveju prie paties modelio reikėtų nurodyti ir filtravimą pagal referuojamų modelių savybes. 

## Susiję:

- https://github.com/ivpk/dsa/issues/121
- https://github.com/atviriduomenys/spinta/issues/1788
- https://github.com/atviriduomenys/spinta/issues/104
- https://github.com/atviriduomenys/spinta/issues/1659
- https://github.com/atviriduomenys/spinta/issues/1680

karina-klinkeviciute:


martynasracct (approved):

Struchid (approved):

spinta/datasets/backends/sql/commands/cast.py
```diff
@@ -87,6 +88,13 @@ def cast_backend_to_python(context: Context, dtype: Ref, backend: Sql, data: dic
                 model_name=dtype.model.name,
                 values=encoding_values,
             )
+        elif not context.get("config").check_ref_filters:
+            # When implicit ref filters are disabled, the referenced row might be
+            # filtered out at the source (e.g. by an explicit filter on the ref
+            # model), so selecting it could find nothing. The `_id` of a keymap
+            # based model is deterministic, so encode it directly instead of
+            # requiring the row to be selectable.
+            id_value = keymap.encode(keymap_name, encoding_values)
```
JustinasKen: There is one concern, some time ago, it was done similarly, by using `keymap.encode` on ref values, but that caused issues with data integrity and keymap value duplications.

This is especially relevant to scenarios where `ref` is not controlled by the same organisation (since they have to synchronize keymaps, otherwise generated values will not match).

I guess this is also raises logical question, can child property create parent's `_id` value ?

For that reason we used `select` to determine if that value combination is valid, which in turn will also generate `_id` value from parent's perspective and not children.

Now in case we cannot access these values (for example, `ref` model is not controlled by the organization) and we actually need the `_id` we should raise an error telling that `spinta keymap sync` needs to be ran in order to fetch latest `_id` values for referenced models.

If this `config` value disables all `_id` ref checks, meaning they are not used anywhere, than it might be better to set `id_value = None` or something like that (but I don't know what exactly this tasks does).

JPriev (approved):

JustinasKen (approved):

spinta/datasets/backends/sql/commands/cast.py
JustinasKen: Reason I am saying this, in case we actually need to reenable `_id` ref checks, `keymap.encode` might corrupt the `keymap` and cause issues untill `keymap sync` is ran.

karina-klinkeviciute: yep, this looks like a problem. 

in agents mode, I don't think that `keymap sync` will work. We don't have a central keymap store there, like in Saugykla. Also, data shouldn't even be accessed by an outside agent, unless it's a valid data call, coming from the end user. 

This task removes implicit filtering. I added a link to the task: https://github.com/atviriduomenys/spinta/issues/1901 

I think, in Saugykla we will use this implicit filtering for a while. 

on the other hand, a task was implemented, which allowed us to ditch the keymap on the agents, but it requires changes to DSA's, prepared by partners. i'm not sure, why we still are not making these changes. Then we wouldn't need to care about keymap at all.

This discussion was discussed with @sirex but I will raise this question tomorrow.

karina-klinkeviciute: it was discussed with Mantas and it was decided to leave it this way. This won't be an issue in and with Saugykla this is an acceptable risk.

Gustascct (approved):

JustinasKen (approved):

Struchid (approved):

Gustascct (approved):

PR #1962, opened 2026-06-02
1945 neveikia spinta push po atnaujinimo

### LINKED ISSUE #1945: Neveikia `spinta push` po atnaujinimo iš 0.1.23 į 0.1.85
Vykdanto `spinta push` komandą, gaunamos tokios klaidos:

```
spinta.cli.push - ERROR - Error when sending and receiving data. Model datasets/gov/lst/standards/OrganisationUnitRelation, items in chunk: 4054, first item in chunk:
{
    '_id': 'b0f0578e-7e30-40d4-aa42-c65b9403571b',
    '_op': 'insert',
    '_type': 'datasets/gov/lst/standards/OrganisationUnitRelation',
    'foreign_tc': {'_id': '22ecce9b-f2a4-4c3c-b0ce-f5ebe6b668c8'},
    'organisation_unit': {'_id': '055cc8b6-eebb-4052-bb9d-51026c521995'},
}
Server response (status=400):
    {
        'errors': [
            {
                'code': 'UniqueConstraint',
                'context': {},
                'message': 'Given value ([UNKNOWN]) already exists.',
                'template': 'Given value ({value}) already exists.',
                'type': 'system',
            },
        ],
    }
```

Ta pati klaida pasireiškia atsinaujinus tiek į 0.1.85, tiek į 0.2.dev24.

Paleidus:

```sh
spinta upgrade
spinta push manifest.xlsx -o put.data.gov.lt --sync
```

situacija nepasikeitė, gaunama ta pati klaida.

JustinasKen:
* Added `model.id_prop` and `model.revision_prop` for easier access (no need to always call `model.properties["_id"]` (this part could be changed, maybe it could be moved to something `model.reserved.id`, etc.).
* Changed `keymap` to contain `copy()`, which would return reset copy of the object (this is needed for forking context, since before it would reuse same object, which would fail connection assertions and error).
* Added `model_requires_source` to backends, to easily check if specific backend requires models to contain source (internal backends don't, while most external backends do (but there are cases like `CSV` or `Memory`, where `source` is optional or not needed)).
* Homogenized how external `dask` and `sql` backends handle `_id` (`Ref` and `PrimaryKey` properties).
  * Added universal `_id` handling pattern with `ExternalIdPattern` class (could be moved out from enum if needed).
  * Added improved `PrimaryKey` handling.
  * Copied over `sql` `Ref` and `PrimaryKey` handling logic to `dask` backend.
  * Changed `PrimaryKey` handling step to happen inside `cast_backend_to_python` function.
  * Moved most common `cast_backend_to_python` handling functions for `Ref` and `PrimaryKey` from `sql` backend to `ExternalBackend` (which means it would also work for `dask`).
* Fixed `_id` handling inside `cast_backend_to_python` (it used to not cast the values before encoding them).
* Added fallback to `_id` generation using select function, which would disable certain `authorized` checks when original request did not have permissions (since the select `_id` request is called by the system and not the user).
* Added a error when trying to use `Ref` property as primary key and it contains multiple additiona properties (need confirmation how exactly should that case be handled, untill then it wont be allowed).

Still need to add:
* CHANGES.rst
* Might want to add `push` test to replicate original issue.


spinta/datasets/backends/helpers.py
```diff
@@ -1,156 +1,60 @@
...
+        has_permission = True
+    # Check if user has getall permission
+    elif authorized(context, ref_model.id_prop, Action.GETALL):
+        has_permission = True
+        select_query = "select()"
+
+    expr_parts = [select_query]
+    for prop in dtype.refprops:
+        expr_parts.append(f'{prop.place}="{data[prop.place]}"')
+    expr = asttoexpr(spyna.parse("&".join(expr_parts)))
```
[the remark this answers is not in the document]
JustinasKen: This may be true, but currently `external` backends don't really support nesting that well and this entire functions is being planned to be removed, so it should be safe to keep it like this for now.

spinta/datasets/helpers.py
```diff
@@ -216,3 +218,98 @@ def decode_id_value(id_prop: Property, value: str | list) -> list:
...
+    scope_formatter: ScopeFormatterFunc = None,
+) -> bool:
+    # Bypass authorization checks for system-generated requests.
+    # TODO: Remove when `_id` generation no longer needs to select from another model.
+
+    system_request = False
+    if context.has("request.system"):
+        system_request = context.get("request.system")
+
+    if system_request:
```
karina-klinkeviciute, line 308: - High: System _id lookup bypasses auth with interpolated query values
    spinta/datasets/backends/helpers.py:41 builds an RQL string by inserting backend/ref values directly, then spinta/datasets/backends/helpers.py:48 reruns the lookup under request.system=True when the caller lacks _id
    permission. That flag is honored as a blanket auth bypass in spinta/datasets/helpers.py:312, including property selection/search checks in the SQL and Dask query builders. A crafted ref value containing RQL syntax,
    for example a quote plus another predicate, can change the generated query while it runs with system permissions. Even without injection, this now exposes referenced model _id resolution through a path the caller was
    not authorized to query before. Build the query as an Expr/AST with literal values instead of parsing an interpolated string, and keep the system lookup narrowly scoped or require an explicit, documented permission
    for resolving hidden referenced IDs.

spinta/datasets/backends/sql/ufuncs/query/ufuncs.py
```diff
@@ -440,7 +441,9 @@ def select(env: SqlQueryBuilder, dtype: Object) -> Selected:
 def select(env: SqlQueryBuilder, dtype: Ref) -> Selected:
     prep = {}
     if not dtype.inherited:
-        prep["_id"] = Selected(prop=dtype.prop, prep=select_ref_foreign_key_properties(env, dtype))
+        prep[ExternalIdPattern.ID_KEY.value] = Selected(
```
karina-klinkeviciute, line 444: - Medium: Nested identifiable refs lose _id in responses
    spinta/datasets/backends/sql/ufuncs/query/ufuncs.py:443 and spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py:369 skip _id selection when dtype.inherited is true. The tests were updated to remove nested
    country._id / region._id, so this looks like an API behavior change: nested identifiable refs that previously included _id now only include denormalized fields. That can break clients that use nested ref IDs for
    linking or de-duplication. If this is intentional, document it as a breaking response-shape change; otherwise keep selecting _id for inherited identifiable refs when the nested ref is included.

karina-klinkeviciute:
4. spinta/datasets/backends/sql/ufuncs/query/ufuncs.py:598 (+ dataframe analog) — pkey combination resolves by name, not place                                                                                             
  for prop_name in required_pk_combination:                                                                                                                                                                                  
      prop = env.resolve_property(prop_name)                                                                                                                                                                                 
  required_keymap_properties holds tuples of prop names, but resolve_property expects a place. Same name != place latent failure as #1 — a nested combination member raises PropertyNotFound at query-build time.            
                                                                                                                                                                                                                             
  Security / altitude                                                                                                                                                                                                        
                                                                                                                                                                                                                             
  5. spinta/datasets/helpers.py:305-313 + backends/helpers.py:47-49 — ambient request.system auth bypass                                                                                                                     
  authorized_or_system_request short-circuits to True whenever the request.system context flag is set, and it replaces authorized() at ~7 select/join sites. The flag is set inside the _id_select fork, but any nested ref  
  resolution within that fork inherits it, so row-level/filter access on referenced models is bypassed transitively. The code acknowledges this ("admin token hack", TODO: Remove…) — flagging it so the escape hatch is     
  scoped to the specific capability rather than fanned out as a global boolean.                                                                                                                                              
                                                                                                                                                                                                                             
  Conventions                                                                                                                                                                                                                
                                                                                                                                                                                                                             
  6. CHANGES.rst — not updated. Verified: git diff master...HEAD --name-only includes no CHANGES.rst, violating the project convention to add a changelog entry for any code change. (No CLAUDE.md exists in-repo, so no     
  other rules apply.)                                                                                                                                                                                                        
                                                                                                                                                                                                                             
  Cleanup (lower priority)                                                                                                                                                                                                   
                                                                                                                                                                                                                             
  7. spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py:43 — testlist production ufunc. Verbatim copy of querybuilder/ufuncs.py:testlist, registered as a real query-language ufunc under a debug-looking name; the   
  duplicated literal-value logic will silently drift between the two.                                                                                                                                                        
                                                                                                                                                                                                                             
  8. spinta/datasets/commands/cast.py:196-198 — dispatch-table index to dodge ambiguity. super_ = commands.cast_backend_to_python[Context, ExternalRef, Backend, dict] hand-pins one registration; if that impl is later     
  specialized/moved this resolves to the wrong function with no error. Fix the overload signatures instead.                                                                                                                  
                                                                                                                                                                                                                             
  9. spinta/datasets/backends/dataframe/ufuncs/query/helpers.py:16-76 — near-verbatim duplicate of the SQL select_ref_foreign_key_properties / select_external_ref_foreign_key_properties; the count check and target logic  
  must now be maintained in two files.                                                                                                                                                                                       
                                                                                                                                                                                                                             
  10. spinta/utils/response.py:153,182 (+ cli/helpers/store.py, cli/pii.py) — keymap.copy() factory. Changing lambda: keymap → lambda: keymap.copy() resets conn=None, forcing a fresh connection checkout per context       
  instead of sharing one. Likely intentional for isolation, but note the per-context reconnect cost on high-request-count workloads.

spinta/datasets/helpers.py
JustinasKen: This was agreed workaround to our current `Ref` handling, it is temporary and will be removed once https://github.com/atviriduomenys/spinta/issues/1979 gets approved and implemented.

spinta/datasets/backends/sql/ufuncs/query/ufuncs.py
JustinasKen: The behavior was inconsistent across backends: `Sql` did not return the nested `ref` `_id`, while `Dask` did. After discussion with the data architects, it was decided that `Dask` should follow the `Sql` behavior.

karina-klinkeviciute:
Correctness                                                                                                                                                                                                                
                                                                                                                                                                                                                             
  1. spinta/datasets/backends/helpers.py:43 — ref-id select crashes on nested refprops                                                                                                                                       
  expr_parts.append(f'{prop.place}="{data[prop.place]}"')                                                                                                                                                                    
  master used data[prop.name]; this branch changed it to data[prop.place]. But the dict passed in (id_processed_data) is keyed by refprop .name — the query-builder helpers build prep[prop.name] = …                        
  (dataframe/ufuncs/query/helpers.py:31,37,62,75) and the object/list fallbacks use {prop.name: …} (cast.py:208,233,250). For any refprop where place != name (a pkey nested inside an object), data[prop.place] raises      
  KeyError instead of building the filter. Confirmed behavior change vs. master.                                                                                                                                             
                                                                                                                                                                                                                             
  2. spinta/datasets/backends/helpers.py:18-21 — multi-candidate guard tests the wrong variable                                                                                                                              
  for row in rows_:                                                                                                                                                                                                          
      if result_value is not None:                                                                                                                                                                                           
          raise MultiplePrimaryKeyCandidatesFound(dtype, values=data)                                                                                                                                                        
      result_value = row.get("_id")                                                                                                                                                                                          
      found_value = True                                                                                                                                                                                                     
  The "more than one match" guard checks result_value is not None, but found_value is what actually tracks "saw a row." If two matching rows both have _id is None, the guard never fires and the function silently returns  
  None — an ambiguous reference is accepted as a null id instead of raising MultiplePrimaryKeyCandidatesFound. Should gate on found_value.                                                                                   
                                                                                                                                                                                                                             
  3. spinta/datasets/helpers.py:284-294 — compare_ref_property_count regresses Array refprops                                                                                                                                
  The Array-cardinality branch is left commented out and every refprop is hardcoded to count 1:                                                                                                                              
  for prop in dtype.refprops:                                                                                                                                                                                                
      # if isinstance(prop.dtype, Array) and env:                                                                                                                                                                            
      #     ...                                                                                                                                                                                                              
      prop_count_mapping[prop.name] = 1                                                                                                                                                                                      
  master's handle_ref_key_assignment resolved env.resolve(prop.external.prepare) to get the real item count for Array-typed refprops. Now an Array refprop gets expected-count 1 → spurious GivenValueCountMissmatch (or     
  wrong-shaped data accepted). The whole body also reduces to len(dtype.refprops); the dict + commented block is dead weight (flagged by 4 angles).

spinta/datasets/backends/helpers.py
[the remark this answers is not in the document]
karina-klinkeviciute: maybe external backends don't support nesting as well as internal one, but there was work done on displaying nested objects, and it works for some of our partners, like RC with XML backends, and maybe for others, with SQL backends. 

As I understand, the fix is quite small and straightforward?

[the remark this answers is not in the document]
JustinasKen: Current suggestion would not work, since it does not handle sub dictionaries, but maybe something like this would work ?:
```python
flattened_data = flatten(data)
refprop_data_keys = set()
for prop in dtype.refprops:
    refprop_data_keys.update({key for key in flattened_data.keys() if key.startswith(prop.place)})
for refprop_key in refprop_data_keys:
    expr_parts.append(f'{refprop_key}="{flattened_data[refprop_key]}"')
```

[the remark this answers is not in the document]
karina-klinkeviciute: looks ok, let's try this

spinta/datasets/backends/helpers.py
```diff
@@ -1,156 +1,60 @@
...
-    for prop in dtype.refprops:
-        expr_parts.append(f'{prop.place}="{data[prop.name]}"')
-    expr = asttoexpr(spyna.parse("&".join(expr_parts)))
-    rows = commands.getall(context, ref_model, ref_model.backend, query=expr)
+def generate_ref_id_using_select(context: Context, dtype: Ref, data: dict) -> str | None:
+    def _find_required_value(rows_: Iterator) -> object | None:
+        found_value = False
+        result_value = None
+        for row in rows_:
+            if result_value is not None:
```
karina-klinkeviciute, line 19: Won't this raise on the first found non-None value if it's not the last one in `rows`?

spinta/datasets/helpers.py
```diff
@@ -216,3 +218,98 @@ def decode_id_value(id_prop: Property, value: str | list) -> list:
...
+    # Backwards compatibility, all nested values are converted to list values without keys
+    encoding_values = flatten_keymap_encoding_values(encoding_values)
+
+    return encoding_values
+
+
+def compare_ref_property_count(dtype: Ref, data: list) -> None:
+    prop_count_mapping = {}
+    for prop in dtype.refprops:
+        # if isinstance(prop.dtype, Array) and env:
```
karina-klinkeviciute: čia turi būti užkomentuota?

spinta/datasets/backends/sql/ufuncs/query/ufuncs.py
```diff
@@ -579,17 +582,22 @@ def select(
...
-    else:
-        result = [env.call("select", prop) for prop in pkeys]
+    result = {
+        ExternalIdPattern.ID_KEY.value: {prop.name: env.call("select", prop) for prop in pkeys},
+        ExternalIdPattern.COMBINATIONS_KEY.value: {},
+    }
+    for required_pk_combination in model.required_keymap_properties:
+        combination_result = {}
+        for prop_name in required_pk_combination:
+            prop = env.resolve_property(prop_name)
```
karina-klinkeviciute, line 598: for prop_name in required_pk_combination:                                                                                                                                                                                  
      prop = env.resolve_property(prop_name)                                                                                                                                                                                 
  required_keymap_properties holds tuples of prop names, but resolve_property expects a place. Same name != place latent failure as #1 — a nested combination member raises PropertyNotFound at query-build time.

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
```diff
@@ -18,21 +17,37 @@
...
+from spinta.types.datatype import Boolean, DataType, ExternalRef, Integer, Number, PrimaryKey, Ref
 from spinta.types.text.components import Text
 from spinta.ufuncs.components import ForeignProperty
+from spinta.ufuncs.querybuilder.helpers import process_literal_value
 from spinta.utils.data import take
 from spinta.utils.schema import NA
 
 
+@ufunc.resolver(DaskDataFrameQueryBuilder, Expr)
+def testlist(env: DaskDataFrameQueryBuilder, expr: Expr) -> tuple:
```
karina-klinkeviciute, line 43: testlist production ufunc. Verbatim copy of querybuilder/ufuncs.py:testlist, registered as a real query-language ufunc under a debug-looking name; the   
  duplicated literal-value logic will silently drift between the two.

spinta/datasets/commands/cast.py
```diff
@@ -0,0 +1,308 @@
...
+        return None
+
+    return processed_data
+
+
+@commands.cast_backend_to_python.register(Context, ExternalRef, ExternalBackend, dict)
+def cast_backend_to_python(context: Context, dtype: ExternalRef, backend: ExternalBackend, data: dict, **kwargs):
+    # Reusing global `Backend` implementation, this is needed here to remove ambiguity with dispatchers
+    # ([Context, ExternalRef, ExternalBackend, object] clashes with [Context, ExternalRef, Backend, dict])
+    super_ = commands.cast_backend_to_python[Context, ExternalRef, Backend, dict]
```
karina-klinkeviciute, line 198: dispatch-table index to dodge ambiguity. super_ = commands.cast_backend_to_python[Context, ExternalRef, Backend, dict] hand-pins one registration; if that impl is later     
  specialized/moved this resolves to the wrong function with no error. Fix the overload signatures instead.

spinta/datasets/backends/dataframe/ufuncs/query/helpers.py
```diff
@@ -0,0 +1,76 @@
...
+    DaskSelected as Selected,
+)
+from spinta.exceptions import (
+    GivenValueCountMissmatch,
+    PropertyNotPartOfRefProps,
+)
+from spinta.types.datatype import ExternalRef, Ref
+
+
+def select_ref_foreign_key_properties(
```
karina-klinkeviciute, line 16: near-verbatim duplicate of the SQL select_ref_foreign_key_properties / select_external_ref_foreign_key_properties; the count check and target logic  
  must now be maintained in two files.

spinta/utils/response.py
```diff
@@ -150,7 +150,7 @@ async def create_http_response(
             if model.keymap:
                 context.attach(
                     f"keymap.{model.keymap.name}",
-                    lambda: model.keymap,
+                    lambda: model.keymap.copy(),
```
karina-klinkeviciute, line 153: keymap.copy() factory. Changing lambda: keymap → lambda: keymap.copy() resets conn=None, forcing a fresh connection checkout per context       
  instead of sharing one. Likely intentional for isolation, but note the per-context reconnect cost on high-request-count workloads.

spinta/cli/helpers/store.py
```diff
@@ -166,4 +166,4 @@ def attach_backends(context: Context, store: Store, manifest: Manifest) -> None:
 
 def attach_keymaps(context: Context, store: Store) -> None:
     for keymap in store.keymaps.values():
-        context.attach(f"keymap.{keymap.name}", lambda: keymap)
+        context.attach(f"keymap.{keymap.name}", lambda: keymap.copy())
```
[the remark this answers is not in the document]
karina-klinkeviciute: keymap.copy() factory. Changing lambda: keymap → lambda: keymap.copy() resets conn=None, forcing a fresh connection checkout per context       
  instead of sharing one. Likely intentional for isolation, but note the per-context reconnect cost on high-request-count workloads.

spinta/cli/pii.py
```diff
@@ -220,7 +220,7 @@ def detect(
                         f"transaction.{resource.backend.name}", validate_and_return_begin, context, resource.backend
                     )
         for keymap in store.keymaps.values():
-            context.attach(f"keymap.{keymap.name}", lambda: keymap)
+            context.attach(f"keymap.{keymap.name}", lambda: keymap.copy())
```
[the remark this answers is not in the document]
karina-klinkeviciute: keymap.copy() factory. Changing lambda: keymap → lambda: keymap.copy() resets conn=None, forcing a fresh connection checkout per context
instead of sharing one. Likely intentional for isolation, but note the per-context reconnect cost on high-request-count workloads.

spinta/datasets/backends/helpers.py
JustinasKen: it shouldn't since at the start we set `result_value` to `None` and we only update this value at the end of the loop, if there are 2 rows returned, second itteration should raise an error, since multiple candidates were found.

spinta/datasets/helpers.py
JustinasKen: good catch, pamiršau atkomentuoti, keista, kad jokių errors iš testų nebuvo

spinta/datasets/backends/dataframe/ufuncs/query/ufuncs.py
JustinasKen: Its required by the way lark rules are setup, main reasaon it was copied over, since `Dask` technically speaking should use `QueryBuilder` implementation, but currently it does. Once it does, it should be removed.

spinta/utils/response.py
JustinasKen: This is mainly done, since at the end of context, conn gets closed, so if we have not separate we would have this issue:
```python
with context: # calls keymap.__enter__ which populates `conn`
    
    forked_context = context.fork("forked")
    with forked_context: # calls keymap.__enter__ again and gets an error since `conn` is already open
        # Lets assume we remove that check and allow reuse of connection
    # Closes `forked_context` which calls keymap.__exit__ which calls `keymap.conn.close()`
    
    # Call any action that requires keymap
    # Error since connection is `None` and its closed because of `forked_context`
```

spinta/datasets/helpers.py
JustinasKen: Panaikinau šitą vietą, kadangi `cast_backend_to_python` neturi prieeigos prie `QueryBuilder`, taip pat nėra nei vieno testo, kuris tikrina šitą case (kai `ref` raktas yra `array`), galimai ši funkcija aplamai yra neteisinga, kadangi neatrodo, kad gerai handlintu nested refproperty keys. Ji buvo perkelta iš seno handler ir galimai reikės perdaryti / panaikinti, jei atsiras cases, kuriom tai aktualu. https://github.com/atviriduomenys/spinta/pull/1962/changes/3d02af33edef86a4b9113906f86eef44e0fdb501

spinta/datasets/backends/sql/ufuncs/query/ufuncs.py
JustinasKen: Fixed https://github.com/atviriduomenys/spinta/pull/1962/changes/3d02af33edef86a4b9113906f86eef44e0fdb501

spinta/datasets/commands/cast.py
JustinasKen: It is required to not raise Ambiguity warnings, since we want to use
`Context, ExternalRef, Backend, dict`
But we have
`Context, ExternalRef, ExternalBackend, object`
Those two conflict, so we need explicit implementation (it would not be needed if we didnt use `object`, but it would require alot of changes).

spinta/datasets/backends/helpers.py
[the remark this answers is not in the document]
JustinasKen: Fixed https://github.com/atviriduomenys/spinta/pull/1962/changes/3d02af33edef86a4b9113906f86eef44e0fdb501

spinta/cli/helpers/store.py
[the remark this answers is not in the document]
JustinasKen: https://github.com/atviriduomenys/spinta/pull/1962#discussion_r3534357321

spinta/datasets/helpers.py
```diff
@@ -216,3 +218,94 @@ def decode_id_value(id_prop: Property, value: str | list) -> list:
...
     return decoded_value
+
+
+def extract_and_cast_properties_from_list(
+    context: Context,
+    backend: Backend,
+    model: Model,
+    keymap: KeyMap,
+    data: dict,
+    cache: dict = None,
```
JPriev:
```suggestion
    cache: dict | None = None,
```

spinta/datasets/helpers.py
```diff
@@ -216,3 +218,94 @@ def decode_id_value(id_prop: Property, value: str | list) -> list:
...
+    return result
+
+
+def process_data_for_pkey(
+    context: Context,
+    backend: Backend,
+    model: Model,
+    keymap: KeyMap,
+    data: dict,
+    cache: dict = None,
```
JPriev:
```suggestion
    cache: dict | None = None,
```

spinta/datasets/helpers.py
```diff
@@ -216,3 +218,94 @@ def decode_id_value(id_prop: Property, value: str | list) -> list:
...
+        raise GivenValueCountMissmatch(dtype, given_count=len(data), expected_count=expected_count)
+
+
+def authorized_or_system_request(
+    context: Context,
+    node: Namespace | Model | Property,
+    action: Action,
+    *,
+    throw: bool = False,
+    scope_formatter: ScopeFormatterFunc = None,
```
JPriev:
```suggestion
    scope_formatter: ScopeFormatterFunc | None = None,
```

tests/datasets/test_sql.py
```diff
@@ -4733,3 +4735,176 @@ def test_keymap_internal_model_after_sync(
         {"code": "LT", "planet": {"_id": er_id}},
         {"code": "LV", "planet": {"_id": mr_id}},
     ]
+
+
+def test_identifiable_ref_primary_key(context, rc, tmp_path, geodb_denorm, reset_keymap):
```
JPriev: missing type hints

JPriev (approved):

spinta/datasets/helpers.py
JustinasKen: Fixed https://github.com/atviriduomenys/spinta/pull/1962/changes/7b9bb57837543efc5de5c59fcf9ebf8131337986

spinta/datasets/helpers.py
JustinasKen: Fixed https://github.com/atviriduomenys/spinta/pull/1962/changes/7b9bb57837543efc5de5c59fcf9ebf8131337986

spinta/datasets/helpers.py
JustinasKen: Fixed https://github.com/atviriduomenys/spinta/pull/1962/changes/7b9bb57837543efc5de5c59fcf9ebf8131337986

tests/datasets/test_sql.py
JustinasKen: Fixed https://github.com/atviriduomenys/spinta/pull/1962/changes/7b9bb57837543efc5de5c59fcf9ebf8131337986
Didn't add all typehints, for stuff like `reset_keymap`, which is pretty much only used to automatically reset keymaps, which could be wrong (maybe there are better ways by using pytest modules).

spinta/datasets/backends/dataframe/ufuncs/query/helpers.py
JustinasKen: Yes they are pretty much copy paste, main difference is how `Selected.item` field is set, idealy `Dask` would use `QueryBuilder` and that function could be extracted to be used in `Dask` and `Sql`. If you think that these should be reworked to work for both without code duplication, I could look into it.

Issue #1030, opened 2024-12-13
Reference pointing to model without primary key


sirex:
```
m | property | type    | ref     | level | access
Country      |         |         | 0     | 
  | id       | integer |         | 4     | open
City         |         |         | 0     | 
  | country  | ref     | Country | 1     | open
```

In this case `City/country` points to `Country` model, but `Country` model does not have a primary key specified, because `model.ref` of `Country` model is empty.

In this case, since `City/country` has `level < 1`, then `City/country/_id` does not exists and can't be used in any operations. All actions and functions, that uses `_id` should be disabled.

When `Country.level < 4` and `> 0`, then `_id` still does not exist and `Country` `model.ref` user provided primary key propertties should be used, accepting whatever value comes from data source. This means, that in order to get single object a queyr like `/City/:getone?id=42` should be used. In this case all actions and functions, that uses `_id`, should used whatever is provided in `model.ref`.

### Related

- https://github.com/atviriduomenys/spinta/issues/400
- https://github.com/ivpk/dsa/issues/39
- https://github.com/atviriduomenys/spinta/issues/1159

martynasracct:
Refinement 2026-02-10:
Išsiaiškintas užduoties sprendimas iš user pusės:
* Jei `model.level` = 0, tai `_id` neturi būti grąžinamas.
* Jei `model.level` = 1, 2, 3, tai `_id` neturi būti grąžinamas.
* Jei `model.level` = 4, 5, tai `_id` turi būti grąžinamas.
* Jei `ref.level` = 0, tai laukas turi matytis response, bet reikšmė `null`.
* Jei `ref.level` = 1, 2, 3, tai turi būti grąžinamas objektas su visais laukais per kuriuos daromas `ref` jungimas.
* Jei `ref.level` = 4, 5, tai turi būti grąžinamas `_id`.

Visi veiksmai naudojantys pirminį raktą (modelio ir `ref`) (kol kas tik getone, kuris nėra įgyvendintas) turi veikti pagal tai, kaip daromas `ref`. Nebūtinai per `_id`

PR #1971, opened 2026-06-09
1970 fix to the logic of refs to models with no primary keys

### LINKED ISSUE #1970: spinta check klaida AssertionError
Paleidus `spinta check` su `metadata` DSA failais, gaunama tokia klaida. 

Kol kas neaišku, kurioje vietoje ir dėl ko ji gaunama.

Prieš tai buvo daug warnings dėl nerastų ref: WARNING - Model 'Model' referenced by 'x' property 'y' was not found in the manifest. Downgrading to object (level 2).

```python
  File "spinta/cli/config.py", line 35, in check
    prepare_manifest(context, ensure_config_dir=True, full_load=True)
  File "spinta/cli/helpers/store.py", line 142, in prepare_manifest
    commands.prepare(context, store.manifest)
  File "spinta/manifests/commands/init.py", line 14, in prepare
    commands.prepare(context, backend, manifest, **kwargs)
  File "spinta/backends/postgresql/commands/init.py", line 27, in prepare
    commands.prepare(context, backend, model, **kwargs)
  File "spinta/backends/postgresql/commands/init.py", line 44, in prepare
    column: List[Any] | Any | None = commands.prepare(context, backend, prop, **kwargs)
  File "spinta/backends/__init__.py", line 226, in prepare
    return prepare(context, backend, prop.dtype, **kwargs)
  File "spinta/backends/postgresql/types/array/init.py", line 23, in prepare
    assert columns is not None
AssertionError
```

karina-klinkeviciute:
Pradinė klaida buvo todėl, kad `backref required` būdavo neatpažįstamas kaip `backref`. 

Ji buvo pataisyta skaitymo metu `backref required` tikrinime sumažinus iki `backerf`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -48,10 +48,16 @@ Bug fixes:
 - Fixed a bug where getone method in sql backend did not get any ref data (`#1900`_)
 - Fixed `count()` function not working properly with `dask` backends (`#1950`_).
 - Fixed `eq` comparison not working properly with `dask` backends on Number, Integer, Boolean values (`#1959`_).
+- Fixed data type detection in the tabular manifest reader so that modifiers
+  such as `required` or `unique` no longer interfere with parent and nesting
+  resolution. An `array` of `backref` is now recognised as an array backref, and
+  re-declaring an `object`/`array` property with a modifier no longer raises a
+  spurious nesting error (`#1970`_).
 
 .. _#1900: https://github.com/atviriduomenys/spinta/issues/1900
 .. _#1950: https://github.com/atviriduomenys/spinta/issues/1950
 .. _#1959: https://github.com/atviriduomenys/spinta/issues/1959
+.. _#1970: https://github.com/atviriduomenys/spinta/issues/1970
 
 0.2dev25 (2026-05-22)
 =====================
--- a/spinta/manifests/tabular/helpers.py
+++ b/spinta/manifests/tabular/helpers.py
@@ -653,9 +653,20 @@ class PropertyReader(TabularReader):
     enums: Set[str]
 
     def read(self, row: Dict[str, str]) -> None:
-        complete_structure, parent_structure, prop_name = _extract_and_create_parent_data(self, row, row["property"])
+        # Parse the data type once, here at the boundary, and pass the parsed
+        # result inward. Downstream code must not re-parse `row["type"]` or
+        # compare against the raw string: modifiers (`required`, `unique`) and
+        # arguments would make a type such as `backref required` or
+        # `object required` unrecognisable. Parent and nesting resolution only
+        # needs the bare type, while `_handle_datatype` keeps the modifiers.
+        dtype = _resolve_dtype(self, row)
+        bare_row = {**row, "type": dtype["type"]}
+        complete_structure, parent_structure, prop_name = _extract_and_create_parent_data(
+            self, bare_row, bare_row["property"]
+        )
 
-        prop_data = _handle_datatype(self, row)
+        prop_data = _handle_datatype(self, row, dtype)
+
         prop_name = _combine_parent_with_prop(prop_name, prop_data, parent_structure, complete_structure)
 
         # Edge case where there is no nesting, need to couple `prop_data` with `complete_structure`
@@ -748,8 +759,11 @@ def _initial_text_property_schema(given_name: str, dtype: dict, row: dict):
     return result
 
 
-def _datatype_handler(reader: PropertyReader, row: dict, initial_data_loader: Callable[[str, dict, dict], dict]):
-    dtype: dict = _resolve_dtype(reader, row)
+def _datatype_handler(
+    reader: PropertyReader, row: dict, initial_data_loader: Callable[[str, dict, dict], dict], dtype: dict = None
+):
+    if dtype is None:
+        dtype = _resolve_dtype(reader, row)
     given_name = row["property"]
     reader.name = _clean_up_prop_name(row["property"].split(".")[-1])
 
@@ -759,7 +773,7 @@ def _datatype_handler(reader: PropertyReader, row: dict, initial_data_loader: Ca
             f"Property {reader.name!r} must be defined in a model context. "
             f"Now it is defined in {context.name!r} {context.type} context."
         )
-    _check_if_property_already_set(reader, row, given_name)
+    _check_if_property_already_set(reader, dtype["type"], given_name)
 
     if reader.state.base and not dtype["type"]:
         dtype["type"] = "inherit"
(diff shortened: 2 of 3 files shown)


JPriev (approved):

CHANGES.rst
```diff
@@ -41,10 +41,18 @@ Bug fixes:
 - Fixed a bug where getone method in sql backend did not get any ref data (`#1900`_)
 - Fixed `count()` function not working properly with `dask` backends (`#1950`_).
 - Fixed `eq` comparison not working properly with `dask` backends on Number, Integer, Boolean values (`#1959`_).
+- Fixed array element type detection so that modifiers such as `required` or
```
sirex:
```suggestion
- Fixed `array` element type detection so that modifiers such as `required` or
```

spinta/manifests/tabular/helpers.py
```diff
@@ -1013,7 +1013,8 @@ def _get_parent_data_array(reader: PropertyReader, given_row: dict, full_name: s
         else:
             raise NestedDataTypeMismatch(initial=current_type, required=DataTypeEnum.ARRAY.value)
 
-    if given_row.get("type") == DataTypeEnum.BACKREF.value:
+    # if we have modifiers, like `required` or `unique`, we need to ignore them
+    if (given_row.get("type") or "").split()[0:1] == [DataTypeEnum.BACKREF.value]:
```
sirex: If I understood correctly, the real issue is here:

https://github.com/atviriduomenys/spinta/blob/7485b530643f897969b60dea2f5b064e5de880bc/spinta/manifests/tabular/helpers.py#L653-L655

Since `_extract_and_create_parent_data` uses `row['type']` which is not yet parsed, then `_handle_datatype` must parse type before calling `_extract_and_create_parent_data`.

This potentially could fix other related issues too.

tests/backends/postgresql/commands/test_init.py
```diff
@@ -151,6 +151,52 @@ def test_prepare_model_ref_unique_constraint(rc: RawConfig):
     )
 
 
+def test_prepare_model_ref_unique_constraint_level3(rc: RawConfig):
+    # A unique constraint on a level 3 ref must reference the column that is
```
sirex:
```suggestion
    # A foreign key constraint on a level 3 ref must reference the column that is
```

?

tests/backends/postgresql/commands/test_init.py
```diff
@@ -151,6 +151,52 @@ def test_prepare_model_ref_unique_constraint(rc: RawConfig):
...
+    # actually created. When the referenced model has a primary key the ref is
```
sirex: What is "actually created"?

tests/backends/postgresql/commands/test_init.py
```diff
@@ -151,6 +151,52 @@ def test_prepare_model_ref_unique_constraint(rc: RawConfig):
...
+    # denormalised into `<prop>.<refprop>` columns, otherwise it falls back to a
+    # single `<prop>._id` foreign key column.
```
sirex: Previously if `model.ref` was not specified, then all model properties were used and primary key.

But this behavior was causing multiple issues.

So there are plans to change it:

- https://github.com/ivpk/dsa/issues/39

But I could not find any information, where it would be explained, that if `model.ref` is not given, then it falls back to `_id` as primary key.

Also there is another related issue:

- https://github.com/atviriduomenys/spinta/issues/1030

As I understand, with this change you are fixing more than one issue?

Anyway, if `property.ref` points to a model without `model.ref`, then `property.level` can't be `3`. According to [DSA:L206](https://ivpk.github.io/dsa/draft/branda.html#l206-nenurodytas-susiejimas), `property.level` can't be greater than `2`.

According to [UAPI / Maturity](https://ivpk.github.io/uapi/draft/#section/Concepts/Maturity) examples, `property.level <= 3` does not have `_id`. So in this case, `property` must provide identifiers in `NoPk[name]` between `[]` in `property.ref` column.

Otherwise, such manifest is not valid.

tests/backends/postgresql/commands/test_init.py
karina-klinkeviciute: čia komentaras buvo vienaip parašytas, tai perrašiau tiksliau. Nes foreign key, kaip suprantu, buvo traktuojama, kaip `unique`, nors išskirtinai `unique` DSA manifeste ir nėra uždėtas. 

Čia, matyt, rodo diff tarp skirtingų commit'ų?

spinta/manifests/tabular/helpers.py
karina-klinkeviciute: do you mean that I should switch those two functions? But the `_handle_datatype` function doesn't change the `row`, so the issue will still appear.

tests/backends/postgresql/commands/test_init.py
karina-klinkeviciute: > Previously if model.ref was not specified, then all model properties were used and primary key.

> But this behavior was causing multiple issues.

> So there are plans to change it:

   >  https://github.com/ivpk/dsa/issues/39
   
Which should I follow now, the all-fields as a primary key, or that the joining is impossible and `ref` to it should have `0` maturity level and shouldn't work?

spinta/manifests/tabular/helpers.py
```diff
@@ -745,8 +756,11 @@ def _initial_text_property_schema(given_name: str, dtype: dict, row: dict):
     return result
 
 
-def _datatype_handler(reader: PropertyReader, row: dict, initial_data_loader: Callable[[str, dict, dict], dict]):
-    dtype: dict = _resolve_dtype(reader, row)
+def _datatype_handler(
+    reader: PropertyReader, row: dict, initial_data_loader: Callable[[str, dict, dict], dict], dtype: dict = None
+):
```
JPriev, line 764: Missing type hints. Also repeats in other functions

spinta/manifests/tabular/helpers.py
JustinasKen: well as I understand main issue right now is that `_extract_and_create_parent_data` uses raw `row` dictionary, which has `row["type"] == "backref required"`, but `_handle_datatype` parses it with this function:
https://github.com/atviriduomenys/spinta/blob/f1f87ea61998c1262237a8bd2c32d8a6a08a236c/spinta/manifests/tabular/helpers.py#L619-L624

https://github.com/atviriduomenys/spinta/blob/f1f87ea61998c1262237a8bd2c32d8a6a08a236c/spinta/manifests/tabular/helpers.py#L535-L569

The question would be, could we use `prop_data` instead of `row`, or atleast parse `type` part before hand ?

In my honest opinion this part of the code (nesting handling) is kind of messy and difficult to maintain, since we have to rely on `dict`, random strings, string matching, etc. Idealy this nesting handle would use more dataclasses, dispatchers and would be more expandable, but this is outside the scope and just my opinion.

JustinasKen (approved):

spinta/manifests/tabular/helpers.py
karina-klinkeviciute: `_handle_datatype` needs full datatype, but it doesn't modify the row. So I changed the row that is passed to `_extract_and_create_parent_data`. If we want `_handle_datatype` to change the type inline, it's a bigger change, I think, and we would need to make sure that it doesn't break anything else

tests/backends/postgresql/commands/test_init.py
karina-klinkeviciute: I moved this part to a different branch, where i will first investigate if this is an issue in spinta or in DSA

PR #2001, opened 2026-07-31
#1990: Configuration documentation rewrite and fixed config overrides

### LINKED ISSUE #1990: Nested `config` source load
Add ability to load multiple `config` sources.

## Problem

Currently you are able to set `config` sources only in one input.
For example:

`config.yml`
```yml
config:
  - .\config_models.yml
  - .\config_citus.yml
```

When we load configs using `spinta -o config=.\config.yml ...` or using `SPINTA_CONFIG` env argument, only config source that gets loaded is `config.yml`, even though inside we set additional `config_models.yml` and `config_citus,yml`.

## Solution

This can be fixed by introducing recursive / nested loading here:
https://github.com/atviriduomenys/spinta/blob/34e73cfd2bc24fe8f8aa6153ba331064fe3b0a4a/spinta/core/config.py#L46-L52

To something like this:
```python
def read_config(args=None, envfile=None):
    rc = RawConfig()
    rc.read(
        [
            Path("spinta", "spinta.config:CONFIG"),
            EnvFile("envfile", envfile or ".env"),
            EnvVars("envvars", os.environ),
            CliArgs("cliargs", args or []),
        ]
    )

    # Inject extension provided defaults
    _read_chained_configs(rc)
    return rc


def _read_chained_configs(rc: RawConfig) -> None:
    loaded: set[str] = set()

    while True:
        configs, origin = rc.get("config", cast=list, default=[], origin=True)
        new_configs = []
        for config in configs:
            if config not in loaded:
                loaded.add(config)
                new_configs.append(config)

        if not new_configs:
            break

        rc.read(
            [Path(config, config) for config in new_configs],
            after=origin,
        )
```

There is another problematic point that raises with this solution and that is `object` concatination.
Currently once new key is found, it gets replaced:
https://github.com/atviriduomenys/spinta/blob/34e73cfd2bc24fe8f8aa6153ba331064fe3b0a4a/spinta/core/config.py#L360-L368

If we were to update it to:
```python
k = tuple(key[:i])
v = config.get(k, env)
if v is not NA:
    # Source has explicit value set.
    if isinstance(v, str):
        v = [x.strip() for x in v.split(",")]
    else:
        v = list(v)

    if k not in keys:
        keys[k] = config, v

    existing = keys[k][1]
    for item in v:
        if item not in existing:
            existing.append(item)

    keys[k] = config, existing
```
This would allow us to separate configuration files and still modify same objects.

## Use case
These changes are mainly needed to better support / maintain these features:
https://github.com/atviriduomenys/spinta/issues/1691 and https://github.com/atviriduomenys/spinta/issues/599

#599 Added a feature to overwrite model's property type (for example to `BigInteger` intead of `integer`). It would look something like:
```yaml
models:
  datasets/gov/example/City:
    properties:
      population:
        type: sqlalchemy.types.BigInteger
      history:
        type:
          name: sqlalchemy.types.String
          lenght: 200
```
For most cases, this is fairly rare requirement (it would touch maybe 20 models), so it could stay inside main `config.yml`, but with the introduction to #1691, `models` field now also supports `distribute` field that allows changing distribution types for specific models:
```yaml
models:
  datasets/gov/example/City:
    distribute: copy
```
And there is a posiblity that there might be over 200 models that will need these changes. So storing them inside main config.yml file together with #599 changes would be difficult to maintain and it could cause bugs / accidental changes.

With this change `config.yml` could set `config_models.yml` and `config_citus.yml` files paths and they would look:
`config_models.yml`
```yaml
models:
  datasets/gov/example/City:
    properties:
      population:
        type: sqlalchemy.types.BigInteger
      history:
        type:
          name: sqlalchemy.types.String
          lenght: 200
```
`config_citus.yml`
```yaml
models:
  datasets/gov/example/City:
    distribute: copy
```
When running `spinta config models` output would be something like:
Origin  |  Name |  Value                      
-------- | ------- | -------
.\var\config_citus.yml | models.datasets/gov/example/City.distribute | copy                       
.\var\config_models.yml | models.datasets/gov/example/City.properties.population.type | sqlalchemy.types.BigInteger
.\var\config_models.yml | models.datasets/gov/example/City.properties.history.type.name | sqlalchemy.types.String
.\var\config_models.yml | models.datasets/gov/example/City.properties.history.type.length | 200

instead of what currently returns:
Origin  |  Name |  Value                      
-------- | ------- | -------          
.\var\config_models.yml | models.datasets/gov/example/City.properties.population.type | sqlalchemy.types.BigInteger
.\var\config_models.yml | models.datasets/gov/example/City.properties.history.type.name | sqlalchemy.types.String
.\var\config_models.yml | models.datasets/gov/example/City.properties.history.type.length | 200

Which would be alot simpler to maintain when we need to change specific property types, it would also allow us to maintain `citus` distribution types using scripts instead of manually changing configuration (which could cause bugs), since we could just keep recreating new config files that would only include citus distribution changes.

sirex:
This change covers three related areas:

1. **Configuration override semantics** — documents and tests how multiple configuration sources override each other, and fixes bugs where keys removed by a higher-priority source were still accessible and forked configuration shadowed lower-priority sources.
2. **RawConfig rework** — simplifies `RawConfig` internals to build the whole effective configuration tree up front, prevents mutation of the global `spinta.config:CONFIG` dict, validates values of object-type keys, and adds type annotations.
3. **Configuration documentation** — rewrites the configuration manual (source precedence order, configuration syntax, `config_path`, recommended `config.yaml` location) and adds a keymap configuration page.

### Configuration overrides

Spinta can be configured from multiple sources: default `spinta.config:CONFIG`, configuration files, `.env` files, environment variables and command line arguments. Each subsequent source overrides the previous ones. This behavior was implemented but not documented and not tested.

Two source types behave differently:

- **Complex values** (dicts), supported only by configuration files and other `pydict`/`Path` sources. A complex value on a parent key replaces the whole subtree of that key.
- **Simple values** (scalars), supported by environment variables, `.env` files and command line arguments. Here parent keys hold lists of child names, which control the structure of the subtree, while only leaf values are set.

New feature: setting a parent key to an empty value (e.g. `SPINTA_BACKENDS=`) resets that subtree, so it can be rebuilt from scratch. Without this there was no way to remove keys that were set by a lower-priority source.

**Bug fixed:** when keys were removed this way, `rc.keys(...)` correctly showed them as gone, but `rc.get(...)` still returned their stale values from lower-priority sources. Removed keys must be unavailable at every level, not just hidden from key listing.

### Fork merging

`RawConfig.fork()` (used throughout the codebase and tests to override configuration per request) often contains only a partial configuration, for example a `keymaps.default` with only `sync_transaction_size` set. Previously such a fork replaced the whole `keymaps.default` subtree and dropped values from lower-priority sources (e.g. `type` or `dsn`). Forks are now flattened into dotted leaf keys and merged with lower-priority sources, while explicitly set values still replace lower-priority ones. Nested dict values in forks are flattened verbatim, so keys that contain dots themselves (e.g. denorm property names like `test.name`) keep working.

### Scalar values for object-type keys

Keys with an object-type schema (e.g. `backends`) hold a list of child key names. Setting a scalar value, e.g. `{'backends': 'one'}`, is now an error in configuration files, `rc.add()` and `RawConfig.fork()`; a list of key names (`{'backends': ['one']}`) or a mapping declaring the whole subtree must be used instead.

Sources that only support scalar values (environment variables, `.env` files, CLI args) still accept comma separated key names, e.g. `SPINTA_BACKENDS=one,two`, which are parsed into lists when the source is read, so `RawConfig` always receives object key values as lists. `rc.get()` on an object-type key now returns the resolved list of child key names (`[]` when reset) instead of the raw scalar value.

Two tests using scalar values for object-type keys (`distribute: "copy"`) were updated to use lists (`distribute: ["copy"]`).

### Global CONFIG mutation

`PyDict.read()` popped `environments` from the given dict without copying it. The global `spinta.config:CONFIG` dict can be imported by multiple `RawConfig` instances, so the first read removed its `environments` section, making environment specific configuration (e.g. test backend DSNs) unavailable in later configuration reads and long running processes. The input dict is now copied before modification.

### RawConfig internals rework

- Simplified internals: the `after` argument of `RawConfig.read()` and `RawConfig.fork()` was removed. Sources argument order decides priority. `read_config()` pre-reads sources to discover config files set via the `config` option, then reads everything in the effective order: `spinta -> config files -> envfile -> envvars -> cliargs`.
- `_keys` now holds the whole effective configuration tree as `KeyNode`s: values from all sources, `environments.<env>.*` overlays of the active environment (prefix dropped) and schema structure with defaults. Read methods (`get`, `getall`, `keys`, `has`, ...) read only this tree and no longer interpret `environments.*` or the configuration schema at read time.
- `environments.*` keys are no longer accessible via `rc.get()`; they are merged into the main tree for the active environment. As a side effect, `rc.get()` on an unset option under a dynamic subtree (e.g. `rc.get("keymaps", "default")` with no keymaps configured) now returns `NA` instead of an empty list or the schema default.
- `spinta config` output has a new `Env` column showing which environment overlay provided each value.
- Type annotations added to `RawConfig`.

### Documentation

The configuration manual had several inaccuracies:

- The source precedence order was listed highest-to-lowest and was wrong (config files read *after* `.env`).
- `config_path` was described as "an additional configuration source", but it is not a source of configuration values — it is a directory for auxiliary data (client credentials, auth keys, keymaps).
- The recommended location for the main configuration file (`{config_path}/config.yaml`) was not documented.
- Reset/removal of configuration keys was not documented.

### `spinta/core/config.py`

- `read_config`: pre-reads all sources to discover additional config files, then reads everything in the effective source order (`spinta -> config files -> envfile -> envvars -> cliargs`).
- `read()` / `fork()`: the `after` argument was removed; sources argument order decides priority.
- `fork()`: new `_flatten_fork_config` helper flattens nested fork dicts into dotted leaf keys, so partial forks merge with lower-priority sources instead of replacing whole subtrees. Scalar values on object-type keys raise an error, while explicit list values (e.g. `{"backends": ["one"]}`) still reset the subtree.
- `KeyNode` / `ConfigKeys`: `_keys` is now a mapping of key paths to `KeyNode`s (resolved type, source, env overlay, value, explicit flag) holding the whole effective configuration tree, rebuilt by `_rebuild()` on every `read()`/`fork()`.
- `_rebuild` / `_process_source` / `_process_key` / `_ensure_union`: process sources in priority order (parents first), applying each source's `environments.<env>.*` overlay for the active environment on top of its base values, then `_add_schema_nodes` merges schema structure and defaults (schema-derived nodes are findable via `get()`, but invisible to `keys()`/`getall()`).
- `get()`: reads only from `_keys`; object-type keys return the resolved list of child key names; a new `_key_exists()` check makes keys removed by higher-priority sources unavailable, while auto-derived union parents never hide anything.
- `keys()`: reads child-name lists from `_keys`, distinguishing object-type and union parent nodes from leaf values that happen to be lists.
- `dump()`: new `Env` column showing which environment overlay provided each value.
- `PyDict.read()`: copies the input dict before popping `environments`, so shared dicts like the global `spinta.config:CONFIG` are never modified.
- `_check_keys`: raises an error when a scalar value is set for an object-type key in sources that support complex values (`PyDict`/`Path`).
- `_parse_object_key_values`: parses comma separated strings into lists for object-type keys in `EnvVars`, `EnvFile` and `CliArgs`.
- Type annotations for `RawConfig` and helpers; `ConfigSource.config` defaults to an empty dict.

### `spinta/types/model.py`

- Custom property `type` can now be given as a dict (parameterized custom type), in addition to a list.

### Tests

- `tests/test_config_overrides.py` (new): parametrized tests covering all 9 base/override source-type pairs (`pydict`, `envvars`, `cliargs`):
  - `test_override_backend_values` — overriding leaf values on an existing key.
  - `test_add_backend` — adding a new backend; a pydict parent value replaces the whole subtree, while simple values merge.
  - `test_reset_backends` — `SPINTA_BACKENDS=` + nested keys rebuilds the subtree; removed keys are `NA`.
  - `test_remove_all_backends` — `SPINTA_BACKENDS=` alone removes all keys recursively at every level.
  - `test_remove_backend` — `SPINTA_BACKENDS=one` keeps only listed keys; removed keys are `NA`.
  - `test_pydict_dotted_keys` — pydict sources behave like simple sources when given dotted keys.
  - `test_pydict_scalar_value_for_keys_raises_error` — scalar values for object-type keys are an error.

  The `_make_config` helper builds the same configuration for all three source types.
- `tests/test_config.py`: updated for the removed `after` argument, the new `Env` column of `dump()` and `environments.*` no longer being accessible via `get()`; new tests: `test_pydict_read_does_not_modify_input`, `test_path_read_does_not_modify_imported_config`, `test_fork_partial_subtree_keeps_lower_priority_keys`, `test_fork_adds_new_keys_to_lower_priority_ones`, `test_fork_explicit_value_still_resets_lower_priority_keys`, `test_fork_scalar_value_for_keys_raises_error`, `test_dump_env_column`.
- `tests/types/test_model.py`: `test_configure_custom_type_with_parameters` — custom property type given as a dict with parameters.
- `tests/cli/admin/test_citus_distribution.py` and `tests/backends/postgresql/commands/migrate/test_citus.py`: `distribute: "copy"` updated to `distribute: ["copy"]` per the new object-key validation.

### Documentation

- `docs/en/manual/configuration/index.rst`: rewritten — new "Quick start" section with a full configuration example and `spinta config` output; "Inspecting configuration" (prefix filters, `-f env`); sources listed in read order with correct precedence; new "Configuration syntax" section (simple vs complex values, merging vs replacing, recursive key removal via empty values, YAML lists for complex sources); `config_path` clarified as a directory for auxiliary data with its contents listed; recommended location of the main config file `{config_path}/config.yaml`.
- `docs/en/manual/configuration/keymap.rst` (new): keymap configuration (SQLAlchemy and Redis backends, Redis persistence notes), linked from the manifest configuration page.
- `CHANGES.rst`: entries under 1.2.0 (unreleased) — backwards incompatible changes, bug fixes and improvements, with `#1990`/`#2001` links.

## Backwards incompatible changes

- The `after` argument of `RawConfig.read()` and `RawConfig.fork()` was removed; sources argument order decides priority.
- `environments.*` keys are no longer accessible via `rc.get()`.
- `rc.get()` on an unset option under a dynamic subtree returns `NA` instead of an empty list or the schema default value.
- `rc.get()` on object-type keys returns the resolved list of child key names.
- Scalar values for object-type keys raise errors in sources that support complex values; env/CLI sources parse comma separated lists into key lists.

## Verification

Manually verified the rewritten config (`#2001`) against real-world flows:

- **Doc use cases (`vadovas/spinta.rst`)** — all 11 scenarios pass: `inspect`/`show`/`copy`/`check` (xlsx, csv, sqlite, XML/XSD via dask), `init` round-trip, config.yml loading (CLI `-o` and `SPINTA_CONFIG`), `run --mode external` serving over HTTP, `comment`/`uncomment` missing-external-refs, `admin add/remove_local_ids`.
- **server.sh flow** — isolated BASEDIR: `key generate`, `upgrade clients`, `client add`, `config backends manifests`, `check`/`wait`/`bootstrap` (tables + changelog verified), serving on :8001, client.sh token auth (basic → Bearer), ns tree rendering, anonymous access, config layering/override semantics (envvars > `.env` > config file, `SPINTA_BACKENDS` filter, per-request `rc.fork()`).
- **Demo manifests** — 7/7 `show`+`check` (one failure is a pre-existing manifest content issue, reproduced on base).
- **push.sh flow** — `bootstrap`, `migrate`, `push --mode external` to :8002 → exactly 3 rows incl. `_page.next`; re-push → no duplicates (push state + keymap).
- **Automated tests** — 97 passed sanity run (`test_config.py`, `test_config_overrides.py`, `types/test_model.py`).

Every failure encountered during verification reproduces identically on the pre-rewrite base (`dfd956e9`) and is unrelated to this change:

1. `SPINTA_BACKENDS=` subtree reset is resurrected by `rc.fork()` in `configure_rc` (`spinta/core/config.py:919`).


spinta/core/config.py
```diff
@@ -77,11 +108,12 @@ def read(self, schema: Schema):
...
         for k, v in self.config.items():
             v = dict(_traverse(v, k))
-            v.update(_get_inner_keys(v, depth=len(k)))
+            inner = _get_inner_keys(v, depth=len(k))
+            v.update(inner)
             config.update(v)
         self.config = config
 
-    def keys(self, env: str = None):
+    def keys(self, env: Optional[str] = None):
```
JustinasKen, line 116: In most recent updates, majority of new code started using built in python type checks, like `env: str | None = None`, it would probably be good idea to apply this here aswell, to keep consistency.

This comment applies to all parts in this PR (noticed in in multiple places, common types are used from typing library, instead of builtin).

spinta/core/config.py
```diff
@@ -33,22 +49,37 @@
...
+            *[Path(c, c) for c in configs],
+            EnvFile("envfile", envfile or ".env"),
+            EnvVars("envvars", os.environ),
+            CliArgs("cliargs", args or []),
+        ]
+        rc = RawConfig()
+        rc.read(final_sources)
+        return rc
+    else:
+        return tmp_rc
```
JustinasKen, line 82: Not the biggest fan how config reads happen (they are tied to external function, like `read_config`).

In https://github.com/atviriduomenys/spinta/issues/1990 I did mention `_read_chained_configs` as example, but later on I decided to rework it and move that logic to `rc.read()`, since then the support for `config` parameter would not be tied to external function, but be built in as part of `RawConfig` feature.

https://github.com/atviriduomenys/spinta/pull/1991/changes#diff-e14ec6276b5a6c3a3d42144ad3e984a7a628f160755d4ef4123f39a1c58ee65bR212-R265

While `read_config` only stored preset order.

But that featured heavily relied on `before/after` and `anchor` parameters.

Otherwise it's fair hacky to utilize this featuer when not using `read_config` function, as you can see inside:
https://github.com/atviriduomenys/spinta/pull/2001/changes#diff-da24b355349b53e4c8047034767d008e4c2ce9ea6c8933a3acd50d0828f00528R360-R435

Here you have to recreate manually external config read, when nested ones will be introduced, it will then require separate loops to keep track. In my opinion nested config read should be part of `rc.read()` method (then runtime you could do rc.read(...) and load new config source on top, otherwise it would not work and you would need separate functions to check if new source had `config` parameter), but thats only my opinion.

PR #2005, opened 2026-08-12
health check endpoint

### LINKED ISSUE #1873: `/health` probe
Įgyvendinti `/health` endpoint, kuris patikrintu, ar spinta ir prijungtas duomenų šaltinis yra gyvi.

`/health` turi patikrinti:

- Ar pasiekiami visi backendai (duomenų šaltiniai).
- Ar yra pakankamai laisvos disko vietos.
- Ar yra pakankamai RAM atminties.

karina-klinkeviciute:
Šiame PR įgyvendinti šie health checkpointo dalykai:

    pačios spintos reportinimas-
    disko vietos reportinimas
    atminties reportinimas

Nėra pranešama, kiek disko vietos ar atminties likę, kad nebūtų atskleista per daug informacijos, tik ar jos užtenka, ar ne.

Disko vietos ir atmintie ribos, ties kuriomis pranešama, yra konfigūruojamos, bet turi savo defaultus.

### DIFF
--- a/spinta/api/health.py
+++ b/spinta/api/health.py
@@ -1,0 +1,116 @@
+"""`/health` probe.
+
+Reports whether Spinta is operational: whether there is enough free disk space
+and enough available memory. The response format, the configuration options and
+how the thresholds are chosen are documented in
+`docs/en/manual/configuration/health.md`.
+
+The checks are a handful of file system reads, so this is a plain synchronous
+handler, which Starlette runs in a worker thread. That keeps a check that does
+block, `statvfs` on an unresponsive network mount for one, off the event loop.
+
+Details of a failed check go to the log and never to the response: the probe is
+not authenticated, so it must not disclose how the service is deployed.
+"""
+
+import logging
+import pathlib
+import threading
+
+import psutil
+from starlette.requests import Request
+from starlette.responses import JSONResponse
+
+from spinta.components import Config, Context
+from spinta.utils.memory import available_memory
+
+log = logging.getLogger(__name__)
+
+MB = 1024 * 1024
+
+# Only one disk check runs at a time, because `statvfs` blocks on an
+# unresponsive mount and every probe would otherwise start another worker thread
+# that never returns. Waiting a moment for a check that is merely in progress
+# keeps two overlapping probes from reporting a healthy disk as unhealthy.
+DISK_CHECK_WAIT = 1
+_disk_check = threading.Lock()
+
+
+def _existing_path(path: pathlib.Path) -> pathlib.Path:
+    """Return the closest existing directory of a possibly not yet created path."""
+    for candidate in (path, *path.parents):
+        if candidate.exists():
+            return candidate
+    return pathlib.Path(path.anchor or ".")
+
+
+def check_disk(config: Config) -> bool:
+    """Check that enough disk space is left where Spinta keeps its data."""
+    if not _disk_check.acquire(timeout=DISK_CHECK_WAIT):
+        log.error("Skipping the disk check, the previous one has not finished yet.")
+        return False
+    try:
+        return _free_disk_space_is_enough(config)
+    finally:
+        _disk_check.release()
+
+
+def _free_disk_space_is_enough(config: Config) -> bool:
+    path = config.data_path
+    try:
+        # `Path.exists()` raises, for example when a parent directory may not be
+        # searched, so resolving the path has to be guarded as well.
+        path = _existing_path(path)
+        free = psutil.disk_usage(str(path)).free // MB
+    except Exception:
+        log.exception("Error while checking free disk space of %s.", path)
+        return False
+
+    if free < config.health_min_free_disk_space:
+        log.error(
+            "Not enough free disk space on %s: %s MB free, %s MB required.",
+            path,
+            free,
+            config.health_min_free_disk_space,
+        )
+        return False
+    return True
+
+
+def check_memory(config: Config) -> bool:
+    """Check that enough memory is left for Spinta to keep working."""
+    try:
+        available = available_memory() // MB
+    except Exception:
+        log.exception("Error while checking available memory.")
+        return False
+
+    if available < config.health_min_free_memory:
+        log.error(
+            "Not enough available memory: %s MB available, %s MB required.",
+            available,
+            config.health_min_free_memory,
+        )
+        return False
+    return True
+
+
+def health(request: Request) -> JSONResponse:
+    context: Context = request.state.context
+    config: Config = context.get("config")
+
+    dependencies = [
+        # Spinta itself answered, everything else it needs is checked separately.
+        {"name": "spinta", "healthy": True},
+        {"name": "disk", "healthy": check_disk(config)},
+        {"name": "memory", "healthy": check_memory(config)},
+    ]
+
+    return JSONResponse(
+        {
+            "healthy": all(dependency["healthy"] for dependency in dependencies),
+            "dependencies": dependencies,
+        },
+        # Probes must always get the current state, never a cached one.
+        headers={"Cache-Control": "no-store"},
+    )
--- a/spinta/config.py
+++ b/spinta/config.py
@@ -267,6 +267,13 @@ CONFIG = {
     # Response HTTP Strict Transport Security (HSTS) header. `max-age` must be at
     # least 31536000 seconds (1 year) and `includeSubDomains` must be specified.
     "http_strict_transport_security": "max-age=31536000; includeSubDomains",
+    # `/health` probe thresholds.
+    "health": {
+        # Minimum amount of free disk space (MB) on `data_path`.
+        "min_free_disk_space": 2048,
+        # Minimum amount of available RAM (MB).
+        "min_free_memory": 256,
+    },
     # Default postgresql backend sharding distribution strategy (set it to `undistributed` to disable sharding)
     "default_distribution_strategy": "schema",
     "default_distribution_property": "_id",
--- a/spinta/utils/memory.py
+++ b/spinta/utils/memory.py
@@ -1,0 +1,153 @@
+"""How much memory this process is still allowed to use.
+
+`psutil` reads `/proc/meminfo`, which is not namespaced: inside a container it
+reports the memory of the host, not the limit the container was given. A
+container about to be killed for using up its memory would therefore look
+perfectly healthy. So the limit of the control group this process belongs to is
+taken into account when it has one.
+
+Neither number replaces the other. A cgroup limit is a ceiling, not a
+reservation, so a container with room to spare is still killed when the machine
+itself runs out of memory. What this process can really use is whichever of the
+two is tighter.
+"""
+
+import pathlib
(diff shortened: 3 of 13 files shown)


spinta/api/health.py
```diff
@@ -0,0 +1,201 @@
...
+    # which must answer in bounded time, but it does mean each backend check has
+    # to be bounded at the driver level as well, see for example
+    # `spinta.backends.postgresql.commands.wait.CONNECT_TIMEOUT`.
+    #
+    # Checks run concurrently, in threads, so each one gets its own context
+    # state instead of mutating the shared request context.
+    with context.fork("health") as fork:
+        check = asyncio.ensure_future(run_in_threadpool(commands.wait, fork, backend, fail=False))
+        try:
+            return await asyncio.wait_for(check, timeout=timeout)
```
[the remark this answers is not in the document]
karina-klinkeviciute: Fixed by keeping one in-flight check per backend: BackendChecks holds a per-backend lock acquired inside the worker thread itself, so a backend whose previous check is still running is reported unhealthy without opening     
  ▎ another connection. Thread count is now bounded by the number of backends, not by probe rate. Note that sas+jdbc and cx_oracle still have no driver-level timeout — their single thread unwinds when the OS TCP connect times   
  ▎ out.

spinta/api/health.py
```diff
@@ -0,0 +1,239 @@
...
+        lock.release()
+
+
+async def _check_backend(context: Context, backend: Backend, timeout: float, checks: BackendChecks) -> bool:
+    # Checks run concurrently, in threads, so each one gets its own context
+    # state instead of mutating the shared request context.
+    with context.fork("health") as fork:
+        check = asyncio.ensure_future(run_in_threadpool(_wait, fork, backend, checks.lock(backend.name)))
+        try:
+            return await asyncio.wait_for(check, timeout=timeout)
```
kastger: Does `health_backend_timeout` guarantee that `_check_backend()` returns within that amount of time What happens to the worker thread when the timeout fires?

kastger:
I think this needs a higher-level rethink before a detailed review. The code mixes `asyncio`, `AnyIO`/`Starlette` `threadpool`, and threading, which makes the concurrency model unnecessarily difficult to reason about.

In particular, the backend timeout cancels the `asyncio` task while the underlying thread keeps running, which then requires additional locking to prevent checks from piling up. Is there a reason we need this level of complexity? Personally, I wouldn't feel confident maintaining this code without spending significant time understanding all the concurrency interactions.

spinta/config.py
```diff
@@ -267,6 +267,18 @@
...
+        # Minimum amount of free disk space (MB) on `data_path`.
+        "min_free_disk_space": 1024,
+        # Minimum amount of available RAM (MB).
+        "min_free_memory": 256,
+        # How long (seconds) to wait for a single backend to respond.
+        "backend_timeout": 5,
+        # How long (seconds) to reuse a health check result. Set to 0 to check
+        # on every request.
+        "cache_time": 5,
+    },
```
JPriev, line 276: Any particular reason why you chose these thresholds? I think Disk space should be higher, because of versioning and logs. 

Also what about RAM in docker. According to this old SO comment (https://stackoverflow.com/a/46213331), if spinta would be in docker it would check hosts RAM available instead of containers.

spinta/api/health.py
```diff
@@ -0,0 +1,239 @@
...
+    return pathlib.Path(path.anchor or ".")
+
+
+def _check_disk(config: Config) -> bool:
+    path = config.data_path
+    try:
+        # `Path.exists()` raises, for example when a parent directory may not be
+        # searched, so resolving the path has to be guarded as well.
+        path = _existing_path(path)
+        free = psutil.disk_usage(str(path)).free // MB
```
JPriev, line 64: According psutil docs it seems that this command will return `Bytes` instead of `MB`
https://psutil.readthedocs.io/stable/index.html#psutil.disk_usage

spinta/config.py
karina-klinkeviciute: all these are configurable. But it's a good point, even the default is probably better when higher. 

As for the RAM, well noted. I will see what could be solution here

karina-klinkeviciute:
> I think this needs a higher-level rethink before a detailed review. The code mixes `asyncio`, `AnyIO`/`Starlette` `threadpool`, and threading, which makes the concurrency model unnecessarily difficult to reason about.
> 
> In particular, the backend timeout cancels the `asyncio` task while the underlying thread keeps running, which then requires additional locking to prevent checks from piling up. Is there a reason we need this level of complexity? Personally, I wouldn't feel confident maintaining this code without spending significant time understanding all the concurrency interactions.

@kastger thanks, I will rethink this. Probably you are right.

spinta/api/health.py
karina-klinkeviciute: there is a division by MB in the end - `// MB` which turns bytes into MBs

JPriev: My bad, looked like comment because of the `//`

spinta/config.py
karina-klinkeviciute: container RAM taken into account, disk limit increased to 2 GB, which correspoimds to 10 GB recommended in the docs quite well

spinta/api/health.py
```diff
@@ -0,0 +1,169 @@
...
```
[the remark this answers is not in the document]
karina-klinkeviciute: fixed

spinta/api/health.py
```diff
@@ -0,0 +1,244 @@
...
+    parts = [part for part in _cgroup_of_this_process(layout.controller).split("/") if part]
+    return [layout.root.joinpath(*parts[:depth]) for depth in range(len(parts), -1, -1)]
+
+
+def _available_within(layout: CgroupLayout, directory: pathlib.Path) -> int | None:
+    """Memory left within one cgroup's limit, or None if it has no limit."""
+    limit = _read_int(directory / layout.limit)
+    usage = _read_int(directory / layout.usage)
+    if limit is None or usage is None or limit >= NO_CGROUP_LIMIT:
+        return None
```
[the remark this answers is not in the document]
karina-klinkeviciute: pataisyta

spinta/api/health.py
```diff
@@ -0,0 +1,232 @@
...
+    parts = [part for part in _cgroup_of_this_process(layout.controller).split("/") if part]
+    return [layout.root.joinpath(*parts[:depth]) for depth in range(len(parts), -1, -1)]
+
+
+def _cgroup_available_memory() -> int | None:
+    """Memory left within this process' cgroup limit, or None if not limited."""
+    for layout in CGROUP_LAYOUTS:
+        # A limit can be set on this process' own cgroup or on any of its
+        # ancestors, and the closest one is the one that applies.
+        for directory in _cgroup_dirs(layout):
```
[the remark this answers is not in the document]
karina-klinkeviciute: fixed

spinta/api/health.py
```diff
@@ -0,0 +1,257 @@
...
+    for candidate in (path, *path.parents):
+        if candidate.exists():
+            return candidate
+    return pathlib.Path(path.anchor or ".")
+
+
+def _check_disk(config: Config) -> bool:
+    if not _disk_check.acquire(blocking=False):
+        log.error("Skipping the disk check, the previous one has not finished yet.")
+        return False
```
kastger: `.acquire(blocking=False)` cannot tell "the previous check is hung on a dead NFS mount" from "another probe arrived while this one was in progress". Starlette runs sync endpoints concurrently in its threadpool and health endpoints are exactly the thing that gets hit constantly. Two overlapping probes make the second one report `disk: healthy=false` on a machine where nothing is wrong. The window is small but the failure is silent and wrong. Add a small timeout:
```
if not _disk_check.acquire(timeout=1):
```

spinta/api/health.py
```diff
@@ -0,0 +1,257 @@
...
+
+Only these flags are reported. Since the probe is not authenticated, it must not
+disclose how the service is deployed, so paths, free space and errors are
+written to the log instead of to the response.
+
+The checks are a handful of file system reads, so this is a plain synchronous
+handler, which Starlette runs in a worker thread. That keeps a check that does
+block, `statvfs` on an unresponsive network mount for one, off the event loop.
+"""
+
```
kastger, line 15: We have `health.md` now and this docstring mostly repeats it. Each appear - UAPI schema shape, "only flags reported because the probe is unauthenticated", the `/proc/meminfo` note - each appear two or three times across this docstring, comments, `health.md` and `CHANGES.rst`.

This PR adds 294 lines of code and 114 comments and docstrings. Code tells us how, comments should tell us why and say it once. Here the same why is written in three or four places, so when it changes we have to remember to update all of them.

tests/test_health.py
```diff
@@ -0,0 +1,254 @@
...
+import pytest
+
+from spinta.api.health import (
+    CgroupLayout,
+    _available_memory,
+    _cgroup_available_memory,
+    _check_disk,
+    _check_memory,
+    health,
+)
```
kastger: 4 of these names are private.

In the book of Pragmatic Unit Testing:
```
In general, you don't want to break any encapsulation for the sake of testing (or as Mom used to say, "don't expose your privates!"). Most of the time, you should be able to test a class by exercising its public methods. If there is significant functionality that is hidden behind private or protected access, that might be a warning sign that there's another class in there struggling to get out.
```

About 150 lines of the 257 in `health.py` are cgroup accounting - `CgroupLayout`, the layouts, `PROC_CGROUP`, `NO_CGROUP_LIMIT`, and everything from `_read_int` through `_check_memory`. None of it has anything to do with HTTP. Moving it to something like `spinta/utils/memory.py` with one public `available_memory()` would leave `health.py` with a single concern and the tests could import a public function instead of reaching inside and breaking encapsulation.

kastger (approved):

spinta/api/health.py
karina-klinkeviciute: Fixed as you suggested — now acquire(timeout=1), so overlapping probes wait instead of reporting a healthy disk as unhealthy, and only a check genuinely stuck on the mount returns false.

spinta/api/health.py
karina-klinkeviciute: docstringas sutrumpintas, palikta tik tai, ką patogu turėti prie kodo, ir įdėta nuoroda į dokumentacijos failą.

tests/test_health.py
karina-klinkeviciute: testai pataisyti ir private funkcijų nebekviečia

spinta/utils/memory.py
```diff
@@ -0,0 +1,149 @@
...
+            return min(headroom)
+    return None
+
+
+def available_memory() -> int:
+    """Memory, in bytes, this process can still use before it is killed."""
+    available = _cgroup_available_memory()
+    if available is None:
+        available = psutil.virtual_memory().available
+    return available
```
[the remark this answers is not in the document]
karina-klinkeviciute: pataisyta, taip pat pataisyti testai

PR #2007, opened 2026-08-14
1873 health backends

### LINKED ISSUE #2011: add backends info to the health probe
Health probe was added, but for speed and simplicity it was done without the backends info

Backends info should be added to the health probe too. 

Ar pasiekiami visi backendai (duomenų šaltiniai).

karina-klinkeviciute:
Šis PR atidedamas vėlesniam laikui,

### DIFF
--- a/spinta/api/health.py
+++ b/spinta/api/health.py
@@ -1,7 +1,8 @@
 """`/health` probe.
 
-Checks if Spinta is operational:
+Checks if Spinta and everything it depends on is operational:
 
+- if all backends (data sources) are reachable,
 - if there is enough free disk space,
 - if there is enough free RAM.
 
@@ -11,25 +12,36 @@ for the whole service and a `dependencies` list, where each item is a named
 dependency with its own `healthy` flag.
 
 Only these flags are reported. Since the probe is not authenticated, it must not
-disclose how the service is deployed, so paths, free space and errors are
-written to the log instead of to the response.
+disclose how the service is deployed, so individual backend names, paths, free
+space and driver errors are written to the log instead of to the response.
 
-Checks read from the file system, which can block, so they run in worker threads
-rather than on the event loop.
+Checks block, so they run in worker threads, and they run at most one probe at a
+time. Nothing here cancels a check: how long a backend check may take is the
+driver's own connect timeout, see `WAIT_CONNECT_TIMEOUT`, and how long the probe
+as a whole may take is up to whoever probes, which for a container or load
+balancer probe is its own timeout.
 """
 
 from __future__ import annotations
 
 import asyncio
+import functools
 import logging
 import pathlib
+import time
+import weakref
+from typing import Awaitable, Callable
 
 import psutil
 from starlette.concurrency import run_in_threadpool
 from starlette.requests import Request
 from starlette.responses import JSONResponse
 
-from spinta.components import Config, Context
+from spinta import commands
+from spinta.backends import Backend
+from spinta.backends.constants import WAIT_CONNECT_TIMEOUT
+from spinta.backends.helpers import get_all_backends
+from spinta.components import Config, Context, Store
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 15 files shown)


spinta/api/health.py
```diff
@@ -0,0 +1,269 @@
...
+
+# Memory accounting of the control group this process belongs to. `psutil` reads
+# `/proc/meminfo`, which inside a container reports the memory of the host, not
+# the limit the container was given, so a container about to be killed for using
+# up its memory would still look perfectly healthy.
+CGROUP_MEMORY = (
+    # cgroup v2
+    pathlib.Path("/sys/fs/cgroup/memory.max"),
+    pathlib.Path("/sys/fs/cgroup/memory.current"),
+    pathlib.Path("/sys/fs/cgroup/memory.stat"),
```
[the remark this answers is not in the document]
karina-klinkeviciute: done

spinta/api/health.py
```diff
@@ -0,0 +1,269 @@
...
+
+    for line in stat.splitlines():
+        # `inactive_file` in cgroup v2, `total_inactive_file` in cgroup v1.
+        name, _, value = line.partition(" ")
+        if name in ("inactive_file", "total_inactive_file"):
+            try:
+                return int(value)
+            except ValueError:
+                return 0
+    return 0
```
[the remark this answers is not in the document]
karina-klinkeviciute: done

tests/test_health.py
```diff
@@ -0,0 +1,295 @@
...
+        "healthy": True,
+        "dependencies": [{"name": name, "healthy": True} for name in DEPENDENCIES],
+    }
+
+
+def test_health_backend_unavailable(app: TestClient, mocker):
+    mocker.patch("spinta.commands.wait", return_value=False)
+
+    resp = app.get("/health")
+    assert resp.status_code == 200
```
kastger, line 38: Why for unavailable health check we are returning status code 200?

tests/test_health.py
```diff
@@ -0,0 +1,295 @@
...
+    assert data["healthy"] is False
+    assert _healthy(data, "backends") is False
+    assert _healthy(data, "spinta") is True
+
+
+def test_health_backend_error(app: TestClient, mocker):
+    mocker.patch("spinta.commands.wait", side_effect=RuntimeError("connection refused"))
+
+    resp = app.get("/health")
+    assert resp.status_code == 200
```
kastger, line 50: Same here, why 200 status code if there are errors? Container probes only look at the status code, so with 200 and `healthy: false` it would report a broken service as healthy.

spinta/api/health.py
```diff
@@ -0,0 +1,269 @@
...
+from spinta.components import Config, Context, Store
+
+log = logging.getLogger(__name__)
+
+MB = 1024 * 1024
+
+# Memory accounting of the control group this process belongs to. `psutil` reads
+# `/proc/meminfo`, which inside a container reports the memory of the host, not
+# the limit the container was given, so a container about to be killed for using
+# up its memory would still look perfectly healthy.
```
kastger, line 53: General note on the PR: the non-test files add 213 of code and 111 lines of comments and docstrings. The `connect_timeout=0` explanation alone appears seven times. If something related to that changes, it would require to find and update all of the comments, otherwise they go stale and tell lies. `libpq` behavior of "wait forever" should be noted at the point of use instead.

spinta/api/health.py
```diff
@@ -0,0 +1,269 @@
...
+CGROUP_MEMORY = (
+    # cgroup v2
```
kastger, line 55: nit: Why does this have a comment about the version, but for the v1 is inside the variable name?

spinta/api/health.py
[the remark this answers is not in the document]
kastger: Copilot has a great point, tho this was resolved without any commits on top. Was the fix left locally or did "done" mean something else here?

spinta/api/health.py
```diff
@@ -0,0 +1,269 @@
...
+    result is served again; with `0` every probe gets a check of its own, still
+    one at a time.
+    """
+
+    def __init__(self):
+        self._result: dict | None = None
+        self._expires: float = 0.0
+        # An `asyncio.Lock` belongs to the event loop it was created in. A served
+        # app has a single loop; more than one appears in tests, where each probe
+        # may get its own, and those then share the result but not the lock.
```
kastger, line 86: It does not - a lock binds the first time it actually has to wait, not at creation. `acquire()` returns early without calling `_get_loop()` when the lock is free.

spinta/api/health.py
```diff
@@ -0,0 +1,269 @@
...
+        self._locks: weakref.WeakKeyDictionary = weakref.WeakKeyDictionary()
+
+    def _lock(self) -> asyncio.Lock:
+        loop = asyncio.get_running_loop()
+        if (lock := self._locks.get(loop)) is None:
+            lock = self._locks[loop] = asyncio.Lock()
+        return lock
```
kastger, line 93: Why is this needed? A single `self._lock = asyncio.Lock()` in `__init__` would be enough for a served app, which has one event loop per worker. Two loops get two different locks while sharing `_result` and `_expires`, so their checks can run concurrently - which is exactly what the lock was added to prevent.

PR #2013, opened 2026-08-21
631 dvms o auth2 endpoints

### LINKED ISSUE #1938: Missing `iss` and `exp` claim validations on decode
`iss` and `exp` claims of the JWT token should be validated during the decoding of the JWT.

JPriev:


spinta/auth.py
```diff
@@ -730,6 +778,31 @@ def create_client_access_token(context: Context, client: Union[str, Client]):
...
 
 
+def get_issuer(config: Config) -> str:
+    """Issuer identifier of this authorization server.
+
+    RFC 8414 section 2 requires `issuer` to be identical to the URL the metadata
+    document was retrieved from with the well-known suffix removed, so it must
+    not carry a trailing slash.
+    """
+    return config.server_url.rstrip("/")
```
sirex, line 788: I would move this normalization to `spinta/types/config.py:74`:

```python
config.server_url = rc.get("server_url")
```

This way, we normalize early and have single canonical value everywhere, avoiding normalization in multiple places.

spinta/auth.py
```diff
@@ -771,6 +844,15 @@ def get_client_file_path(path: pathlib.Path, client: str) -> pathlib.Path:
     return client_file
 
 
+def client_has_scope(context: Context, client: Client, scope: Union[Scopes, str]) -> bool:
+    config = context.get("config")
+
+    if isinstance(scope, Scopes):
+        scope = scope.value
```
sirex, line 851: Currently there is only one `client_has_scope` caller, which passes `scope: Scopes`, so I would reduce this to just `Scopes` and remove this if check.

spinta/auth.py
```diff
@@ -771,6 +844,15 @@ def get_client_file_path(path: pathlib.Path, client: str) -> pathlib.Path:
...
+
+    return bool({f"{config.scope_prefix}{scope}", f"{config.scope_prefix_udts}:{scope}"} & client.scopes)
```
sirex, line 853: Maybe extract scope formatting to `get_scope_name(context, prefix, None, scope)`, changing `get_scope_name` signature to `get_scope_name(context: Context, node: Union[Namespace, Model, Property, None], action: Union[Action, Scopes], is_udts: bool = False)`?

Since we have multiple transitions in scope format in progress, keeping it in one place, would be easier to manage.

And there was already discussions, on how to format system copes, when `node` is not given.

spinta/auth.py
```diff
@@ -192,6 +200,37 @@ def __init__(
...
+    CLIENT_AUTH_METHODS = ["client_secret_basic"]
+
+    def query_token(self, token_string: str, token_type_hint: str) -> Token | None:
+        if token_type_hint and token_type_hint != "access_token":
+            return None
+
+        protector: ResourceProtector = self.server.context.get("auth.resource_protector")
+        try:
+            return authenticate_token(protector, token_string, "bearer")
+        except (InvalidToken, JoseError, KeyError):
```
sirex, line 213: Is there a reason to add generic `KeyError` here? If this is just for:

```python
self.expires_in = self._token["exp"] - self._token["iat"]
token_kty = decode_kty_from_alg(token_header["alg"])
```

I would add `.get(...)` and avoid using `KeyError`, and silencing pottential bugs.

spinta/auth.py
```diff
@@ -192,6 +200,37 @@ def __init__(
         self.register_token_validator(Validator(context))
 
 
+class IntrospectionEndpoint(rfc7662.IntrospectionEndpoint):
+    CLIENT_AUTH_METHODS = ["client_secret_basic"]
+
+    def query_token(self, token_string: str, token_type_hint: str) -> Token | None:
```
sirex, line 206:
```suggestion
    def query_token(self, token_string: str, token_type_hint: str | None) -> Token | None:
```

`token_type_hint` might be `None`, see:

```python
        token = self.query_token(
            request.form["token"], request.form.get("token_type_hint")
        )
```

From `rfc7662/introspection.py:39`.

spinta/auth.py
```diff
@@ -730,6 +778,31 @@ def create_client_access_token(context: Context, client: Union[str, Client]):
...
+    document was retrieved from with the well-known suffix removed, so it must
+    not carry a trailing slash.
+    """
+    return config.server_url.rstrip("/")
+
+
+def get_authorization_server_metadata(context: Context) -> dict:
+    config = context.get("config")
+    issuer = get_issuer(config)
+    return {
```
sirex, line 794: I would return this as `AuthorizationServerMetadata` from `authlib.oauth2.rfc8414` and would change function type annotation to `-> AuthorizationServerMetadata`, just make others aware, that a specific domain model is used, not a free form dict.

Also would be nice to run `AuthorizationServerMetadata.validate()` at least in tests, to check if returned metadata is correct.

tests/test_auth.py
```diff
@@ -1367,3 +1368,122 @@ def test_success_if_token_has_less_scopes_than_contract(self, rc: RawConfig):
...
+    resp = _introspect(introspect_app, token, auth=("reader", "reader-secret"))
+    assert resp.status_code == 200, resp.text
+    assert resp.json() == {"active": False}
+
+
+def test_introspect_own_token_requires_scope(introspect_app):
+    token = _get_access_token(introspect_app, "reader", "reader-secret")
+
+    resp = _introspect(introspect_app, token, auth=("reader", "reader-secret"))
+    assert resp.status_code == 200, resp.text
```
sirex, line 1458: `reader` client does not have `auth_introspect` scope, so it does not even have access to the protected resource and should get 403 HTTP error with `InsufficientPermission` error code.

See specification:

> `insufficient_scope`
>
> The request requires higher privileges than provided by the
> access token.  The resource server SHOULD respond with the HTTP
> 403 (Forbidden) status code and MAY include the "scope"
> attribute with the scope necessary to access the protected
> resource.

-- <https://datatracker.ietf.org/doc/html/rfc6750#section-3.1>

200 with `active=false` should be returned only if the instrospected token is not  properly authorized, not the client:

> If the introspection call is properly authorized but the token is not
> active, does not exist on this server, or the protected resource is
> not allowed to introspect this particular token, then the
> authorization server MUST return an introspection response with the
> "active" field set to "false".  Note that to avoid disclosing too
> much of the authorization server's state to a third party, the
> authorization server SHOULD NOT include any additional information
> about an inactive token, including why the token is inactive.

spinta/auth.py
```diff
@@ -730,6 +778,31 @@ def create_client_access_token(context: Context, client: Union[str, Client]):
...
+        "issuer": issuer,
+        "token_endpoint": f"{issuer}/auth/token",
+        "introspection_endpoint": f"{issuer}/auth/introspect",
+        "jwks_uri": f"{issuer}/.well-known/jwks.json",
+        "grant_types_supported": ["client_credentials"],
+        "response_types_supported": [],
```
sirex, line 800: This is a required field, a value must be provided, see `authlib`'s own validation logic:

```python
    def validate_response_types_supported(self):
        """REQUIRED.  JSON array containing a list of the OAuth 2.0
        "response_type" values that this authorization server supports.
        The array values used are the same as those used with the
        "response_types" parameter defined by "OAuth 2.0 Dynamic Client
        Registration Protocol" [RFC7591].
        """
        response_types_supported = self.get("response_types_supported")
        if not response_types_supported:
            raise ValueError('"response_types_supported" is required')
        if not isinstance(response_types_supported, list):
            raise ValueError('"response_types_supported" MUST be JSON array')
```

tests/test_auth.py
```diff
@@ -1367,3 +1368,122 @@ def test_success_if_token_has_less_scopes_than_contract(self, rc: RawConfig):
...
+    assert metadata["token_endpoint_auth_methods_supported"] == ["client_secret_basic"]
+
+
+def test_metadata_issuer_matches_token_issuer(introspect_app):
+    """RFC 8414 requires the advertised issuer to match the `iss` claim."""
+    metadata = introspect_app.get("/.well-known/oauth-authorization-server").json()
+    token = _get_access_token(introspect_app, "reader", "reader-secret")
+
+    payload = _introspect(introspect_app, token).json()
+    assert payload["iss"] == metadata["issuer"]
```
sirex, line 1489: Could you also add a test, where token has different issuer, this should validate the RFC requirement property.

Also, I guess, this will trigger this issue:

- https://github.com/atviriduomenys/spinta/issues/1938

sirex (commented):
Part B from issue description is completely missing. I would suggest to increase `authlib` version requirement:

```toml
authlib = ">=1.7.2"
```

To make sure, that listed CVEs will not be installed.

Also, I could not find anything regarding "Patikrinta `alg:none` hipotezė realiu testu → jei pasitvirtina, atskira skubi užduotis".

In the test report, the "NON AUTH INTROSPECTION USER doing it's own Introspection" test should not work, because `DR_TOKEN` does not have `auth_introspect` scope, but you demonstrate, that request is successful, even if client does not access to call `/auth/introspect`.

I see, that pytest tests show different behaviour:

```python
def test_introspect_own_token_requires_scope(introspect_app):
    token = _get_access_token(introspect_app, "reader", "reader-secret")

    resp = _introspect(introspect_app, token, auth=("reader", "reader-secret"))
    assert resp.status_code == 200, resp.text
    assert resp.json()["active"] is False
```

It would be nice, if test reports would be provided as a `.md` or `.sh` files in the repository, inside `notes/auth/introspect.md` file, so that everyone could reproduce this test.

spinta/auth.py
```diff
@@ -192,6 +200,37 @@ def __init__(
...
+        try:
+            return authenticate_token(protector, token_string, "bearer")
+        except (InvalidToken, JoseError, KeyError):
+            return None
+
+    def check_permission(self, token: Token, client: Client, request: OAuth2Request) -> bool:
+        return client_has_scope(self.server.context, client, Scopes.AUTH_INTROSPECT)
+
+    def introspect_token(self, token: Token) -> dict:
+        return {
```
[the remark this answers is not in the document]
JPriev: Decoding does validate `exp`: `_decode` runs `JWTClaimsRegistry(exp={"essential": True}).validate()` (`spinta/auth.py:322`). That raises `ExpiredTokenError`, `query_token` returns `None`, and Authlib's `create_introspection_payload` returns `{"active": False}`. Authlib also checks `token.is_expired()` or `token.is_revoked()` before calling `introspect_token`. `test_introspect_expired_token` covers this.

spinta/api/__init__.py
```diff
@@ -114,6 +115,33 @@ async def get_verification_keys(request: Request) -> JSONResponse:
...
+            query=request.query_params,
+        ),
+    )
+
+
+async def authorization_server_metadata(request: Request) -> JSONResponse:
+    context = request.state.context
+    auth_server = context.get("auth.server")
+    if not auth_server.enabled():
+        raise NoAuthServer()
```
sirex, line 140: Duplicated Code. Both new handlers repeat the same prologue already present in `auth_token` `spinta/api/__init__.py:120`:

```python
auth_server = context.get("auth.server")
if not auth_server.enabled():
    raise NoAuthServer()
```

Three call sites now share this shape → extract a small helper (e.g. `_get_auth_server(request)`).

spinta/auth.py
```diff
@@ -273,8 +337,9 @@ def decode_token(self, token_string: str) -> dict:
             if kid := (token_header.get("kid") or token_header.get("key")):
                 for key in self._all_public_keys:
                     if key.kid and str(key.kid) == str(kid):
-                        return jwt.decode(token_string, key, algorithms=ALLOWED_JWT_ALGORITHMS).claims
-
+                        return self._decode(token_string, key)
```
sirex, line 340: `BearerTokenValidator._decode` builds a fresh `JWTClaimsRegistry` (identical essential-claim specs) on every decode.

Maybe build once in `__init__`?

spinta/types/config.py
```diff
@@ -71,7 +71,7 @@ def load(context: Context, config: Config) -> Config:
         rc.get("data_path") or DEFAULT_DATA_PATH,
     )
     config.credentials_file = pathlib.Path(rc.get("credentials_file") or DEFAULT_CONFIG_PATH / "credentials.cfg")
-    config.server_url = rc.get("server_url")
+    config.server_url = rc.get("server_url").rstrip("/")
```
sirex, line 74: Raises `AttributeError` if `server_url` is set to `null`.

tests/test_auth.py
```diff
@@ -1367,3 +1536,192 @@ def test_success_if_token_has_less_scopes_than_contract(self, rc: RawConfig):
...
+            "sub": "reader",
+            "aud": "https://example.com",
+            "iat": iat,
+            "exp": iat + 600,
+            "scope": "spinta_getall",
+            "jti": str(uuid.uuid4()),
+        },
+        private_key,
+        algorithms=ALLOWED_JWT_ALGORITHMS,
+    )
```
sirex, line 1724: Currently there is a pre-existing `generate_jwt(...)`, a new `_encode_token(...)`, plus a third inline `jwt.encode` payload here — same shape (header + claims + RS512).

Maybe use one builder with claim overrides?

CHANGES.rst
```diff
@@ -1,15 +1,60 @@
...
+1.1.0 (unreleased)
 =====================
 
 Backwards incompatible:
 
+- Access tokens are now validated against their ``iss`` (issuer), ``aud``
+  (audience), ``client_id``, ``exp`` (expiration) and ``iat`` (issued-at) claims
+  at decode time: a token is rejected unless its ``iss`` equals the
+  authorization server identifier, its ``aud`` contains this resource server,
+  and it carries ``client_id``, ``exp`` (not passed) and ``iat``. A validated
```
sirex, line 13:
```suggestion
  and it carries ``client_id``, ``exp`` (not expired) and ``iat``. A validated
```

?

sirex (approved):

spinta/auth.py
JPriev: Skipped this one. Before the claims registry is built, `_require_auth_config()` has to check that `token_issuer` and `resource_server_url` are set. If I move that into `BearerTokenValidator.__init__()`, the check runs every time a validator is created, not only when a token is decoded. That would raise `RequiredConfigParam` in places that never decode anything, e.g. the CLI helper and tests that construct the validator directly.

The overhead of rebuilding the registry per `_decode` call is negligible: `JWTClaimsRegistry.__init__` just stores the kwargs, and the RSA signature verification in the same call costs far more. So I'd rather leave it as is for now.

docs/en/manual/configuration/auth.rst
```diff
@@ -21,14 +21,66 @@ Client name and secret will be used from `<config_path>/clients` directory.
...
+  that server's `iss` **exactly** — including a trailing slash if the issuer
+  uses one (e.g. some providers publish ``https://issuer.example.com/``). The
+  value is compared verbatim and is not normalised.
+- when Spinta issues its **own** tokens, set it to Spinta's own authorization
+  identity (the `iss` you want stamped on issued tokens).
+
+
+Resource server identifier (``resource_server_url``)
+----------------------------------------------------
+`resource_server_url` is the identifier of this resource server — the value
```
sirex: Maybe it's not too late to rename `resource_server_url` to `resource`?

Currently we have multiple names for the same thing:

- `instance_id` - a new `AgentEvnironment` property
- `agent_id` - a new proposed `config.yml` option for the `resource_server_url`.
- `resource_server` - from `credentials.cfg`, used by `spinta sync`
- `resource` - from RFC 8707, used in authorization grants
- `aud` - from RFC 7519, used in Access Token

Maybe, we should normalize naming, and stick to the names used in specifications?

## Proposal

- `resource` - use this as Resource Server URI, consistent with the RFC 8707.
- `resource_server` - use this as Resource Server URL, consistent with existing naming from `credentials.cfg`?

PR #2026, opened 2026-09-04
1058 allow clicking on url type files

### LINKED ISSUE #1058: Allow clicking on "URL" type files
Currently, when we create an ADSA file and specify that a certain field is of "URL" type, the result that we get in Spinta is simple plain text.
What is the purpose of this type? Is it possible to make the text become actual valid hyperlinks that take you somewher on clicking the link?

Here is an example:
In https://get.data.gov.lt/datasets/gov/llti/patarles_priezodziai/Tipas the last column, savado_url is of URL type in its ADSA. But you can't do anything with this information except select the entire thing and paste it into your browser.

Image

Can you please implement the ability to click on any valid links within the URL column?

Thank you.
Best regards,
Adrian Moroz

JustinasKen:


spinta/formats/html/commands.py
```diff
@@ -418,6 +419,26 @@ def prepare_dtype_for_response(
...
+    *,
+    data: Dict[str, Any],
+    action: Action,
+    select: dict = None,
+):
+    link = data.pop("_link", True)
+    if link:
+        return Cell(
+            value,
+            link=value,
```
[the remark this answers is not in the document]
JustinasKen: added validations

spinta/formats/html/commands.py
```diff
@@ -418,6 +421,28 @@ def prepare_dtype_for_response(
...
+    *,
+    data: Dict[str, Any],
+    action: Action,
+    select: dict = None,
+):
+    # Validate URL scheme to prevent XSS attacks
+    if _is_safe_url(value):
+        return Cell(
+            value,
+            link=value,
```
sirex, line 439: **Scheme-less URLs produce broken links.** Spec: *"actual valid hyperlinks that take you somewhere on clicking"*. `_is_safe_url()` accepts URLs with no scheme (e.g. `www.example.com`: empty scheme, empty netloc → `True`), so it renders `<a href="www.example.com">…</a>`, which the browser resolves as a **relative path inside the Spinta app** — the click does not take you anywhere useful. The tests never cover this common real-world case, despite the issue's example being exactly a data-entry URL column.

sirex, line 439: **Validation uses stripped input, output uses raw input.** `_is_safe_url` does `url = url.strip()` but the caller passes the *unstripped* `value` into `Cell(value, link=value)`. A value like `" https://example.com "` passes validation but the link carries leading/trailing whitespace (encoded `%20`) — again not "a valid hyperlink that takes you somewhere".

spinta/formats/html/commands.py
```diff
@@ -935,5 +960,50 @@ def prepare_dtype_for_response(
...
+        return False
+
+    scheme = parsed.scheme.lower()
+
+    if scheme:
+        if scheme not in SAFE_SCHEMES:
+            return False
+
+        # Network URLs should actually have a host.
+        if scheme in {"http", "https", "ftp", "ftps"}:
```
sirex, line 990: **Duplicated Code — ⚠ the `if scheme in {...}` set literal is written twice:**
```python
SAFE_SCHEMES = {"http", "https", "ftp", "ftps", "mailto"}
...
        if scheme in {"http", "https", "ftp", "ftps"}:
```
The same scheme-classification logic ("which schemes are network-hosted") is hardcoded in two places inside one function. Fix: derive a `NETWORK_SCHEMES` frozenset once next to `SAFE_SCHEMES`.

tests/formats/test_html.py
```diff
@@ -543,6 +543,30 @@ def test_limit_iter(limit, exhausted, result):
...
 
+@pytest.mark.parametrize(
+    "url, result",
+    [
+        ("https://www.example.com/path", True),
+        ("HTTP://www.example.com", True),
+        ("ftp://files.example.com/file.txt", True),
+        ("ftps://files.example.com/file.txt", True),
+        ("mailto:user@example.com", True),
+        ("/example/Country", True),
```
sirex, line 554: Internal links should not be safe. We should add links, only to external links.

spinta/formats/html/commands.py
sirex, line 439: Maybe add `target="_blank" ref="noopener noreferrer"`, so that links open external pages in a new window and without setting the `Window.opener` attribute for security reasons?

sirex (approved):

PR #2028, opened 2026-09-07
2024 refactor push_state_db migration logic

### LINKED ISSUE #2024: Refactor push_state_db migration logic
Currently `push_state_db` uses if else check in order to migrate structure. This was fine when only few changes were needed, but over time this kind of system becomes less and less maintainable. We need to refactor the logic to follow same system as `SqlAlchemyKeymap` uses, by utilizing `spinta upgrade` commands in order to apply migrations.

For that we need to replicate migration tables from the keymap.
`Keymap` changed were done with https://github.com/atviriduomenys/spinta/issues/1307 task, need to check PR, since those changed were not clearly documented inside the task description

JustinasKen:
## Summary

- Added `SqliteMigratableDb` to centralize SQLite engine, connection, metadata, table creation, and migration tracking.
- Refactored push state and SQLAlchemy keymap implementations to use the shared migratable database abstraction.
- Added push state upgrade scripts for:
  - initializing migration metadata;
  - renaming the legacy `rev` column to `checksum`.
- Added detection of outdated push state databases and the `PushStateMigrationRequired` error.
- Push state migrations are skipped unless `push_state_path` is explicitly provided.
- Added `--target` and `--tag` options to the `upgrade` command, allowing upgrade scripts to be filtered by their target or category.
- Preserved legacy push state schema migration support for outdated model and pagination tables.
- Reworked table-template handling so existing SQLAlchemy `Table` definitions are reused instead of relying on `extend_existing=True`.

Still need to add CHANGES.rst

### DIFF
--- a/spinta/cli/helpers/upgrade/scripts/keymaps/sqlalchemy/redirect_support.py
+++ b/spinta/cli/helpers/upgrade/scripts/keymaps/sqlalchemy/redirect_support.py
@@ -64,7 +64,7 @@ def migrate_table(keymap: "SqlAlchemyKeyMap", table: sa.Table):
             if temp_table in keymap.metadata.tables:
                 temp_table_exists = True
 
-            new_table = keymap._create_table(temp_table)
+            new_table = keymap.get_table(temp_table, create_missing=True)
             temp_table_exists = True
 
             count_stmt = sa.select(sa.func.count()).select_from(table)
--- a/spinta/cli/helpers/upgrade/scripts/push_state/helpers.py
+++ b/spinta/cli/helpers/upgrade/scripts/push_state/helpers.py
@@ -1,0 +1,50 @@
+from typing import Callable
+
+from spinta.cli.helpers.message import cli_message
+from spinta.cli.helpers.push.components import PUSH_STATE_PATH, PushState
+from spinta.components import Context
+from spinta.utils.sqlite import apply_migration_to_outdated_db, outdated_sqlite_db
+
+
+def requires_migration(context: Context, migration: str, additional_check: Callable | None = None, **kwargs) -> bool:
+    push_state = get_push_state(context)
+
+    if push_state is None:
+        cli_message(
+            f"Skipped {migration!r} check, please provide -o push_state_path=<path> in order to run this migration"
+        )
+        return False
+
+    return outdated_sqlite_db(context, push_state, migration, additional_check, **kwargs)
+
+
+def apply_migration_to_push_state(context: Context, migration: str, apply_migration: Callable, **kwargs):
+    push_state = get_push_state(context)
+
+    if push_state is None:
+        cli_message(
+            f"Skipped {migration!r} check, please provide -o push_state_path=<path> in order to run this migration"
+        )
+        return
+
+    with push_state:
+        apply_migration_to_outdated_db(
+            context, push_state, migration, apply_migration, push_state.dsn or "push_state_db", **kwargs
+        )
+
+
+def get_push_state(context: Context) -> PushState | None:
+    rc = context.get("rc")
+    push_state_path = rc.get(PUSH_STATE_PATH)
+
+    # RawConfig given push state path is only file path
+    if push_state_path:
+        push_state_path = f"sqlite+spinta:///{push_state_path}"
+        return PushState(push_state_path)
+
+    # Context given push state path is set when calling init_push_state, which set dsn and not direct path
+    if context.has(PUSH_STATE_PATH):
+        push_state_path = context.get(PUSH_STATE_PATH)
+        return PushState(push_state_path)
+
+    return None
--- a/spinta/cli/helpers/upgrade/scripts/push_state/rename_rev.py
+++ b/spinta/cli/helpers/upgrade/scripts/push_state/rename_rev.py
@@ -1,0 +1,40 @@
+import sqlalchemy as sa
+
+from spinta.cli.helpers.push.components import PushState
+from spinta.cli.helpers.upgrade.components import Script
+from spinta.cli.helpers.upgrade.scripts.push_state.helpers import apply_migration_to_push_state, requires_migration
+from spinta.components import Context
+
+
+def requires_push_state_rename_rev_migration(context: Context, **kwargs) -> bool:
+    return requires_migration(context, Script.PUSH_REV_RENAME.value, None, **kwargs)
+
+
+def push_state_rename_rev_migration(context: Context, **kwargs):
+    apply_migration_to_push_state(context, Script.PUSH_REV_RENAME.value, apply_migration, **kwargs)
+
+
+def apply_migration(context: Context, push_state: PushState, migration: str):
+    def migrate(push_state_table: sa.Table):
+        from alembic.migration import MigrationContext
+        from alembic.operations import Operations
+
+        connection = push_state.conn
+        ctx = MigrationContext.configure(
+            connection, opts={"target_metadata": push_state.metadata, "transactional_ddl:": True}
+        )
+        op = Operations(ctx)
+        table_name = push_state_table.name
+        op.alter_column(table_name=table_name, column_name="rev", new_column_name="checksum")
+
+    insp = sa.inspect(push_state.engine)
+    table_names = insp.get_table_names()
+    push_state.metadata.reflect()
+    for table in table_names:
+        if table.startswith("_"):
+            continue
+
+        columns = insp.get_columns(table)
+        column_names = [col["name"] for col in columns]
+        if "rev" in column_names and "checksum" not in column_names:
+            migrate(push_state_table=push_state.get_table(table))
--- a/spinta/cli/helpers/data.py
+++ b/spinta/cli/helpers/data.py
@@ -36,11 +36,11 @@ def _get_row_count(context: components.Context, model: components.Model, page_da
 def count_rows(
     context: Context,
     models: List[Model],
-    limit: int = None,
+    limit: int | None = None,
     *,
-    initial_page_data: dict = None,
+    initial_page_data: dict | None = None,
     stop_on_error: bool = False,
-    error_counter: ErrorCounter = None,
+    error_counter: ErrorCounter | None = None,
     no_progress_bar: bool = False,
 ) -> Dict[str, int]:
     counts = {}
@@ -70,11 +70,11 @@ def count_rows(
 def read_model_data(
     context: components.Context,
     model: components.Model,
-    limit: int = None,
+    limit: int | None = None,
     stop_on_error: bool = False,
-    params: QueryParams = None,
-    page: Page = None,
-    query: Expr = None,
+    params: QueryParams | None = None,
+    page: Page | None = None,
+    query: Expr | None = None,
 ) -> Iterable[Dict[str, Any]]:
     if limit is not None:
         if query is None:
@@ -103,7 +103,7 @@ def read_model_data(
         yield item
 
 
-def filter_allowed_props_for_model(model: Model) -> (list, bool):
+def filter_allowed_props_for_model(model: Model) -> tuple[list, bool]:
     allowed_props = list(model.properties.keys())
     if model.base:
         for name, prop in model.base.parent.properties.items():
@@ -126,7 +126,7 @@ def iter_model_rows(
     context: Context,
     models: List[Model],
     counts: Dict[str, int],
-    limit: int = None,
+    limit: int = None | None,
     *,
     stop_on_error: bool = False,
     no_progress_bar: bool = False,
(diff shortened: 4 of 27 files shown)


spinta/cli/helpers/upgrade/scripts/push_state/rename_rev.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def apply_migration(context: Context, push_state: PushState, migration: str):
+    def migrate(push_state_table: sa.Table):
+        from alembic.migration import MigrationContext
+        from alembic.operations import Operations
+
+        connection = push_state.conn
+        ctx = MigrationContext.configure(
+            connection, opts={"target_metadata": push_state.metadata, "transactional_ddl:": True}
```
sirex, line 24:
```suggestion
            connection, opts={"target_metadata": push_state.metadata, "transactional_ddl": True}
```

spinta/cli/helpers/upgrade/scripts/keymaps/sqlalchemy/redirect_support.py
sirex, line 50:
```suggestion
    ctx = MigrationContext.configure(connection, opts={"target_metadata": keymap.metadata, "transactional_ddl": True})
```

spinta/cli/helpers/upgrade/scripts/push_state/helpers.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+        return
+
+    with push_state:
+        apply_migration_to_outdated_db(
+            context, push_state, migration, apply_migration, push_state.dsn or "push_state_db", **kwargs
+        )
+
+
+def get_push_state(context: Context) -> PushState | None:
```
sirex, line 36: I don't like ad-hoc, undocumented rc key `push_state_path`, and the use of `context` key set by `init_push_state` as a way to pass an argument.

Currently `push_state_path` is known from two places:

- `{data_path}/push/{remote}.db` - computed by `spinta push`
- `spinta push --state=...` - explicit state db location

The `{data_path}/push/{remote}.db` could be auto-discovered and returned by `get_push_state` as a list of all states db files for all remotes.

For the custom state db, I would suggest to attach it to `spinta upgrade --target=push_state_db=/path/to/state.db`.

This would require to extend the `--target` option, but in a backwards compatible way.

**Suggestion** - extends existing `spinta upgrade -t/--target`, in a backwards-compatible way:
- `--target {target}[={artifact}]` - add an optional target artifact.
- `--target push_state` — as today, *but* now means: auto-discover and migrate **all** `{data_path}/push/*.db` (per-remote push-state DBs).
- `--target push_state=/path/to/state.db` — migrate that explicit artifact only.
- No `--target` at all (full run) — `push_state` scripts also run with auto-discovered DB paths.
- Path on non-artifact targets (`--target backend=/x`) → error. Unknown target name → error (today: silent no-op). Keymap/backend overrides deferred (their DSNs are already config-declared and auto-discovered via store).
- No new config option; `-o push_state_path` keeps working (rc branch of `get_push_state`, helpers.py:38 — back-compat).
- No need to pass `push_state_path` via context.

sirex (commented):
I see duplicated migration logic (`is_fresh_database`, `mark_migrations`, `validate_migrations`) in two upgrade targets — `PushState` (`spinta/cli/helpers/push/state.py:60-103`) and `SqlAlchemyKeyMap` (`spinta/datasets/keymaps/sqlalchemy.py:370-420`). Both copies are near-identical (registry lookup, `sort_scripts_by_required`, `upgrade_mode` early return, fresh-DB heuristic), differing only in target enum and exception type.

This creates several problems:

1. **Duplication** — any change to migration policy has to be made twice, and the copies already drifted slightly (keymap splits table names on `_` before `has_model` checks, push_state doesn't).
2. **Bidirectional coupling** — upgrate targets imports `upgrade_script_registry` and script enums, while `spinta upgrade` scripts imports the upgrate targets back. Upgrate targets know about migration policy details (`upgrade_mode`, script tags/sorting) that belong solely to `spinta upgrade`.
3. **Mixed concerns** — `SqlAlchemyKeyMap(SqliteMigratableDb, KeyMap)` inherits both storage plumbing and migration logic through `SqliteMigratableDb`, even though these are unrelated responsibilities.

I would suggest to separate these layers:

- **Storage** (`SqliteDatabase`) — engine/conn/metadata/table handling, no migration knowledge.
- **Migrations** (`SqliteMigrations`) — the `_migrations` table, `mark_migration`/`contains_migration`.
- **Migration policy** (`spinta/cli/helpers/upgrade/migrations.py`) — `ensure_migrated()` decides: fresh DB → mark all migrations as applied; `upgrade_mode` → no-op; otherwise raise on first pending migration.

Upgrade targets (`PushState`, `SqlAlchemyKeyMap`) would use composition instead of inheritance and call a single function — `ensure_migrated(context, ...)` — asking "are migrations needed?" without knowing about the script registry, script sorting, or `upgrade_mode`. The dependency graph becomes strictly one-way: upgrade targets → upgrade policy → storage, with upgrade scripts still importing upgrade targets (migrations know what they migrate).

Below you can find more details on how this could be implemented.

## Overview

```
spinta/utils/sqlite.py
  # Pure storage, zero upgrade registry knowledge.

  class SqliteDatabase
    # engine, conn, metadata, get_table, metatable_templates, configure_engine,
    # __enter__/__exit__, create_table

  class SqliteMigrations
    # _migrations table, mark_migration, contains_migration;
    # composed over SqliteDatabase

spinta/cli/helpers/upgrade/migrations.py
  # NEW: policy layer

  def get_target_migrations(context, target)
    # registry lookup + sort_scripts_by_required

  def is_fresh_database(context, db, *, table_namer)
    # fresh-DB heuristic w/ hook

  def ensure_migrated(context, db, book, target, error_factory, table_namer)
    # fresh -> mark all;
    # upgrade_mode -> no-op;
    # else raise on first pending

spinta/datasets/keymaps/sqlalchemy.py

  class SqlAlchemyKeyMap(KeyMap)
    # self.db = SqliteDatabase(dsn)
    # self.migrations = SqliteMigrations(self.db)
    # Delegates reponsibility:
    #   conn/engine/metadata/get_table/__enter__/__exit__/copy/mark_migration/...

spinta/cli/helpers/push/components.py
  class PushState
    # same composition pattern
```

Import graph stays one-way: `upgrade targets → upgrade/migrations.py → utils/sqlite.py`, with upgrade scripts importing upgrade targets as before.

`SqliteMigratableDb` becomes an alias for back-compat (`SqliteMigratableDb = SqliteDatabase`) since `SqliteMigrations` no longer needs to be the base.

## Steps

### 1. `spinta/utils/sqlite.py` — split into two classes

- Extract everything except migration bookkeeping into `SqliteDatabase`: engine/conn/metadata properties, `configure_engine`, `__enter__`/`__exit__`/`is_entered`, `create_table`, `get_table`, `metatable_templates`, `_default_table_template` (abstract).
- `SqliteMigrations`:
  ```python
  class SqliteMigrations:
      def __init__(self, db: SqliteDatabase, migration_table_name: str = "_migrations"):
          self.db = db
          # registers the _migrations metatable template into db.metatable_templates
      def contains_migration(self, name) -> bool   # uses self.db.get_table/conn
      def mark_migration(self, name)               # unchanged logic
  ```
- Keep `SqliteMigratableDb = SqliteDatabase` alias so `tests/utils/test_sqlite.py` and any stragglers keep working (note: that test file exercises `mark_migration` on the base — it will move to target `SqliteMigrations`).
- `outdated_sqlite_db` / `apply_migration_to_outdated_db` stay here, signatures change to take `(db, book)` where they used `sqlite_db` — both call sites (`push_state/helpers.py`, `keymaps/sqlalchemy/helpers.py`) pass the composed objects.

### 2. NEW `spinta/cli/helpers/upgrade/migrations.py`

```python
def get_target_migrations(context, target: ScriptTarget) -> dict[str, Script]:
    scripts = upgrade_script_registry.get_all(targets={target.value}, tags={ScriptTag.DB_MIGRATION.value})
    return sort_scripts_by_required(scripts)

def is_fresh_database(context, db: SqliteDatabase, *, table_namer=lambda t: t) -> bool:
    # generalized from the two copies; manifest from context.get("store").manifest
    # has_model check via commands.has_model(context, manifest, table_namer(table))

def ensure_migrated(context, db, book, target, error_factory, table_namer=lambda t: t):
    if is_fresh_database(context, db, table_namer=table_namer):
        book.create_all_metatables()... # actually db.create_all_metatables() + mark all as applied
    else:
        if config.upgrade_mode:
            return
        for script in get_target_migrations(context, target).values():
            if script.check(context):
                raise error_factory(script.name)
```

`create_all_metatables` lives on `SqliteDatabase` (it already does). Marking migrations as applied on fresh DBs uses `book.mark_migration(script.name)`.

### 3. `spinta/datasets/keymaps/sqlalchemy.py` — keymap uses composition

```python
class SqlAlchemyKeyMap(KeyMap):
    def __init__(self, dsn=None):
        self.db = SqliteDatabase(dsn)
        self.migrations = SqliteMigrations(self.db)
        self.db.metatable_templates[self.sync_table_name] = ...

    # delegates (keep existing API for ~70 call sites)
    dsn, conn, engine, metadata, get_table, configure_engine, __enter__/__exit__,
    is_entered, create_all_metatables, mark_migration, contains_migration,
    migration_table_name (property -> self.migrations.migration_table_name)
```

- `copy()` (line 51-55): `copied = copy(self); copied.db._conn = None; return copied` (or better: give `SqliteDatabase` its own `copy()` that resets connection state).
- `_default_table_template` moves to a constructor argument: `SqliteDatabase(dsn, default_table_template=...)` since it was the only virtual dispatch point; keymap's lambda (line 259-267) becomes that argument.
- `prepare()` (341-347) becomes:
  ```python
  def prepare(context, keymap, **kwargs):
      ensure_migrated(
          context, keymap.db, keymap.migrations, ScriptTarget.SQLALCHEMY_KEYMAP,
          error_factory=lambda m: KeymapMigrationRequired(keymap, migration=m),
          table_namer=lambda t: t.split("_")[0],
      )
  ```
- Delete `is_fresh_database`, `validate_migrations`, `initialize_meta_tables` (lines 370-420) and the `upgrade_script_registry` import.
- Upgrade scripts (`keymaps/sqlalchemy/initial_setup.py:24-25`, `helpers.py:26-27`) keep working via delegates — no changes.

### 4. `spinta/cli/helpers/push/components.py` + `push/state.py` — PushState composition

- `PushState` composes `self.db = SqliteDatabase(dsn)` + `SqliteMigrations(self.db)` with the same delegate set; `_default_table_template` (components.py:37-62) becomes the `default_table_template` constructor arg. The `_page` metatable registration stays.
- `init_push_state` (state.py:21-57) becomes:
  ```python
  state = PushState(dburi)
  if not context.has(PUSH_STATE_PATH):
      context.set(PUSH_STATE_PATH, dburi)
  with state:
      ensure_migrated(context, state.db, state.migrations, ScriptTarget.PUSH_STATE_DB,
                      error_factory=lambda m: PushStateMigrationRequired(dsn=state.dsn, migration=m, path=state.db.engine.url.database))
      state.create_all_metatables()
      for model in models:
          state.get_table(name=model.name, model=model)
      # legacy migrate_table self-healing block: unchanged, uses state.db.engine/metadata
  ```
- Delete `is_fresh_database`, `mark_migrations`, `validate_migrations` from state.py (lines 60-103).

JustinasKen:
@sirex Currently I have few concerns with your suggestion of utilizing composition over inheritance:

1. `context` state management, since `__enter__` and `__exit__` methods are stored under `db` property, keymap and state db will have to implement their own `__enter__` and `__exit__` methods in order to call `db` context (if we want to attach root objects to `context` and make sure the `db` property exits and enters properly) so we could also do `with push_state:`, or `with keymap:`, intead of `with push_state.db:` and `with.keymap.db:`.
2. We will need to add additional methods like:
```python
def get_table(self, name: str, **kwargs) -> sa.Table:
    return self.db.get_table(name, **kwargs)
```
Otherwise we need to change all existing areas that used `keymap` and `push_state` in order to call these methods.

Second part is just convenience, but first part seems more practical and makes better use of `context` object's context manager capabilities.

spinta/cli/helpers/upgrade/scripts/push_state/rename_rev.py
JustinasKen: Fixed

spinta/cli/helpers/upgrade/scripts/keymaps/sqlalchemy/redirect_support.py
JustinasKen: Fixed

spinta/cli/helpers/upgrade/scripts/push_state/helpers.py
JustinasKen: Changed it, no need for `-o push_state_path` backwards compatibility, since this was introcuded in this PR and was not merged anywhere else.

